PR #200, opened 2022-05-13
📖 added two hyperlink to hcould-variables

guettli:
**What type of PR is this?**

/kind documentation     Adds documentation.

### DIFF
--- a/docs/topics/preparation.md
+++ b/docs/topics/preparation.md
@@ -72,8 +72,10 @@ export HCLOUD_CONTROL_PLANE_MACHINE_TYPE=cpx31 \
 export HCLOUD_WORKER_MACHINE_TYPE=cpx31
 ```
 
-HCLOUD_SSH_KEY: The SSH Key name you loaded in HCloud.
-HCLOUD_IMAGE_NAME: The Image name of your operating system.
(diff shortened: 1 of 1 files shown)


batistein:
Thank your very much! Only some minor things:
- Please change the PR Title wit `:book:` as a prefix. [Read more here.](https://github.com/syself/cluster-api-provider-hetzner/blob/main/CONTRIBUTING.md) This is needed for the automatic release notes generation.
- I think it's better if we name it something like HCLOUD_X_MACHINE_TYPE or MACHINE_TYPE
- If possible please change all the four list items to an unordered list or leave a new line so they are on top of each other

guettli:
@batistein I updated the PR. Feel free to change it, if there are still small things to change.

batistein:
Thank you very much it looks fine now! It needs only a squashing of commits. Please see: https://kubernetes.io/docs/contribute/new-content/open-a-pr/#squashing-commits

guettli:
> Thank you very much it looks fine now! It needs only a squashing of commits. Please see: https://kubernetes.io/docs/contribute/new-content/open-a-pr/#squashing-commits

squashed.

PR #235, opened 2022-06-29
:sparkles: Add specs to choose public net

### LINKED ISSUE #233: IPv6 Only/Dual Stack support?
/kind feature

Hetzner recently added a feature to allow IPv6-only VMs on HCloud. IPv4s are now considered an extra cost for each instance.
Is there currently a way to do IPv6 Only or Dual Stack deployments? I haven't seen any documentation on this anywhere so I assume not.

janiskemper:
**What type of PR is this?**
/kind feature           New functionality.

Hetzner has released a new feature in
https://github.com/hetznercloud/hcloud-go/releases/tag/v1.35.0 that
allows choosing the primary IP address of an HCloud server on creation.
This commit introduces support for this.

**Which issue(s) this PR fixes**
Fixes #233 

**TODOs**:
- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


pkg/services/hcloud/server/server.go
```diff
@@ -571,38 +579,40 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, server *h
...
-			conditions.MarkTrue(s.scope.HCloudMachine, infrav1.RateLimitExceeded)
-			record.Event(s.scope.HCloudMachine,
-				"RateLimitExceeded",
-				"exceeded rate limit with calling hcloud function AddTargetServerToLoadBalancer",
-			)
-		}
-		if hcloud.IsError(err, hcloud.ErrorCodeTargetAlreadyDefined) {
-			return nil
+	// Only if server has IPv4, otherwise Hetzner cannot handle it
+	if server.PublicNet.IPv4.IP != nil {
```
Wouter0100, line 583: Won't this if-statement break the use of a private IPv4 addresses in the Load Balancer as well? This seems because the `hasPrivateIP`-var is not taken into consideration in this if-statement. 

Or, doesn't Hetzner support private IPv4 adresses without public IPv4 being attached?

janiskemper: Hi @Wouter0100! This is a great catch, thank you! 
We would have found this problem before the next release because we would have added some e2e tests for this setup without IPv4 addresses. But you saved us definitely some time since we don't have to search for the error anymore. Thanks!

PR #273, opened 2022-08-29
:bug: Fix/kubeadm reset

janiskemper:
**What type of PR is this?**
/kind bug 

At the moment, the cloud controller manager continues to run on servers
that have been deprovisioned. This causes problems with rate limits
almost immediately, as it continues to retry requests that return
errors. For this, we add an ssh command to reset kubeadm.

Additionally, there is a small issue with ssh key names that are
uploaded to Hetzner. If they contain a line break at the end, the code
fails to recognize that it actually is the same secret. We trim line
breaks now.

Furthermore, this PR fixes the bare metal e2e test. The CAPI objects created for this test contained
wrong references to each other and there are issues with the latest
HCloud version, so we go back to 1.35.1.

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


batistein (changes requested):
please explain in the docs how to run the e2e test locally. The SSH Key needs to be exported as a base64 encoded string with the command `base64 -w 0`.

batistein (approved):

PR #313, opened 2022-10-13
:bug: Fix enabling IP address logic for HCloud

### LINKED ISSUE #312: Public IPv4 network settings for machine templates apply incorrectly
/kind bug

After discovering the functionality to disable public network access does exist, I gave it a try and [found it was not applying the configuration correctly](https://github.com/syself/cluster-api-provider-hetzner/issues/312#issuecomment-1277454481). See the comments below.

Also there is a [typo in the documentation](https://github.com/syself/cluster-api-provider-hetzner/issues/312#issuecomment-1277424020).

## Original Request for Optional Public IPv4 addresses for nodes

IPv4 addresses are optional on Hetzner instances and are not strictly necessary when nodes are only intended to be accessible within the cluster and/or external access is always done in front of a load balancer. Alternatively, one might want to exclusively use IPv6 for external access.

It would be nice to recover the expense of IPv4 addresses for pools of nodes that do not strictly need those addresses.

**Environment:**

- cluster-api-provider-hetzner version: 1.0.0-beta.3
- Kubernetes version: 1.25.2
- OS: N/A

janiskemper:
**What type of PR is this?**
/kind bug

Currently, the user can only enable/disable IPv4 and IPv6 if there is no private network. However, since the enabling/disabling is independent of the public/private net, we remove this if clause.

To make sure that users don't disable the IPv4 if there is no private net, we always enable it in that case. This is necessary for Hetzner load balancers, as they can only work with IPv4 of HCloud servers.

Fixes #312

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/reference/hcloud-machine-template.md
+++ b/docs/reference/hcloud-machine-template.md
@@ -13,6 +13,6 @@ In ```HCloudMachineTemplate``` you can define all important properties for ```HC
 | template.spec.sshKeys.hcloud.name | string | | yes | Name of SSH key |
 | template.spec.sshKeys.hcloud.fingerprint | string | | no| Fingerprint of SSH key - used by the controller |
 | template.spec.placementGroupName | string | | no | Placement group of the machine in HCloud API, must be referencing an existing placement group |
-| template.spec.publicNetwork | object | {enabledIPv4: true, enabledIPv6: true} | no | Specs about primary IP address of server. If both IPv4 and IPv6 are disabled, then the private network has to be enabled |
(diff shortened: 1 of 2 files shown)


batistein (changes requested):
please update the docs: https://github.com/syself/cluster-api-provider-hetzner/blob/main/docs/reference/hcloud-machine-template.md

There is also typo enabled -> enable

batistein (dismissed):

PR #501, opened 2022-12-05
📖 Small fixes from first readthrough

apricote:
**What type of PR is this?**

/kind documentation

Fixes typos, small inconsistencies and missing fields in the docs.

I found a few further points that are worth discussing before making any changes:

- The terms `bare metal` and `dedicated` are both used in the docs. AFAICT you used `Dedicated` where you directly reference the Hetzner Dedicated products, and then bare metal for anything internal. is there a reason why you called the CRDs `HetznerBareMetal[...]` instead of `HetznerDedicated[...]`? Probably hard to fix now though
- The quickstart path is missing some explanation of Cluster API concepts, which makes it harder to understand than necessary.
  - What is a bootstrap cluster, management cluster, workload cluster?
  - The CAPI quickstart guide links to items in their Glossary for these terms, they also have a [Concepts](https://cluster-api.sigs.k8s.io/user/concepts.html) page that we could link to.
- Quickstart guide is non-linear
  - After the cluster is up and running the user is asked to delete the cluster
  - In the following section the ["Bootstrap & Pivot"](https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html#bootstrap--pivot) strategy is shown, but not really explained how it links up with the rest of the guide.

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests


batistein:
Great! Thanks for improving the docs.

Now to your questions:
* The name has historically been used to provide a clearer distinction from hcloud dedicated. You are of course right that this feels a bit more bumpy in the overall view. However, I know that many of Hetzner's customers also call it Bare Metal internally. Changing the CRDs now would mean a lot of effort at this point and would not bring any technical advantage. 
* Currently, a caph user is expected to have a broad knowledge of working with cluster api. So far, the use of caph has been more in the corporate sector, where this knowledge was already available. However, I also see that this is an unnecessary obstacle for newcomers, and this may slow down the adoption rate. I would be very happy if you don't mind sharing your thoughts on this and/or making a PR.

PR #642, opened 2023-03-08
:bug: fixed: install matching golangci version

guettli:

There was a typo in the Makefile, so that the version of golangci-lint was not updated.


batistein:
@guettli Only for the record: This is only affecting make lint-golang if used via `make golangci-lint && BUILD_IN_CONTAINER=false make lint-golang` right?

PR #710, opened 2023-03-29
:seedling: Fix error handling of nil errors in controllers

janiskemper:
After switching to standard lib (fmt) for errors, there is a different behavior. fmt does not return nil with fmt.Errorf if err == nil. This did not work with breakReconcile, so we remove it.

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


janiskemper:
Github actions don't trigger due to an incident of Github. Close and re-open.

PR #719, opened 2023-03-31
:sparkles: Notify user when robot ssh key already exists

janiskemper:
There is currently a case where a bare metal server cannot start due to a wrong ssh key, but the user is not notified except for log messages.

Now an event is published and the host machine sets a fatal error, so that it is obvious to the user that something is wrong.

The use case is when the user uses an ssh key for bare metal which has been already uploaded to Hetzner robot under a different name. As the keys are compared via name, the controller does not figure out that the key exists in reality and uploads it again. This is when Hetzner Robot returns an error.

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


batistein (approved):

guettli (approved):
lgtm.

PR #724, opened 2023-03-31
:bug: Fix rate limit handling and refactoring SSH validation

janiskemper:
Handle rate limits everywhere where HCloud or robot API is called. Use a utility function to reduce code. Additionally, there is some refactoring of SSH key validation

**TODOs**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -38,16 +38,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/record"
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
-	// hoursBeforeDeletion      time.Duration = 36
-	// rateLimitTimeOut         time.Duration = 660
-	// rateLimitTimeOutDeletion time.Duration = 120.
-
 	sshResetTimeout      time.Duration = 5 * time.Minute
 	softwareResetTimeout time.Duration = 5 * time.Minute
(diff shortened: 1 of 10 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -112,71 +107,24 @@ func SaveHostAndReturn(ctx context.Context, client client.Client, host *infrav1.
 	return res, nil
 }
 
-func (s *Service) getSSHKeysAndUpdateStatus() (osSSHSecret *corev1.Secret, rescueSSHSecret *corev1.Secret, err error) {
-	// Set ssh status if none has been set so far
```
guettli, line 116: I think this change does not fit to the title of the PR (rate limit handling). Why was this code deleted, or is this just a "on the fly" clean up?

janiskemper: Very true. That is not related to the PR title. As it is very minor, I would still prefer keeping it in rather than re-running all tests. Thanks for being attentive!

pkg/services/baremetal/host/state_machine.go
```diff
@@ -154,14 +167,28 @@ func (hsm *hostStateMachine) updateSSHKey() actionResult {
...
-			return actResult
+		if err := validateSSHKey(rescueSSHSecret, hsm.reconciler.scope.HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef); err != nil {
+			return hsm.reconciler.recordActionFailure(infrav1.PreparationError, infrav1.ErrorMessageMissingOrInvalidSecretData)
 		}
 	}
 	return actionComplete{}
 }
 
+func validateSSHKey(sshSecret *corev1.Secret, secretRef infrav1.SSHSecretRef) error {
+	creds := sshclient.Credentials{
```
guettli, line 178: same here. The PR title does not match to this change.

guettli (commented):
lgtm, except that the ssh changes should be added to the PR title.

PR #740, opened 2023-04-04
:sparkles: Track target cluster status through conditions 

janiskemper:
We now track the status of the target cluster (workload cluster) through conditions in HetznerCluster object.

When a cluster starts, there are multiple errors that occur temporarily, but are expected. However, they litter the log files and make it harder to debug a cluster, especially for users who are not so familiar with the code.

To reduce the error messages, we use conditions that indicate that a certain step cannot be done yet. If a rare case happens when the cluster is stuck forever in a certain state which is related to errors that we ignore at the beginning, then the user can use the conditions in HetznerCluster to find out what happens.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -80,6 +80,15 @@ const (
 )
 
 const (
+	// HetznerClusterTargetClusterReady reports on whether the kubeconfig in the target cluster is ready.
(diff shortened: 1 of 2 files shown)


batistein (approved):

guettli (approved):
lgtm.

PR #744, opened 2023-04-04
:sparkles: Add conditions to HetznerBareMetalHost

janiskemper:
Adding conditions in case that something is wrong with secrets or credentials which the user has to specify. There are secrets for Hetzner robot and for the SSH keys used to provision the bare metal servers.

Removing the error message of spec.status from printcolumns, as it is used only internally for the controller to manage the provisioning process.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/baremetal/host/state_machine.go
+++ b/pkg/services/baremetal/host/state_machine.go
@@ -24,6 +24,8 @@ import (
 	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
 	sshclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/ssh"
 	corev1 "k8s.io/api/core/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/record"
 )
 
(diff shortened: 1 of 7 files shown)


batistein (dismissed):

pkg/services/baremetal/host/state_machine.go
```diff
@@ -142,6 +147,9 @@ func (hsm *hostStateMachine) updateSSHKey() actionResult {
 			}
 		}
 		if err := validateSSHKey(osSSHSecret, hsm.host.Spec.Status.SSHSpec.SecretRef); err != nil {
+			msg := fmt.Sprintf("ssh credentials for os are invalid: %s", err.Error())
```
guettli, line 150: I would remove `for os`. But maybe add the hostname to the error message.

pkg/services/baremetal/host/state_machine.go
```diff
@@ -168,6 +176,8 @@ func (hsm *hostStateMachine) updateSSHKey() actionResult {
 			}
 		}
 		if err := validateSSHKey(rescueSSHSecret, hsm.reconciler.scope.HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef); err != nil {
+			msg := fmt.Sprintf("ssh credentials for rescue are invalid: %s", err.Error())
```
guettli, line 179: Here I would use "for rescue system".

guettli (commented):
lgtm, just small comments.

pkg/services/baremetal/host/state_machine.go
janiskemper: Sorry, which hostname?

guettli: I thought about this hostname of the `hostStateMachine`. But since the message will be in the condition of this hsm, it makes no sense. Ignore my comment :-)

pkg/services/baremetal/host/state_machine.go
janiskemper: done

pkg/services/baremetal/host/state_machine.go
janiskemper: done

PR #745, opened 2023-04-04
:bug: Fix wrong lb type

janiskemper:
Updating code of load balancer library to make it easier to read and understand. Fixing error that load balancer type was translated wrongly from the value in the spec of HetznerCluster to the load balancer create options. Adding further unit tests.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/hcloud/loadbalancer/loadbalancer.go
+++ b/pkg/services/hcloud/loadbalancer/loadbalancer.go
@@ -19,18 +19,18 @@ package loadbalancer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"github.com/hetznercloud/hcloud-go/hcloud"
 	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
 	"github.com/syself/cluster-api-provider-hetzner/pkg/scope"
 	hcloudutil "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/util"
 	"github.com/syself/cluster-api-provider-hetzner/pkg/utils"
-	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/record"
-	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 // Service is a struct with the cluster scope to reconcile load balancers.
@@ -40,19 +40,18 @@ type Service struct {
 
 // NewService creates a new service object.
 func NewService(scope *scope.ClusterScope) *Service {
-	return &Service{
-		scope: scope,
-	}
+	return &Service{scope: scope}
 }
 
 // Reconcile implements the life cycle of HCloud load balancers.
-func (s *Service) Reconcile(ctx context.Context) (err error) {
-	log := ctrl.LoggerFrom(ctx)
-	log.V(1).Info("Reconcile load balancer")
-
+func (s *Service) Reconcile(ctx context.Context) error {
 	if !s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
 		return nil
 	}
+
+	log := s.scope.Logger.WithValues("reconciler", "load balancer")
+
+	log.V(1).Info("Start reconciling")
 
 	// find load balancer
 	lb, err := s.findLoadBalancer(ctx)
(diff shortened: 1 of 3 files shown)


batistein (dismissed):

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -138,163 +133,152 @@ func (s *Service) reconcileNetworkAttachement(ctx context.Context, lb *hcloud.Lo
...
-			},
-		}); err != nil {
+	lbSpec := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer
+
+	// check if type has been updated
+	if lbSpec.Type != lb.LoadBalancerType.Name {
+		opts := hcloud.LoadBalancerChangeTypeOpts{LoadBalancerType: &hcloud.LoadBalancerType{Name: lbSpec.Type}}
+		if _, err := s.scope.HCloudClient.ChangeLoadBalancerType(ctx, lb, opts); err != nil {
 			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "ChangeLoadBalancerType")
 			multierr = append(multierr, fmt.Errorf("failed to change load balancer type: %w", err))
```
guettli, line 143: about `multierr`: Since caph uses Go 1.20, you could use errors.Join(): https://pkg.go.dev/errors#Join

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -138,163 +133,152 @@ func (s *Service) reconcileNetworkAttachement(ctx context.Context, lb *hcloud.Lo
...
 		hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "CreateLoadBalancer")
-		record.Warnf(
-			s.scope.HetznerCluster,
-			"FailedCreateLoadBalancer",
-			"Failed to create load balancer: %s",
-			err)
-		return nil, fmt.Errorf("error creating load balancer: %w", err)
+		reterr := fmt.Errorf("failed to create load balancer: %w", err)
+		record.Warnf(s.scope.HetznerCluster, "FailedCreateLoadBalancer", reterr.Error())
+		return nil, reterr
```
guettli, line 249: What do you think about using this?

`err := fmt.Errorf("failed to create load balancer: %w", err)` instead of creating a `reterr`?

There are several places with this pattern.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -358,42 +343,39 @@ func (s *Service) findLoadBalancer(ctx context.Context) (*hcloud.LoadBalancer, e
...
 		case hcloud.LoadBalancerTargetTypeIP:
-			targets = append(targets, infrav1.LoadBalancerTarget{
+			targetObjects = append(targetObjects, infrav1.LoadBalancerTarget{
 				Type: infrav1.LoadBalancerTargetTypeIP,
 				IP:   target.IP.IP,
 			},
 			)
 		default:
-			return infrav1.LoadBalancerStatus{}, fmt.Errorf("unknown load balancer target type %s", target.Type)
+			panic(fmt.Sprintf("unknown load balancer target type %s", target.Type))
```
guettli, line 369: Are there alternatives to panic?

Maybe create an event?

guettli (commented):
lgtm, just some comments.

pkg/services/hcloud/loadbalancer/loadbalancer.go
janiskemper: I did this because err already exists there. I didn't want to "secretely overwrite". But I am open to do that if you think it is better

guettli: reading `err` feels a bit natural for me. If the value is `reterr` then it looks a bit strange. I personally think overwriting `err` makes sense in this case. It is not a different err. It is the same err, but wrapped with an additional message.

pkg/services/hcloud/loadbalancer/loadbalancer.go
janiskemper: Events would create an overhead here. The panic should never come up, because this is something that won't change immediately. It would make it obvious though that there was a big change.
I guess the alternative would be for me to just not say anything and just not add the load balancer targets with unknown type. 
What do you think?

pkg/services/hcloud/loadbalancer/loadbalancer.go
janiskemper: okay, I will update it!

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: ok, logging an error and setting the value to same sane default value would be good. I would avoid to panic.

PR #766, opened 2023-04-12
✨ Add support for Hetzner Cloud ARM Server Types

apricote:

See banner on https://www.hetzner.com/cloud

Fixes #

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/hcloud/client/fake/hcloud_client.go
+++ b/pkg/services/hcloud/client/fake/hcloud_client.go
@@ -33,6 +33,9 @@ const DefaultCPUCores = 1
 // DefaultMemoryInGB defines the default memory in GB for HCloud machines' capacities.
 const DefaultMemoryInGB = float32(4)
 
+// DefaultArchitecture defines the default cpu architecture for HCloud server types.
+const DefaultArchitecture = hcloud.ArchitectureX86
+
 type cacheHCloudClient struct {
 	serverCache         serverCache
 	placementGroupCache placementGroupCache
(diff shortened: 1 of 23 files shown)


janiskemper:
Thanks @apricote ! I have a very small thing: we don't need to query the serverTypes with IDs right? Because then I think we can use the function that is specific for names. What do you think?

apricote:
> Thanks @apricote ! I have a very small thing: we don't need to query the serverTypes with IDs right? Because then I think we can use the function that is specific for names. What do you think?

Makes sense with the current `HCloudMachine` types :+1: We can also replace all usages of `ListServerTypes` with the new `GetServerType`, but I wanted to keep this PR as small as possible without any additional refactorings.

I see that some tests are failing, but cant find the reason for this, they worked locally yesterday. Do you see the issue so I can fix it in one go?

janiskemper:
The unit tests are sometimes flaky. I restarted them and they work now!

janiskemper:
And if `GetServerType` makes things more efficient, then we can surely implement it in the future. But as you said, it doesn't have to be implemented now!

apricote:
> Thanks @apricote ! I have a very small thing: we don't need to query the serverTypes with IDs right? Because then I think we can use the function that is specific for names. What do you think?

Done

> @apricote and please squash the commits ;)

& Done :)

pkg/services/hcloud/client/fake/hcloud_client.go
```diff
@@ -500,26 +503,51 @@ func (c *cacheHCloudClient) DeleteServer(ctx context.Context, server *hcloud.Ser
...
 	}, nil
 }
 
+func (c *cacheHCloudClient) GetServerType(ctx context.Context, name string) (*hcloud.ServerType, error) {
+	serverType := &hcloud.ServerType{
+		Cores:        DefaultCPUCores,
+		Memory:       DefaultMemoryInGB,
+		Architecture: DefaultArchitecture,
+	}
+	if name == "cpx11" {
```
janiskemper, line 535: plz use a switch here :)

janiskemper:
LGTM

janiskemper:
sorry wrong button haha

PR #815, opened 2023-05-31
:seedling: create warning if err occurs

guettli:

Create a warning if reconciling fails.


janiskemper:
afaik we said that we don't want to pursue this @batistein @guettli . I'll close it now. If we want, we can still re-open in the future.

PR #879, opened 2023-08-31
:seedling: Do not mark conditions false if no network exists

janiskemper:
There are two conditions that are automatically set on false if there is no network. This can be misleading as there is no network intentionally. This commit changes the logic so that conditions are only set on false if there is a (temporary) problem

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/loadbalancer/loadbalancer.go
+++ b/pkg/services/hcloud/loadbalancer/loadbalancer.go
@@ -90,6 +90,11 @@ func (s *Service) reconcileNetworkAttachement(ctx context.Context, lb *hcloud.Lo
 		return nil
 	}
 
+	// nothing to do if no network is specified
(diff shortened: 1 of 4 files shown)


pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -352,6 +357,8 @@ func statusFromHCloudLB(lb *hcloud.LoadBalancer, hasNetwork bool, log logr.Logge
 	for _, target := range lb.Targets {
 		switch target.Type {
 		case hcloud.LoadBalancerTargetTypeServer:
+			p := &target.Type
+			fmt.Println("p", p)
```
guettli, line 361: @janiskemper there is a debug statement: fmt.Println()

janiskemper: merci

PR #886, opened 2023-09-01
:seedling: use controller-gen v12

guettli:
No other changes. Just called `make generate`.


guettli:
my version was outdated. The main branch was already up to date.

PR #898, opened 2023-09-08
:seedling: replace deprecated TTS for getting temp hetzner cloud tokens

apricote:

TTS is being replaced by TPS. TPS now uses (primarily) GitHub Actions OIDC tokens and has a reusable GitHub Action to get the token.

This does require additional permissions for the jobs that need to talk to TPS, to generate the OIDC tokens.

Right now, there is no way (and no need) to cleanup these tokens, they will be deleted at the end of their configured lifetime.

TPS creates temporary projects instead of tokens for one project.

You can remove the `TTS_TOKEN` from the repository settings once you have merged & validated this PR.

### DIFF
--- a/.github/actions/e2e/action.yaml
+++ b/.github/actions/e2e/action.yaml
@@ -7,9 +7,6 @@ inputs:
   e2e_make_target:
     description: "e2e_make_target"
     required: true
-  e2e_tts_token:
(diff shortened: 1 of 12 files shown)


guettli (approved):
looks good to me. Thank you @apricote!

Issue #926, opened 2023-09-19
Larger than expected request rate to hcloud `GET /v1/servers/:id` endpoint


apricote:
/kind bug

**What steps did you take and what happened:**

While adding machines to my cluster I noticed that I am being severly rate-limited. After adding metrics to the hcloud-go Client, I saw that a lot of requests are made to the `GET /v1/servers/id` endpoint every few minutes:

![screenshot-requests-servers-by-id](https://github.com/syself/cluster-api-provider-hetzner/assets/3393079/bf48ce96-09fd-4844-a497-68b3bd20118f)

Upon looking into the code, it seems that all these calls are coming from [`findServers`](https://github.com/syself/cluster-api-provider-hetzner/blob/4504b3bc4f77165a6e0a53f2c15bd81ed04e5c1d/pkg/services/hcloud/server/server.go#L583-L599). This method is called once per server reconcile. I added some logs (see patch below), and they confirm that server reconciles happen in a very spiky manner. A minute or two of silence, and then every server a few times.

<details><summary>Log Patch</summary>

```patch
diff --git a/pkg/services/hcloud/server/server.go b/pkg/services/hcloud/server/server.go
index 484cfe6d..8267bd62 100644
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -20,6 +20,7 @@ package server
 import (
 	"context"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
@@ -594,6 +595,7 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 
 		// if server has been found, return it
 		if server != nil {
+			log.Printf("[findServer#server] serverID: %d, labels: %+v", serverID, s.createLabels())
 			return server, nil
 		}
 	}
@@ -616,9 +618,11 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 	}
 
 	if len(servers) == 0 {
+		log.Printf("[findServer#none  ] serverID: %d, labels: %+v", serverID, s.createLabels())
 		return nil, nil
 	}
 
+	log.Printf("[findServer#label ] serverID: %d, labels: %+v", serverID, s.createLabels())
 	return servers[0], nil
 }
```

</details>

<details><summary>Logs</summary>

```
[...]

PR #934, opened 2023-09-20
🌱 ignore cluster condition updates when enqueueing events

apricote:

Ignore cluster condition updates when enqueueing events.

Related to #926 

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -20,8 +20,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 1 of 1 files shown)


controllers/hcloudmachine_controller.go
```diff
@@ -294,3 +301,57 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
...
+
+			oldCluster.ResourceVersion = ""
+			newCluster.ResourceVersion = ""
+
+			oldCluster.Status.Conditions = nil
+			newCluster.Status.Conditions = nil
+
+			if reflect.DeepEqual(oldCluster, newCluster) {
+				// Only insignificant fields changed, no need to reconcile
+				log.V(4).Info("Update to resource only changes insignificant fields, will not enqueue event")
```
janiskemper, line 344: we don't use V(4) and higher anywhere. If you wanna include it, then use V(1).

janiskemper (approved):
added a small comment about the log level. Otherwise
LGTM

controllers/hcloudmachine_controller.go
apricote: :+1: Copied those levels from the "not-paused" predicate in the CAPI repo. Will update in the next push.

apricote: I do not think these need to be in the default logging level, but its helpful if one can enable these through a flag. Whats the level I should use for that?

apricote: IIUC
- `V(1)` equals the Zap debug logging, which is the default in caph
- `V(0)` would be Zap info logging

I set both the `V(1)` and rebased.

apricote:
I will add some tests in a follow up PR.

janiskemper:
thanks :)

PR #935, opened 2023-09-21
✨ Handle HCloudMachineTemplate controller clusterclass as a owner

### LINKED ISSUE #921: CAPH: Handle HCloudMachineTemplates with ClusterClass as owner
## Description
The HCloudMachineTemplate controller assumes that the owner is a cluster. This is not necessarily the case when we use ClusterClasses, because then the ClusterClass is the owner - not the cluster.

In these cases, we just want to abort. We don't need to do anything if the owner is a clusterClass.

Reference: 
https://github.com/syself/cluster-api-provider-hetzner/blob/main/controllers/hcloudmachinetemplate_controller.go#L67

aniruddha2000:

Fixes #921 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachinetemplate_controller.go
+++ b/controllers/hcloudmachinetemplate_controller.go
@@ -22,22 +22,24 @@ import (
 	"fmt"
 	"time"
 
+	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
+	"github.com/syself/cluster-api-provider-hetzner/pkg/scope"
+	secretutil "github.com/syself/cluster-api-provider-hetzner/pkg/secrets"
+	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
+	"github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/machinetemplate"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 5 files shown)


controllers/hcloudmachinetemplate_controller.go
```diff
@@ -65,6 +66,13 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
 
 	// Fetch the Cluster.
 	cluster, err := util.GetOwnerCluster(ctx, r.Client, machineTemplate.ObjectMeta)
+	if cluster == nil {
```
janiskemper, line 69: maybe make the call for clusterClass first and then if you get an error, then you get an error you just ignore it and try GetOwnerCluster

janiskemper (changes requested):
I think you don't have to put it in utils package as we don't need it anywhere else

controllers/hcloudmachinetemplate_controller.go
```diff
@@ -65,20 +67,17 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
 	log = log.WithValues("HCloudMachineTemplate", klog.KObj(machineTemplate))
 
 	// Fetch the Cluster.
-	cluster, err := util.GetOwnerCluster(ctx, r.Client, machineTemplate.ObjectMeta)
-	if cluster == nil {
-		_, err := utils.GetOwnerClusterClass(ctx, r.Client, machineTemplate.ObjectMeta)
-		if err != nil {
-			return reconcile.Result{Requeue: true}, fmt.Errorf("failed to retrieve ownerReference as clusterClass: %w", err)
+	var cluster *clusterv1.Cluster
+	_, err := getOwnerClusterClass(ctx, r.Client, machineTemplate.ObjectMeta)
```
janiskemper, line 71: i like it better if you actually do it like this:
if a cluster class is found, return. 
else (you don't need it in fact because you return before): continue

janiskemper (changes requested):

janiskemper (approved):

janiskemper (changes requested):
sorry two things:
ctrlClient -> client

and why do you use githubErrors?

janiskemper (changes requested):

janiskemper (approved):

controllers/hcloudmachinetemplate_controller.go
```diff
@@ -64,13 +66,33 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
 
 	log = log.WithValues("HCloudMachineTemplate", klog.KObj(machineTemplate))
 
+	patchHelper, err := patch.NewHelper(machineTemplate, r.Client)
+	if err != nil {
+		log.Error(err, "failed to init patch helper for HCloudMachineTemplate")
```
janiskemper, line 71: remove log statement

janiskemper (dismissed):
just remove the one log statement

janiskemper (approved):

PR #942, opened 2023-09-21
:seedling: Set condition in case of errors with  hcloud images

janiskemper:
Adding conditions in case of ambiguous images or in case no image is found for a hcloud server on creation

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


aniruddha2000 (approved):
Looks good

PR #944, opened 2023-09-22
:bug: Add timeout to fix bug that hardware reboot is triggered right after software reboot

janiskemper:
Adding a timeout after every reboot so that it does not trigger another
reboot within the first (milli)seconds.

This fixes the issue that always a hardware reboot was triggered right
after the software reboot.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (approved):
lgtm.

PR #945, opened 2023-09-22
:bug: Add kubeapi service to existing load balancer

janiskemper:
The feature of taking over existing load balancers did not pay attention to the kubeapi service which is not added to the load balancer. This is due to the fact that this was done in the creation process only.

This commit adds it to the reconcilement of the extraServices so that it gets added to an existing load balancer as well.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [x] add unit tests


guettli (approved):
lgtm.

PR #948, opened 2023-09-25
:sparkles: Add logging to hetzner robot API client.

guettli:

Add logging to hetzner robot client, so that we have a better understanding why the rate-limit get reached.

### DIFF
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -49,6 +50,7 @@ import (
 type HetznerBareMetalMachineReconciler struct {
 	client.Client
 	APIReader           client.Reader
+	RateLimitWaitTime   time.Duration
 	HCloudClientFactory hcloudclient.Factory
 	WatchFilterValue    string
 }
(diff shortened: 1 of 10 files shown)


controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -148,6 +150,12 @@ func (r *HetznerBareMetalMachineReconciler) Reconcile(ctx context.Context, req r
 		}
 	}()
 
+	// check whether rate limit has been reached and if so, then wait.
+	if wait := reconcileRateLimit(hbmMachine, r.RateLimitWaitTime); wait {
+		log.Info("Rate limit exceeded - requeue in 30 seconds")
```
janiskemper, line 155: this log statement is redundant -> all infos are saved in conditions

pkg/services/baremetal/client/robot/robot_client.go
```diff
@@ -45,10 +50,32 @@ type Factory interface {
...
+
+var replaceHex = regexp.MustCompile(`0x[0123456789abcdef]+`)
+
+func (lt *LoggingTransport) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+
+	stack := replaceHex.ReplaceAllString(string(debug.Stack()), "0xX")
+
+	resp, err = lt.RoundTripper.RoundTrip(req)
+	if err != nil {
+		log.Printf("hetzner robot API. Error: %s %4s %s %q", err, req.Method, req.URL, stack)
```
janiskemper, line 65: use https://github.com/syself/cluster-api-provider-hetzner/blob/main/pkg/utils/utils.go#L195 with whatever logLevel you want

janiskemper (changes requested):
you can just remove this one line and change to our usual logger. 

I'm not too familiar with the RoundTripper, but if you tested it then it is okay.

pkg/services/baremetal/client/robot/robot_client.go
```diff
@@ -45,10 +50,34 @@ type Factory interface {
...
+		return resp, err
+	}
+	log.Info("hetzner robot API called.", "statusCode", resp.StatusCode, "method", req.Method, "url", req.URL, "stack", stack)
+	return resp, nil
+}
+
 // NewClient creates new HCloud clients.
 func (f *factory) NewClient(creds Credentials) Client {
+	client := &http.Client{
+		Transport: &LoggingTransport{http.DefaultTransport},
```
janiskemper, line 77: you can define the logger in this struct LoggingTransport I think! Redefining it would actually be worse than using the other logger I suppose

controllers/hetznerbaremetalmachine_controller.go
guettli: ok, I removed it there, and at the place where I copied it from.

janiskemper (approved):

PR #950, opened 2023-09-25
:sparkles: Introducing permanent error for host objects

janiskemper:
Introducing the permanent error on host objects to leverage the fact that there are some errors that occur during provisioning of a host that cannot be fixed, namely that the Hetzner robot device does not exist with that respective serverID, or that there is no rescue mode for the server.

In cases where there are errors that cannot be recovered without provisioning another time, the FatalError is used which leads to a FailureMessage on the bm machine.

In cases where there is a configuration problem or a failure of the scripts that the user provided, there is an error that stays on the host as well as a condition that shows a user-facing error, e.g. that there is no cloud-init installed, or that an ssh key already exists under a different name and cannot be uploaded

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (approved):
lgtm,.

PR #951, opened 2023-09-25
:sparkles: debug rate-limiting

guettli:

Debug rate-limiting issue.

If we provision several baremetal server and the ssh-key already exists, rate-limit gets reached.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -90,7 +90,7 @@ func NewService(s *scope.BareMetalHostScope) *Service {
 func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err error) {
 	initialState := s.scope.HetznerBareMetalHost.Spec.Status.ProvisioningState
 
-	oldHost := *s.scope.HetznerBareMetalHost
+	oldHost := s.scope.HetznerBareMetalHost.DeepCopy()
 
 	hostStateMachine := newHostStateMachine(s.scope.HetznerBareMetalHost, s, s.scope.Logger)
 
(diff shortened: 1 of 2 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -101,15 +101,15 @@ func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err e
...
+			result, err = SaveHostAndReturn(ctx, s.scope.Client, s.scope.HetznerBareMetalHost) // the original result gets overwritten.
 		}
 	}()
 
 	// reconcile state
 	actResult := hostStateMachine.ReconcileState()
 	result, err = actResult.Result()
 	if err != nil {
-		return reconcile.Result{Requeue: true}, fmt.Errorf("action %q failed: %w", initialState, err)
+		return reconcile.Result{RequeueAfter: time.Second * 15}, fmt.Errorf("action %q failed: %w", initialState, err)
```
guettli, line 112: @janiskemper  I added a delay here. Is this ok?

janiskemper: is there a reason? I don't think this is good. Probably this will also break our unit tests

janiskemper (changes requested):
We have HCloud twice now in 
hcloudutil.HandleHcloudRateLimitExceeded
I don't think we should do this change here

pkg/services/baremetal/host/host.go
janiskemper: apparently it doesn't, but it might as soon as we actually use this error for something... Because our timeout will be smaller than 15 seconds.

guettli: ok, I reverted this

guettli:
> We have HCloud twice now in hcloudutil.HandleHcloudRateLimitExceeded I don't think we should do this change here

ok, I reverted this change.

janiskemper (approved):

batistein (approved):

PR #969, opened 2023-09-27
:bug: Fix requeue on error

janiskemper:
There are multiple places where the controller requeues by setting the parameter explicitly or by returning an error, where it is not necessary or even bad because it leads to a loop where the controller might run in rate limits of hcloud or hrobot APIs.

This commit introduces multiple fixes:
- if something is not correct while creating hcloud servers that needs user intervention, the controller requeues after 5 minutes and does not return an error.
- in other places where there is a requeue that is not needed it is removed
- when provisioning the host fails and there is a timeout reached due to which the controller does not expect the server to provision successfully, there is no an error set on the host as well as an exponential backoff time for requeues

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -19,6 +19,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -42,8 +43,9 @@ const (
 )
 
 var (
-	errWrongLabel   = fmt.Errorf("label is wrong")
-	errMissingLabel = fmt.Errorf("label is missing")
+	errWrongLabel              = fmt.Errorf("label is wrong")
+	errMissingLabel            = fmt.Errorf("label is missing")
+	errServerCreateNotPossible = fmt.Errorf("server create not possible - need action")
 )
 
 // Service defines struct with machine scope to reconcile HCloudMachines.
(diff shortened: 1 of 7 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -97,6 +99,9 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 	if server == nil {
 		server, err = s.createServer(ctx)
 		if err != nil {
+			if errors.Is(err, errServerCreateNotPossible) {
+				return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
guettli, line 103: I think 5 minutes are a bit long. What about three minutes?

janiskemper: I think 5 min are a good number, considering the fact that manual intervention of users is required. they might not see this for hours. 

If the user wants a more responsive system, they can reduce the sync period as well.

@batistein ?

guettli (approved):
lgtm.

pkg/services/hcloud/server/server.go
batistein: 5 min are enough to be responsive enough. From my understanding this case only happens when something is misconfigured right?

PR #970, opened 2023-09-27
:sparkles: Try ssh reboot in rescue

janiskemper:
If a bare metal server is provisioned for the second time after a first successful provisioning, and as long as the ssh key did not change, we can still access it with ssh and can perform an ssh reboot instead of an API reboot (sw, hw). In this way, we can save some API calls.

This commit introduces this change

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


pkg/services/baremetal/host/host.go
```diff
@@ -204,6 +204,25 @@ func (s *Service) actionPreparing() actionResult {
...
+		Port:       s.scope.HetznerBareMetalHost.Spec.Status.SSHSpec.PortAfterCloudInit,
+		IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
+	})
+
+	// Check hostname with sshClient
+	out := sshClient.GetHostName()
+	if trimLineBreak(out.StdOut) != "" {
+		// we managed access with ssh - we can do an ssh reboot
+		if err := handleSSHError(sshClient.Reboot()); err != nil {
+			return actionError{err: fmt.Errorf("failed to reboot server: %w", err)}
```
guettli, line 218: I would use "failed to reboot server via ssh: %w".

guettli (approved):
lgtm, just one comment.

PR #971, opened 2023-09-27
:seedling: handle deprovision of bare metal.

guettli:

Up to now deprovisioning of bare-metal servers failed. The reason was because there was output on stdout.

We fixed that and improved the deprovisioning (delete all pods, stop and disable kubelet via systemclt, remove /etc/kubernetes).

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -1426,9 +1426,11 @@ func (s *Service) actionDeprovisioning() actionResult {
 			IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
 		})
 		out := sshClient.ResetKubeadm()
-		if err := handleSSHError(out); err != nil {
-			record.Warnf(s.scope.HetznerBareMetalHost, "FailedResetKubeAdm", "failed to reset kubeadm: %s", err.Error())
-			s.scope.Error(err, "failed to reset kubeadm")
+		s.scope.V(1).Info("Output of ResetKubeadm", "stdout", out.StdOut, "stderr", out.StdErr, "err", out.Err)
+		if out.Err != nil {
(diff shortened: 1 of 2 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -1426,9 +1426,13 @@ func (s *Service) actionDeprovisioning() actionResult {
 			IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
 		})
 		out := sshClient.ResetKubeadm()
-		if err := handleSSHError(out); err != nil {
+		s.scope.Info("Output of ResetKubeadm", "stdout", out.StdOut, "stderr", out.StdErr, "err", out.Err)
```
janiskemper, line 1429: you only need this for debugging - not in the controller I guess

pkg/services/baremetal/host/host.go
```diff
@@ -1426,9 +1426,13 @@ func (s *Service) actionDeprovisioning() actionResult {
...
+		if out.Err != nil {
+			err := fmt.Errorf("failed to perform ssh command: %w", out.Err)
```
janiskemper, line 1431: not needed I think

pkg/services/baremetal/host/host.go
```diff
@@ -1426,9 +1426,13 @@ func (s *Service) actionDeprovisioning() actionResult {
...
 			record.Warnf(s.scope.HetznerBareMetalHost, "FailedResetKubeAdm", "failed to reset kubeadm: %s", err.Error())
 			s.scope.Error(err, "failed to reset kubeadm")
```
janiskemper, line 1433: you can remove this log statement - it's included in the event above

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
guettli: @janiskemper I would like to keep that. I hope it is ok for you, since it does not happen often.

janiskemper (approved):

PR #984, opened 2023-09-28
:bug: Fix bm remediation if no host attached

janiskemper:
There was a bug where the remediation object is there for a HetznerBareMetalMachine that has no host attached. This commit handles this case and stops remediation instead of returning an error.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (approved):
lgtm.

PR #991, opened 2023-09-29
:bug: Fix robot log and switch to pflag

janiskemper:
Fix logger of robot api and switch to pflag package

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -24,6 +24,11 @@ settings = {
     "cabpt_version": "v0.5.6",
     "cacppt_version": "v0.4.11",
     "cert_manager_version": "v1.11.0",
+    "extra_args": {
+        "hetzner": [
+            "--log-level=debug",
+        ],
+    },
     "kustomize_substitutions": {
         "HCLOUD_REGION": "fsn1",
         "CONTROL_PLANE_MACHINE_COUNT": "3",
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -27,8 +27,8 @@ spec:
       containers:
         - command:
             - /manager
+          args:
             - "--leader-elect=true"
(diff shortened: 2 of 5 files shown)


Tiltfile
```diff
@@ -24,6 +24,11 @@ settings = {
     "cabpt_version": "v0.5.6",
     "cacppt_version": "v0.4.11",
     "cert_manager_version": "v1.11.0",
+    "extra_args": {
+        "hetzner": [
+            "--log-level=info",
```
batistein, line 29: i think this should be debug

batistein (changes requested):

batistein (approved):

apricote:
Just FYI, this caused my existing deployment to fail after updating the image because one flag i was using had a single hyphen (`-metrics-bind-address`) and this was no longer allowed.

---

OT: It would be very nice if CAPH supported the same flags as other CAPI components, to make using it with the capi-operator nicer. In this case, the capi-operator expects the flag [`--metrics-bind-addr`](https://github.com/kubernetes-sigs/cluster-api-operator/blob/cf9a1ae3cc1a878cb94bb3d0b396c255379a971c/internal/controller/component_customizer.go#L193-L195) instead of CAPHs `--metrics-bind-address`. Because of this I need to add the flag manually instead of using the field provided in the `InfrastructureManager` CR.

I will open an issue for this when I find the time.

janiskemper:
then you were smarter than us, because we realized that our example deployments did it wrongly and the flags did not take effect (they were not changing anything really, that's why we also didn't realize).

I guess your proposed change makes sense. AFAIK we should be mostly aligned with other providers though.

@batistein can we document this change somewhere in case other people run into this?

PR #1002, opened 2023-10-09
:seedling: Installimage: create event containing debug.txt

guettli:

Up to now it is hard to debug `installimage` because we don't see the error messages, except we ssh into the machine.

But the machine could get provisioned again, if the first run was not successful.

This PR creates an event which contains the stdout/stderr and /root/debug.txt


pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -243,12 +243,17 @@ EOF`, cmd))
 		return out
 	}
 
-	out = c.runSSH(`sh /root/install-image-script.sh`)
-	if out.Err != nil {
-		return out
+	installImageOut := c.runSSH(`sh /root/install-image-script.sh`)
+
+	debugTxtOut := c.runSSH(`cat /root/debug.txt`)
+	installImageOut.StdOut = fmt.Sprintf("stdout:\n%s\n###################\n\nstderr:\n%s###################\n\ndebug.txt:\n%s%s",
```
janiskemper, line 249: should we not have "stdout:" etc. as well as "err:" of the debugTxtOut in the print statement? 
Two problems I see: 
 a) It is not so visible what is stdout and stderr of the cat command. You are the expert and have to say whether that's okay
 b) what if the command does not work? E.g. due to a temporary problem of accessing the server with ssh? Then stdout and maybe stderr will be both empty. In that case we should have the error printed that it is clear why.

guettli: > a) It is not so visible what is stdout and stderr of the cat command. You are the expert and have to say whether that's okay

I think that's ok, since stderr of `cat /root/debug.txt` should almost always be empty. But I will add a newline between both, so that it is easier to read, if there is stderr.

> b) what if the command does not work? 

Good point. I will add it. But installimage is successful, but getting debug.txt is not? Will this result in returning an err?

janiskemper: I think you shouldn't return an error of failing to get debug.txt, but rather print the error  in the above output..

guettli: > I think you shouldn't return an error of failing to get debug.txt, but rather print the error in the above output..

good point. I updated the PR.

pkg/services/baremetal/host/host.go
```diff
@@ -1181,13 +1189,43 @@ func verifyConnectionRefused(sshClient sshclient.Client, port int) bool {
...
 	sshClient := s.scope.SSHClientFactory.NewClient(sshclient.Input{
 		PrivateKey: sshclient.CredentialsFromSecret(s.scope.OSSSHSecret, s.scope.HetznerBareMetalHost.Spec.Status.SSHSpec.SecretRef).PrivateKey,
 		Port:       s.scope.HetznerBareMetalHost.Spec.Status.SSHSpec.PortAfterCloudInit,
 		IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
 	})
 
+	defer func() {
+		// Create an Event which contains the content of /var/log/cloud-init-output.log
+
+		if _, ok := ar.(actionContinue); ok {
```
janiskemper, line 1202: I think you should also ignore actionError. In that case we return an error message because of an unexpected error... 
Note that this does not include the case where we realize that cloud init failed, then we return actionFailed instead.

pkg/services/baremetal/host/host.go
```diff
@@ -1181,13 +1189,43 @@ func verifyConnectionRefused(sshClient sshclient.Client, port int) bool {
...
+		}
+		out := sshClient.GetCloudInitOutput()
+		if out.Err != nil || out.StdErr != "" {
+			record.Warnf(s.scope.HetznerBareMetalHost, "GetCloudInitOutputFailed",
+				fmt.Sprintf("GetCloudInitOutput failed: stdout %q, stderr %q, err %q",
+					out.StdOut, out.StdErr, out.Err.Error()))
+			return
+		}
+		_, ok := ar.(actionComplete)
+		if ok {
```
janiskemper, line 1214: you don't need this if else statement.. You can just make a simple one where you define errString and then you just append the stdout of cloud init to the error string (which might be empty)

janiskemper (changes requested):

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -395,3 +429,53 @@ func (c *sshClient) runSSH(command string) Output {
...
+		// Check if the line matches any of the regexes
+		matches := false
+		for _, re := range compiledRegexes {
+			if re.MatchString(line) {
+				matches = true
+				break
+			}
+		}
+
+		// If the line does not match any regex, print it
```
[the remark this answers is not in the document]
guettli: The comment is correct: if a regex matches, then the line should get ignored (not printed). I removed the comment. I think the code speaks for itself.

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -395,3 +429,53 @@ func (c *sshClient) runSSH(command string) Output {
...
+		if matches {
+			continue
+		}
+		new = append(new, line)
+	}
+
+	if err := scanner.Err(); err != nil {
+		return fmt.Sprintf("Error reading string: %v\n", err)
```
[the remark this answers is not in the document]
guettli: I am happy with this. If there is an error at this level an event containing an error message is fine. 

It is very unlikely that an error occurs here, since no network or disk IO is happening here. It is just in memory (reading a string).

pkg/services/baremetal/host/host.go
```diff
@@ -1053,6 +1057,10 @@ func (s *Service) actionProvisioning() actionResult {
 
 		isSSHTimeoutError, isSSHConnectionRefusedError, err := analyzeSSHOutputInstallImage(out, rescueSSHClient, portAfterInstallImage)
 		if err != nil {
+			if errors.Is(err, errUnexpectedHostName) {
```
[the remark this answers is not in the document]
guettli: Sorry, this was not in the text of the PR.

Since I saw these places, I added a solution for "Detect if bm machine is owned by a different cluster".

pkg/services/baremetal/host/host.go
guettli: I have concerns: Is it possible that the controller hangs in returning actionError until a MHC deletes the machine, and then we don't get the content of the file?

pkg/services/baremetal/host/host.go
```diff
@@ -1200,6 +1238,10 @@ func (s *Service) actionEnsureProvisioned() actionResult {
 
 		isTimeout, isSSHConnectionRefusedError, err := analyzeSSHOutputProvisioned(out)
 		if err != nil {
+			if errors.Is(err, errUnexpectedHostName) {
```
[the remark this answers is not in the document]
guettli: Here is the code:

```
	out := sshClient.GetHostName()
	if trimLineBreak(out.StdOut) != wantHostName {
		// give the reboot some time until it takes effect
		if s.hasJustRebooted() {
			return actionContinue{delay: 2 * time.Second}
		}

		isTimeout, isSSHConnectionRefusedError, err := analyzeSSHOutputProvisioned(out)
		if err != nil {
				// One possible reason: The machine gets used by a second wl-cluster
				record.Warn(s.scope.HetznerBareMetalHost, "UnexpectedHostName", err.Error())
			}
			return actionError{err: fmt.Errorf("failed to handle incomplete boot - provisioning: %w", err)}
		}
```

`trimLineBreak(out.StdOut)` could be empty because of ssh-timeout.

That's why I check for `errUnexpectedHostName`.

I want to create an Warning-Event if the host changed. This can happen if the hbmh gets used accidentally in a second cluster. I wasted some times because of this some days ago.

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -395,3 +429,52 @@ func (c *sshClient) runSSH(command string) Output {
...
+		`^Unpacking .* \.\.\..*`,                     // Unpacking libatk1.0-0:amd64 (2.35.1-1ubuntu2) ...\r
+	}
+
+	// Compile the regexes
+	compiledRegexes := make([]*regexp.Regexp, 0, len(regexes))
+	for _, re := range regexes {
+		compiled, err := regexp.Compile(re)
+		if err != nil {
+			msg := fmt.Sprintf("removeUselessLinesFromCloudInitOutput: Failed to compile regex %s: %v\n", re, err)
+			fmt.Println(msg)
```
janiskemper, line 449: you can remove that right?

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -395,3 +429,52 @@ func (c *sshClient) runSSH(command string) Output {
...
+		}
+
+		if matches {
+			continue
+		}
+		output = append(output, line)
+	}
+
+	if err := scanner.Err(); err != nil {
+		return fmt.Sprintf("Error reading string: %v\n", err)
```
janiskemper, line 477: I think you didn't push the code yet we looked at together...

janiskemper (changes requested):

pkg/services/baremetal/client/ssh/ssh_client.go
guettli: yes, I pushed these changes now.

pkg/services/baremetal/client/ssh/ssh_client.go
guettli: yes, I pushed these changes now.

janiskemper (approved):

PR #1004, opened 2023-10-10
 🌱Added function to update control plane endpoint

### LINKED ISSUE #977: Add utility function to update controlPlaneEndpoint
## Description
Update the current code of the HetznerCluster controller (https://github.com/syself/cluster-api-provider-hetzner/blob/main/controllers/hetznercluster_controller.go#L202-L225)  with creating a separate utility function that contains the existing code and which gets the hetznerCluster as input

Then unit test this utility function accordingly.

razashahid107:
This one fixes a part of the issue #977. Added utiltiy function to update control plane endpoints. 
Fixes #977 

This is not the complete PR. I'll add the unit tests. I request to get my function reviewed. I want to know what I can make better with my function. I'm currently feeling a little unsure about the false condition that I've added to the end of the function which can potentially set the status of the hetzner cluster to false. Is it okay if it stays that way? Or do I edit it up? I'd love feedback. 

No it does not
**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hetznercluster_controller.go
+++ b/controllers/hetznercluster_controller.go
@@ -198,31 +198,9 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	if err := placementgroup.NewService(clusterScope).Reconcile(ctx); err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to reconcile placement groups for HetznerCluster %s/%s: %w", hetznerCluster.Namespace, hetznerCluster.Name, err)
 	}
-
-	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
-		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
-			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
-			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)
-
-			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
-				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
-					Host: defaultHost,
-					Port: defaultPort,
-				}
-			} else {
-				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
-					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
-				}
-				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
-					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
-				}
-			}
-
-			hetznerCluster.Status.Ready = true
-		}
-	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
-		hetznerCluster.Status.Ready = true
-	}
+	
+	// update control plane endpoint 
+	hetznerCluster.Status.Ready = SetControlPlaneEndpoint(hetznerCluster)
 
 	// delete deprecated conditions of old clusters
 	conditions.Delete(clusterScope.HetznerCluster, infrav1.DeprecatedHetznerClusterTargetClusterReadyCondition)
@@ -258,6 +236,33 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	conditions.MarkTrue(hetznerCluster, infrav1.TargetClusterSecretReadyCondition)
 
 	return reconcile.Result{}, nil
+}
+
+func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
+	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
+		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
+			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
+			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)
+
+			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
+				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
+					Host: defaultHost,
+					Port: defaultPort,
+				}
+			} else {
+				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
+					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
+				}
+				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
+					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
+				}
+			}
+			return true
+		}
+	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
+		return true
+	}
+	return false
 }
 
 func (r *HetznerClusterReconciler) reconcileDelete(ctx context.Context, clusterScope *scope.ClusterScope) (reconcile.Result, error) {
--- a/controllers/hetznercluster_controller.go
+++ b/controllers/hetznercluster_controller.go
@@ -249,6 +249,7 @@ func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
 					Host: defaultHost,
 					Port: defaultPort,
 				}
+				return true
 			} else {
 				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
 					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
@@ -256,8 +257,8 @@ func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
 				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
 					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
 				}
+				return true
 			}
-			return true
 		}
 	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
 		return true
--- a/controllers/hetznercluster_controller_test.go
+++ b/controllers/hetznercluster_controller_test.go
@@ -850,12 +850,12 @@ func TestSetControlPlaneEndpoint(t *testing.T) {
 		want bool
 	}{
 		{
-			name: "the function should return true if Loadbalancer is enabled, IPV4 does not include a <nil> string and the ControlPlaneEndpoint is set to nil",
+			name: "the function should return true if Loadbalancer is enabled, IPV4 does not include a <nil> string and the ControlPlaneEndpoint is set to nil. This will enable changes to the ControlPlaneEndpoint's host and port",
 			hetznerCluster: &infrav1.HetznerCluster{Spec: infrav1.HetznerClusterSpec{ControlPlaneLoadBalancer: infrav1.LoadBalancerSpec{Enabled: true}, ControlPlaneEndpoint: nil}, Status: infrav1.HetznerClusterStatus{ControlPlaneLoadBalancer: &infrav1.LoadBalancerStatus{IPv4: "xyz"}}},
 			want: true,
 		},
 		{
-			name: "the function should return true if Loadbalancer is enabled, IPV4 does not include a <nil> string and the ControlPlaneEndpoint is NOT set to nil",
+			name: "the function should return true if Loadbalancer is enabled, IPV4 does not include a <nil> string and the ControlPlaneEndpoint is NOT set to nil. This testcase will enable changes to the ControlPlaneEndpoint's host and port",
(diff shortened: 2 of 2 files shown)


controllers/hetznercluster_controller.go
```diff
@@ -260,6 +238,33 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
 }
 
+func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
+	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
+		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
+			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
+			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)
+
+			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
+				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
```
razashahid107, line 248: Just noticed with my current code, this line won't take effect, because the objects are passed by value and not by reference hence the Host and Port will never get set right? 
I was thinking maybe I can just pass the hetznerCluster parameter by reference and then use it in the whole function by reference, this way I don't need to return any value either. Is this approach okay?

janiskemper: you store everything in hetznerCluster which is a pointer. therefore it should take effect!

janiskemper: and yes, the approach is okay so far!

razashahid107: Okay so I do not need to alter with the SetControlPlaneEndpoint() function? 
I can begin with the unit test then right?

janiskemper: if you didn't change any logic, then all is well! Some minor cosmetics I would change. But go ahead with unit tests for now!

janiskemper: I would use the standard Go library for the unit tests of this function, instead of Gingko/gomega we use with envtest

razashahid107: Aight I'll get started with the unit tests. Also thank you for making time to review the PR Janis.

razashahid107:
Hi @janiskemper . Okay basic question. I have worked with a few unit tests earlier however they were for relatively simpler Go code probably hence they contained functions such as func Testxxx(). Now going through hetznercluster_controller_tests.go its a little different. I've tried going through the test file and I'm still understanding things, but can you maybe just point me to the right direction with where do I start. 
Like do I just create a function like func TestSetControlPlaneEndpoint() and maybe proceed to write code there? I'm sorry I just feel a little lost and I feel a little direction would really help.

janiskemper:
that's exactly why I said
> I would use the standard Go library for the unit tests of this function, instead of Gingko/gomega we use with envtest

yes you can and should proceed like  that

razashahid107:
Oh okay okay. Aight. I'll proceed this way. Thank you Janis

razashahid107:
Hey @janiskemper can you please go through the below code? 
So I'm basically trying to pass the hetznercluster as input to the function, now what I'm not able to understand is how do I pass it as an argument in the code below? I'm sure there is something very basic I'm missing. I'd love if you can help out. 

> tests := []struct{
		name string
		args args
		want bool
	}{
		{
			name: "should return true",
			args: args{hetznerCluster},
			want: true,
		},
		{
			name: "should return false",
			args: args{hetznerCluster},
			want: false,
		},
	}

janiskemper:
I'm not sure what "args" is in your code...

razashahid107:
Oh I'm sorry. I had created a type which includes the hetznercluster. Wait I'll put in the code: (cc @janiskemper)

> type args struct {
		hetznerCluster *infrav1.HetznerCluster
	}
	tests := []struct{
		name string
		args args
		want bool
	}{
		{
			name: "should return true",
			args: args{hetznerCluster},
			want: true,
		},
		{
			name: "should return false",
			args: args{hetznerCluster},
			want: false,
		},
	}

razashahid107:
Hey @janiskemper . I'd really appreciate if you can provide feedback on the latest commit. As of now I've kept the two test cases almost same, I understand I will have to make changes to them. But I feel a little lost as to how do I make updates with the hetznercluster parameter. Like how do I set a case in which the test returns false, and how do I set a case in which a test returns true, this is what is confusing me. I hope I'm able to convey my point. 

I also hope I am going in the right direction. I'd really appreciate your feedback.

janiskemper:
From the code side you have created a pattern that you can use to write tests. But I would say that you struggle right now what you should test.. 
Well, pretty much tests that cover most of or even all of the logic that the function contains. 

According to the pattern 
if I do xyz, then I expect abc.
If i do xyz2, then I expect abc2. 

One shortcoming of your current approach is that not only it is interesting which boolean the function returns, but also the changes to the hetznerCluster have to be validated. 

I suggest that you first start with a list of tests that you wanna implement... Not writing code yet, just what would make sense to do.

razashahid107:
```
func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)

			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
					Host: defaultHost,
					Port: defaultPort,
				}
			} else {
				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
				}
				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
				}
			}
			return true
		}
	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
		return true
	}
	return false
}
```
Okay so according to the code above I feel we'd want a few specefic test cases to take place which return true:

1. So if the Loadbalancer is enabled and its IPV4 holds a string value of "nil" then the test should by default just return true (a few things will get set in the if condition, but they do not matter to us, what matters is the return value which will be true). 
2. Otherwise if that is not true and if the ControlPlaneEndpoint is nil then we should also return true (just that this time we did not set the values for other things as we did in the previous if statement)

If none of these match then the function should return false. 

Seeing the above conditions, I think I should include three test cases. However, the problem still persists, how should I mention these conditions when writing the unit test? I dont think its supposed to be mentioned in the tests struct or is it?

janiskemper:
I believe that there are a lot more possible tests. Pretty much every if-else statement gives you another possibility to test something

razashahid107:
Oh, I'll look into this. I thought I'd maybe base my test cases solely on the return statements. 
Can you maybe suggest an example for the other test cases possible? I can maybe take direction from that

razashahid107:
@janiskemper whats confusing me is that if I take the if statement 
`if hetznerCluster.Spec.ControlPlaneEndpoint == nil` and add it to the test case, what do I put in the want? Do I just put in true (I think if the function is at this stage of the program, then it will eventually return true)? Or am I mistaking?

razashahid107:
Hi @janiskemper . So I started writing my test cases again, and I'm finding myself comfortable addressing possible cases, thank you for the activity earlier. 
I have a few questions as I write the tests. 
```
func SetControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) (bool) {
		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)

			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
					Host: defaultHost,
					Port: defaultPort,
				}
			} else {
				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
				}
				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
				}
			}
			return true
		}
	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
		return true
	}
	return false
}

```
So with the if conditions above in hetznercluster_controller.go at
```
else {
  if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
```
what if the values are as such: Host == "xzy" and port == "1". The function will still return true. I think it should return false, because we did not set the values of Host and Port. Do we need to edit this in the function?

razashahid107:
@janiskemper Can you please review the latest commit?

janiskemper:
the main part of the test should also be to check the changes that have been made to the hetznercluster... that you don't test currently"

razashahid107:
Ahan. @janiskemper can you please elaborate on that a little. Like are we talking about the changes to `hetznerCluster.Spec.ControlPlaneEndpoint.Host` and `hetznerCluster.Spec.ControlPlaneEndpoint.Port`?

janiskemper:
about hetznerCluster.Spec.ControlPlaneEndpoint in general, yeys

razashahid107:
I'm sorry @janiskemper I feel a little lost as to what am I missing. I'll add another commit maybe rephrasing the test case names. I've tried addressing changes being made to the ControlPlaneEndpoint.Host and ControlPlaneEndpoint.Port in the first two test cases. Can you please go through it? 
I'd love feedback at what I'm missing

janiskemper:
you just have to also include test cases for changes of hetznerCluster.Spec.ControlPlaneEndpoint.. At the moment you only test the return boolean in https://github.com/syself/cluster-api-provider-hetzner/pull/1004/files#diff-17dd1bcc2376d005c272189be910b2625e9a01d4bc08d71663eb972ed88f0df6R877.
Just not focus on coding for now.. Look at the function from a high level perspective and think about the logic it contains and for stuff that you might want to test for

janiskemper:
yes, exactly!

razashahid107:
@janiskemper. So I was earlier adding test cases, but I think the test cases are fine, like I do not need to add more. The condition I added should be okay right? Or would you recommend adding another test case?
If everything is okay, then do I just squash the commits and push?

janiskemper:
can you maybe take a step back and think about every test that you could potentially do? Because I think this structure of the fixed table-driven test holds you back. Just make a list of tests that could be implemented. I believe we are still missing out many

razashahid107:
- I can test if ControlPlaneLoadbalancer.IPV4 is `"<nil>"` then the variables `defaultHost` and `defaultPort` should not be nil. (I dont think I can do this though, because these are local variables and they are not being passed in the function. Or can I?)
- I can also try another test which checks if Host is empty like `hetznerCluster.Spec.ControlPlaneEndpoint.Host == ""` then as the test ends it should be equal to `hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4`. 
- I can do the same thing with `hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0` and check if it equals `hetznerCluster.Spec.ControlPlaneLoadBalancer.Port`
Am I approaching this correct?

janiskemper:
just for an example: What if controlPlaneLoadBalancer is not enabled? Then other things happen.. You can really go through the whole logic and pretty much any possible combination of things that might happen can be (theoretically) tested.

janiskemper:
you can check what happens if load balancer is not enabled and if the controlPlaneEndpoint is nil / not nil. 
You can check what happens if the load balancer is enabledand the IPv4 is "<nil>", or is not.
You can check what happens if ControlPlaneEndpoint is nil or not nil and what happens if host and port are set or not. 
You can check what happens for different values of ControlPlaneEndpoint and ControlPlaneLoadBalancer. 
You can check all of that!

razashahid107:
🤦 . I'm sorry @janiskemper. Earlier you mentioned about checking changes made to hetznercluster, I think I got very zoomed in on that only. I was trying to avoid want statements, which is why it did not make sense to me. Can you please go through the latest commit. I feel that there can be more test cases with more values, but they'll more or less be same. Is there still something that I'm missing.

janiskemper:
maybe you can do the following: I think the table-driven test really makes you think very one-directional. Can you completely get rid of it and do one test case after another? I think that would really help you... I mentioned a lot of tests in my previous comment that you can potentially do.

razashahid107:
I thought I made test cases for all the test cases you suggested? I'm defo missing some major point. I can discard all my tests and restart from line one. Would you suggest I do that?

janiskemper:
you implented table driven tests. Can you re-implement what you currently have (and maybe new tests) in a manner that doesn't use table-driven tests?

razashahid107:
Writing a function for each test counts? Something similar to whats mentioned in this link [https://golang.cafe/blog/golang-table-test-example]

janiskemper:
yes, but maybe a pattern that uses `t.Run("InvalidMajorVersion", func(t *testing.T) {// test code}` within a test is better.

janiskemper:
closing as this is replaced by #1023

PR #1008, opened 2023-10-17
🌱 Refactor events for remediation controllers

aniruddha2000:

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/remediation/remediation.go
+++ b/pkg/services/baremetal/remediation/remediation.go
@@ -60,7 +60,7 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 			record.Warn(s.scope.BareMetalRemediation, "FailedSettingConditionOnMachine", err.Error())
 			return res, err
 		}
-		record.Eventf(s.scope.BareMetalRemediation, "ExitRemediation", "exit remediation because bare metal machine has no host annotation")
+		record.Event(s.scope.BareMetalRemediation, "ExitRemediation", "exit remediation because bare metal machine has no host annotation")
 		return res, nil
 	}
 
@@ -158,12 +158,15 @@ func (s *Service) remediate(ctx context.Context, host infrav1.HetznerBareMetalHo
 	// add annotation to host so that it reboots
 	host.Annotations, err = addRebootAnnotation(host.Annotations)
 	if err != nil {
+		record.Warn(s.scope.BareMetalRemediation, "FailedAddingRebootAnnotation", err.Error())
 		return fmt.Errorf("failed to add reboot annotation: %w", err)
 	}
 
 	if err := patchHelper.Patch(ctx, &host); err != nil {
 		return fmt.Errorf("failed to patch: %s %s/%s %w", host.Kind, host.Namespace, host.Name, err)
 	}
+
+	record.Event(s.scope.BareMetalRemediation, "AnnotationAdded", "Reboot annotation is added to the BareMetalHost")
 
 	// update status of BareMetalRemediation object
 	now := metav1.Now()
@@ -191,6 +194,7 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 		record.Warn(s.scope.BareMetalRemediation, "FailedSettingConditionOnMachine", err.Error())
 		return res, err
 	}
+	record.Event(s.scope.BareMetalRemediation, "SetOwnerRemediatedCondition", "because retryLimit is reached and reboot timed out")
 
 	return res, nil
 }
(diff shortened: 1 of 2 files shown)


aniruddha2000:
@janiskemper @guettli please have a look. I tried to add events for most of the places, but however, I don't understand this 

> when the machine is healthy again and the remediation object gets removed.

janiskemper:
to me it looks mostly good. The case that you have mentioned there is something that we cannot do without adding a finalizer to the remediation object, which I find a bit extreme to get an event..
Right now, the remediation object gets deleted and since we don't have a finalizer, we don't even realize. 

@guettli would you say that this is so important that we should introduce a finalizer?

guettli:
@janiskemper at the moment I think it is not needed to add finalizer to get an event. Maybe we will add this later.

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -158,8 +158,10 @@ func (s *Service) remediate(ctx context.Context, host infrav1.HetznerBareMetalHo
 	// add annotation to host so that it reboots
 	host.Annotations, err = addRebootAnnotation(host.Annotations)
 	if err != nil {
+		record.Warn(s.scope.BareMetalRemediation, "FailedAddingRebootAnnotation", err.Error())
 		return fmt.Errorf("failed to add reboot annotation: %w", err)
 	}
+	record.Event(s.scope.BareMetalRemediation, "AnnotationAdded", "Reboot annotation is added to the BareMetalHost")
```
janiskemper, line 164: this is only the case after patching

pkg/services/hcloud/remediation/remediation.go
```diff
@@ -58,8 +58,10 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		s.scope.HCloudRemediation.Status.Phase = infrav1.PhaseDeleting
 
 		if err := s.setOwnerRemediatedCondition(ctx); err != nil {
+			record.Warn(s.scope.HCloudRemediation, "FailedSettingConditionOnMachine", err.Error())
 			return res, fmt.Errorf("failed to set conditions on CAPI machine: %w", err)
 		}
+		record.Warn(s.scope.HCloudRemediation, "FailedToFindServer", "Server doesn't exist")
```
janiskemper, line 64: make a comment like the one above with "exit remediation because bare metal server does not exist"
-> FailedToFindServer is not very helpful if you see only that event

pkg/services/hcloud/remediation/remediation.go
```diff
@@ -143,6 +149,7 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 	s.scope.HCloudRemediation.Status.Phase = infrav1.PhaseDeleting
 
 	if err := s.setOwnerRemediatedCondition(ctx); err != nil {
+		record.Warn(s.scope.HCloudRemediation, "FailedSettingConditionOnMachine", err.Error())
```
janiskemper, line 152: you can make another event below setOwnerRemediatedCondition that this condition was set now and we exit remediation. 

Please check again all occurrences of setOwnerRemediatedCondition in both controllers to make sure that you found all cases and set events

janiskemper (changes requested):

pkg/services/hcloud/remediation/remediation.go
```diff
@@ -152,6 +152,7 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 		record.Warn(s.scope.HCloudRemediation, "FailedSettingConditionOnMachine", err.Error())
 		return res, fmt.Errorf("failed to set conditions on CAPI machine: %w", err)
 	}
+	record.Event(s.scope.HCloudRemediation, "SetOwnerRemediatedCondition", "exit remediation because owner remediated condition was set on machine")
```
janiskemper, line 155: because retryLimit is reached and reboot timed out

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -193,6 +194,7 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 		record.Warn(s.scope.BareMetalRemediation, "FailedSettingConditionOnMachine", err.Error())
 		return res, err
 	}
+	record.Event(s.scope.BareMetalRemediation, "SetOwnerRemediatedCondition", "exit remediation because owner remediated condition was set on machine")
```
janiskemper, line 197: because retryLimit is reached and reboot timed out

janiskemper (changes requested):

janiskemper (approved):

PR #1016, opened 2023-10-20
:sparkles: use vendored installimage

guettli:

Use a vendored installimage.

We wanted a clearly versioned installimage.


Makefile
```diff
@@ -788,3 +788,9 @@ tilt-up: env-vars-for-wl-cluster $(ENVSUBST) $(KUBECTL) $(KUSTOMIZE) $(TILT) clu
 .PHONY: watch
 watch: ## Watch CRDs cluster, machines and Events.
 	watch -c -n 2 hack/output-for-watch.sh
+
+.PHONY: create-hetzner-installimage-tgz
+create-hetzner-installimage-tgz:
```
batistein, line 793: What is this market target necessary for if we hardcode the version? I would say we could remove this or change it to download the latest version of hetzner-installimage. Another thing we should change is that it's not clear which version we're currently using. We should change the filename under data/hetzner-installimage.tgz to data/hetzner-install-image-v1.0.0.tgz so it's clear which version of installimage we're currently using.

In the code where we're unpacking, it's okay to have no version information.

batistein (changes requested):

PR #1045, opened 2023-11-06
:seedling: Add event if host is provisioned

janiskemper:
Adding an event to show that the provisioning process has been completed. This is important to find out based on events where an issue happens. Even if it is successfully provisioned, it might not be added to the cluster, as the CCM does some work after CAPH. To know more about the state, this event will be helpful.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (approved):
lgtm

PR #1052, opened 2023-11-13
:seedling: Remove deprecated rate limit condition

### LINKED ISSUE #1026: Remove deprecated RateLimitExceeded condition
/kind bug

In cases where objects had RateLimitExceeded condition set and the CAPH was updated to a version where this condition is not used anymore, it was still set on "false", so that the whole cluster or machine is "not ready". 

See #1025 for an example. 

The condition should be removed automatically by the controller, as is already done in the case of other deprecated conditions. This can affect all objects that have rate limit handling and should be therefore removed in all of our custom resources where this is applicable.

janiskemper:
Removing deprecated rate limit conditions of all objects.

Fixes #1026

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (dismissed):
lgtm.

PR #1071, opened 2023-12-04
:sparkles: download BM images via oci-registry.

guettli:

Up to now, bare-metal machine images could only be downloaded from http/https URLs.

This PR make it possible to download the machine images from oci-registries.

You can use a tool like [oras](https://oras.land) to upload the machine image in tgz format to an oci registry.


janiskemper:
does it make sense to have the shell script in the Go code? Shouldn't it be rather in hack/?

guettli:
> does it make sense to have the shell script in the Go code? Shouldn't it be rather in hack/?

The shell script does not get executed like during build/compile time like the other scripts.

The shell scripts get copied from the controller intpkg/services/baremetal/client/ssh/download-from-oci.sho the rescue-system. I added a comment at the top.

I think hack/ is no good place for this script.

I could copy the script into the Go code like (instead of using embed):

```
var script=`#!/bin/bash
....
```

@janiskemper what do you think? I can copy it to hack/ if you prefer that.

janiskemper:
I think I prefer having it hard-coded in the Go code!

guettli:
> I think I prefer having it hard-coded in the Go code!

ok, I put the shell script into the Go code.

batistein (approved):

PR #1015, opened 2023-10-20
✨ update to pkgs.k8s.io

kranurag7:
this commit updates deps from apt.kuberentes.io to pkgs.k8s.io

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1013 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

PR #1073, opened 2023-12-06
:sparkles: Upgrade to Kubernetes 1.28

guettli:

Upgrade to Kubernetes 1.28

Based on https://github.com/syself/cluster-api-provider-hetzner/pull/1015


pkg/services/baremetal/host/host.go
```diff
@@ -1290,7 +1297,8 @@ func (s *Service) actionEnsureProvisioned() (ar actionResult) {
 		if err != nil {
 			if errors.Is(err, errUnexpectedHostName) {
 				// One possible reason: The machine gets used by a second wl-cluster
-				record.Warn(s.scope.HetznerBareMetalHost, "UnexpectedHostName", err.Error())
+				record.Warn(s.scope.HetznerBareMetalHost, "UnexpectedHostName",
```
janiskemper, line 1300: warnf instead of warn with sprintf

pkg/services/baremetal/host/host.go
```diff
@@ -1501,7 +1511,8 @@ func (s *Service) actionProvisioned() actionResult {
 			if err != nil {
 				if errors.Is(err, errUnexpectedHostName) {
 					// One possible reason: The machine gets used by a second wl-cluster
-					record.Warn(s.scope.HetznerBareMetalHost, "UnexpectedHostName", err.Error())
+					record.Warn(s.scope.HetznerBareMetalHost, "UnexpectedHostName",
```
janiskemper, line 1514: warnf

Tiltfile
```diff
@@ -33,7 +33,7 @@ settings = {
         "HCLOUD_REGION": "fsn1",
         "CONTROL_PLANE_MACHINE_COUNT": "3",
         "WORKER_MACHINE_COUNT": "3",
-        "KUBERNETES_VERSION": "v1.25.2",
+        "KUBERNETES_VERSION": "v1.27.3",
```
janiskemper, line 36: is this correct?

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -243,7 +243,7 @@ var _ = Describe("HetznerBareMetalHostReconciler", func() {
 			By("making sure the it has been deleted")
 			Eventually(func() bool {
 				return apierrors.IsNotFound(testEnv.Get(ctx, key, host))
-			}, timeout, time.Second).Should(BeTrue())
+			}, 10*time.Second, time.Second).Should(BeTrue())
```
janiskemper, line 246: We have talked about this already. We should find out the true reason why this is flaky. 10 seconds won't help

pkg/services/baremetal/host/host.go
```diff
@@ -45,7 +45,7 @@ import (
 )
 
 const (
-	rebootWaitTime       time.Duration = 15 * time.Second
+	rebootWaitTime       time.Duration = 30 * time.Second
```
janiskemper, line 48: are you sure that we want to increase this? This is about the time the server needs to start the reboot. AFAIK this should be done in a few seconds

janiskemper (changes requested):

Tiltfile
kranurag7: Yes, the stable version as per [kubernetes](https://www.downloadkubernetes.com/) for 1.27 is 1.27.8, but we don't have the kindest/node image with that tag. 
I used the latest tag that was present for 1.27. 
Ref: https://hub.docker.com/r/kindest/node/tags?page=1&name=v1.27

pkg/services/baremetal/host/host.go
guettli: fixed on next push.

pkg/services/baremetal/host/host.go
guettli: fixed on next push.

controllers/hetznerbaremetalhost_controller_test.go
guettli: @janiskemper I agree. I could not reproduce it. On my machine, the test was successful in 1s. I will use the old timeout again.

main.go
```diff
@@ -98,16 +98,17 @@ func main() {
...
+		HealthProbeBindAddress: probeAddr,
+		WebhookServer: webhook.NewServer(webhook.Options{
+			Port: 9443,
+		}),
 		LeaderElection:                enableLeaderElection,
 		LeaderElectionID:              "hetzner.cluster.x-k8s.io",
 		LeaderElectionNamespace:       leaderElectionNamespace,
 		LeaderElectionResourceLock:    "leases",
 		LeaderElectionReleaseOnCancel: true,
-		Namespace:                     watchNamespace,
```
janiskemper, line 110: why do you remove this? CAPI still uses it https://github.com/kubernetes-sigs/cluster-api/blob/main/main.go#L301

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
guettli: ok, I will revert my change.

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -336,13 +336,16 @@ var _ = Describe("HetznerBareMetalHostReconciler", func() {
 				}, timeout, time.Second).Should(BeNil())
 
 				Eventually(func() bool {
-					// TODO: Add logging to trace flaky unit-test.
 					if err := testEnv.Get(ctx, key, host); err != nil {
+						testEnv.GetLogger().Info("......... reaches the state image installing. Get failed", "err", err)
```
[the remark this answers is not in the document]
guettli: done.

main.go
guettli: I get:

> unknown field Namespace in struct literal of type manager.Options

From controller-runtime release page:

> cache.Options.Namespace has been removed in favor of cache.Options.Namespaces

janiskemper: yes just check out the CAPI code!

guettli: According to the [contract](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/support-multiple-instances#contract), it is required. I added it again.

janiskemper (approved):

PR #1075, opened 2023-12-06
:seedling: fix flaky test, increase timeout.

guettli:

This test was flaky. Increasing the timeout helped.

```
Summarizing 1 Failure:
  [FAIL] HetznerBareMetalHostReconciler Tests with bm machine Test root device life cycle [It] reaches the state image installing
  /home/guettli/syself/caph2/controllers/hetznerbaremetalhost_controller_test.go:347

Ran 95 of 95 Specs in 54.401 seconds
FAIL! -- 94 Passed | 1 Failed | 0 Pending | 0 Skipped

```

And decrease the timeout for tests. They take 2.5 min today. I decreased from 10min (default) to 5min, so that you don't want too long if they hang.


aniruddha2000 (approved):

kranurag7 (approved):
Thanks for including this, this was required. LGTM 👍🏻

PR #1091, opened 2023-12-15
:seedling: better error message if secret for BM is missing.

guettli:

better error message if secret for BM is missing.

### DIFF
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -218,14 +218,15 @@ func (r *HetznerBareMetalHostReconciler) getSecrets(
 		osSSHSecret, err = secretManager.ObtainSecret(ctx, osSSHSecretNamespacedName)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
+				msg := fmt.Sprintf("%s: %s", infrav1.ErrorMessageMissingOSSSHSecret, err.Error())
 				conditions.MarkFalse(
 					bmHost,
 					infrav1.CredentialsAvailableCondition,
 					infrav1.OSSSHSecretMissingReason,
 					clusterv1.ConditionSeverityError,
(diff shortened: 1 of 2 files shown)


controllers/hetznerbaremetalhost_controller.go
```diff
@@ -218,14 +218,15 @@ func (r *HetznerBareMetalHostReconciler) getSecrets(
 		osSSHSecret, err = secretManager.ObtainSecret(ctx, osSSHSecretNamespacedName)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
+				msg := fmt.Sprintf("%s: %s", infrav1.ErrorMessageMissingOSSSHSecret, err.Error())
```
janiskemper, line 221: what exactly do you intend to do? 
you change the message from
"could not find OSSSHSecret" 
to 
"could not find OSSSHSecret: not found" 
right now... Do you intend that?
Seems to not add much value

guettli: If findSecret() fails, then the error contains the namespace and the name. That's what I would like to have in the error message:

```
func (sm *SecretManager) ObtainSecret(ctx context.Context, key types.NamespacedName) (*corev1.Secret, error) {
	secret, err := sm.findSecret(ctx, key)
	if err != nil {
		return nil, fmt.Errorf("failed to fetch secret %s in namespace %s: %w", key.Name, key.Namespace, err)
	}
	err = sm.claimSecret(ctx, secret, nil, false, false)

	return secret, err
}
```

guettli: @janiskemper I think the PR improves the err msg. Could you please have a second look? Thank you

janiskemper: okay, maybe you can add this directly to the PR, because without this explanation that this error is wrapped in another place already with name and namespace it is impossible to understand.

janiskemper (approved):

PR #1103, opened 2023-12-28
✨ use CEL for CRD validation

kranurag7:
this commit uses CEL to validate CRDs and remove the associate webhook code from the repo.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper:
@kranurag7 what about this?

kranurag7:
This needs rebase to main.

kranurag7:
This needs rebase to main.

api/v1beta1/hcloudmachinetemplate_webhook.go
```diff
@@ -1,82 +0,0 @@
...
-	}
-
-	req, err := admission.RequestFromContext(ctx)
-	if err != nil {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
-	}
-
-	var allErrs field.ErrorList
-
-	if !topology.ShouldSkipImmutabilityChecks(req, newHCloudMachineTemplate) && !reflect.DeepEqual(newHCloudMachineTemplate.Spec, oldHCloudMachineTemplate.Spec) {
```
janiskemper, line 72: @kranurag7 I don't think that your current validation incorporates this feature, does it? I guess we need it though

api/v1beta1/hetznerbaremetalmachinetemplate_webhook.go
```diff
@@ -1,82 +0,0 @@
...
-	}
-
-	req, err := admission.RequestFromContext(ctx)
-	if err != nil {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
-	}
-
-	var allErrs field.ErrorList
-
-	if !topology.ShouldSkipImmutabilityChecks(req, newHetznerBareMetalMachineTemplate) && !reflect.DeepEqual(newHetznerBareMetalMachineTemplate.Spec, oldHetznerBareMetalMachineTemplate.Spec) {
```
janiskemper, line 72: same here

janiskemper (changes requested):

api/v1beta1/hcloudmachinetemplate_webhook.go
kranurag7: Yes, this is not implemented via CEL and I don't think we can as of now searching a little about the same. Why are we performing a dry-run and what are the cases where we need this? 
I think enforcement via `kubectl edit` is still in place via the PR but I'm not sure about why we have the former one here.

janiskemper:
@kranurag7 plz test the following:
Create a hcloudmachinetemplate and then update it by changing something (e.g. imageName). The webhook should not trigger in this case. However, I wonder whether the CEL validation automatically checks this! If this goes through, then do the same with an hcloudmachine object. Then paste the results here!

kranurag7:
#### hcloudmachinetemplate 
```bash
$ cat demo_hct.yaml
kind: HCloudMachineTemplate
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
metadata:
  name: test-demo-hct
spec:
  template:
    spec:
      type: cpx31
      imageName: "ubuntu-22.04"

$ kubectl apply -f demo_hct.yaml
hcloudmachinetemplate.infrastructure.cluster.x-k8s.io/test-demo-hct created

$  $ kubectl edit hcloudmachinetemplates.infrastructure.cluster.x-k8s.io test-demo-hct 

> I update type from cpx31 to cpx21 

> following is the message I get within my editor 
# hcloudmachinetemplates.infrastructure.cluster.x-k8s.io "test-demo-hct" was not valid:
# * spec.template.spec.type: Invalid value: "string": HcloudMachine.Spec.Type is immutable.
```
--- 
#### hcloudmachine 

```bash
$ cat hcm.yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: HCloudMachine
metadata:
  name: test-demo-hcm
spec:
  imageName: ubuntu-22.04
  type: cpx31
  sshKeys:
    - name: demo
$ kubectl apply -f hcm.yaml
hcloudmachine.infrastructure.cluster.x-k8s.io/test-demo-hcm created

> I used kubectl edit to update the type from cpx31 to cpx21 and it gives me the following message in my editor. 

# hcloudmachines.infrastructure.cluster.x-k8s.io "test-demo-hcm" was not valid:
# * spec.type: Invalid value: "string": HcloudMachine.Spec.Type is immutable.

```

guettli:
Since the current implementation needs to use `ShouldSkipImmutabilityChecks()` we can't use CEL validation here.

For more about this see [ShouldSkipImmutabilityChecks() docs](https://github.com/syself/cluster-api-provider-hetzner/blob/61613f3f0e8764f9470291a40c7a40bbcef4e277/vendor/sigs.k8s.io/cluster-api/util/topology/topology.go#L31)

This is needed since cluster-classes use InfraProviderMachineTemplate to create InfraProviderMachineTemplate. There is no InfraProviderMachineTemplateTemplate.

Issue #845, opened 2023-07-09
Inability to provision Cluster with an externally managed Control Plane


prometherion:
/kind bug

**What steps did you take and what happened:**

Try to create an `HetznerCluster`  resource with the following manifest:

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: HetznerCluster
metadata:
  name: workload
  namespace: default
spec:
  controlPlaneEndpoint:
    host: ""
    port: 443
  controlPlaneLoadBalancer:
    enabled: false
  controlPlaneRegions: []
  hcloudNetwork:
    enabled: false
  hcloudPlacementGroups:
  - name: md-0
    type: spread
  hetznerSecretRef:
    key:
      hcloudToken: hcloud
    name: hetzner
  sshKeys:
    hcloud:
    - name: REDACTED
```

Getting the following error:

```
The HetznerCluster "workload" is invalid: 
* spec.controlPlaneRegions: Invalid value: []v1beta1.Region{}: control plane regions must not be empty
* spec.controlPlaneEndpoint: Invalid value: :443: controlPlaneEndpoint has to be specified if controlPlaneLoadBalancer is not enabled
```

**What did you expect to happen:**

The said `Cluster` is referencing an externally managed Control Plane thanks to [Kamaji](https://kamaji.clastix.io).

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: workload
  namespace: default
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - 10.244.0.0/16
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
    kind: KamajiControlPlane
    name: workload-control-plane
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: HetznerCluster
    name: workload
```

Kamaji is offering a [CAPI Control Plane provider](https://github.com/clastix/cluster-api-control-plane-provider-kamaji) which is responsible for creating the required Control Plane endpoint, and then patching it back to the provided Infrastructure Provider.

Cluster API contract allows specifying an externally managed Control Plane, by not giving for granted it must be on the same infrastructure provider as per VM/machine.

My original plan was to take advantage of the `HetznerClusters.spec.controlPlaneLoadBalancer.enabled` field set to `false`, however, with the said error, I'm confused about how it should be used.

[...]

PR #848, opened 2023-07-11
🌱 Allow Control Plane endpoint deferral assignment

prometherion:
**What this PR does / why we need it**:

This PR drafts the support for the provisioning of Hetzner-based worker nodes using an externally managed Control Plane.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #845

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

PR #1106, opened 2024-01-03
:seedling: support externally managed control planes

rpahli:

This PR enables provisioning of Hetzner-based worker nodes using an externally managed Control Plane.

Fixes https://github.com/syself/cluster-api-provider-hetzner/issues/845

This PR is related to the PR https://github.com/syself/cluster-api-provider-hetzner/pull/848. With this code I was able to create a HetznerCluster with an external control plane and hetzner worker nodes.

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/hetznercluster_types.go
+++ b/api/v1beta1/hetznercluster_types.go
@@ -26,6 +26,9 @@ const (
 	// resources associated with HetznerCluster before removing it from the
 	// apiserver.
 	ClusterFinalizer = "hetznercluster.infrastructure.cluster.x-k8s.io"
+	// AllowEmptyControlPlaneAddressAnnotation allows HetznerCluster Webhook
+	// to skip some validation steps for external managed controle planes.
+	AllowEmptyControlPlaneAddressAnnotation = "capi.syself.com/allow-empty-control-plane-address"
 )
 
 // HetznerClusterSpec defines the desired state of HetznerCluster.
@@ -43,7 +46,7 @@ type HetznerClusterSpec struct {
 	SSHKeys HetznerSSHKeys `json:"sshKeys"`
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
-	ControlPlaneEndpoint *clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
+	ControlPlaneEndpoint *clusterv1.APIEndpoint `json:"controlPlaneEndpoint,omitempty"`
 
 	// ControlPlaneLoadBalancer is optional configuration for customizing control plane behavior. Naming convention is from upstream cluster-api project.
 	ControlPlaneLoadBalancer LoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
--- a/api/v1beta1/hetznercluster_webhook.go
+++ b/api/v1beta1/hetznercluster_webhook.go
@@ -73,7 +73,12 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
 	hetznerclusterlog.V(1).Info("validate create", "name", r.Name)
 	var allErrs field.ErrorList
 
-	if len(r.Spec.ControlPlaneRegions) == 0 {
+	allowEmptyControlPlaneAddress := false
+	if r.Annotations != nil {
+		allowEmptyControlPlaneAddress = r.Annotations[AllowEmptyControlPlaneAddressAnnotation] == "true"
+	}
+
+	if !allowEmptyControlPlaneAddress && len(r.Spec.ControlPlaneRegions) == 0 {
 		allErrs = append(allErrs, field.Invalid(
 			field.NewPath("spec", "controlPlaneRegions"),
 			r.Spec.ControlPlaneRegions,
@@ -108,8 +113,9 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
 		}
 	}
 
-	// Check whether controlPlaneEndpoint is specified if controlPlaneLoadBalancer is not enabled
-	if !r.Spec.ControlPlaneLoadBalancer.Enabled {
+	// Check whether controlPlaneEndpoint is specified if allow empty is not set or false
+
+	if !allowEmptyControlPlaneAddress && !r.Spec.ControlPlaneLoadBalancer.Enabled {
 		if r.Spec.ControlPlaneEndpoint == nil ||
 			r.Spec.ControlPlaneEndpoint.Host == "" ||
 			r.Spec.ControlPlaneEndpoint.Port == 0 {
--- a/controllers/hetznercluster_controller.go
+++ b/controllers/hetznercluster_controller.go
@@ -203,33 +203,7 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 		return reconcile.Result{}, fmt.Errorf("failed to reconcile placement groups for HetznerCluster %s/%s: %w", hetznerCluster.Namespace, hetznerCluster.Name, err)
 	}
 
-	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
-		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
-			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
-			defaultPort := int32(hetznerCluster.Spec.ControlPlaneLoadBalancer.Port)
-
-			if hetznerCluster.Spec.ControlPlaneEndpoint == nil {
-				hetznerCluster.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
-					Host: defaultHost,
-					Port: defaultPort,
-				}
-			} else {
-				if hetznerCluster.Spec.ControlPlaneEndpoint.Host == "" {
-					hetznerCluster.Spec.ControlPlaneEndpoint.Host = defaultHost
-				}
-				if hetznerCluster.Spec.ControlPlaneEndpoint.Port == 0 {
-					hetznerCluster.Spec.ControlPlaneEndpoint.Port = defaultPort
-				}
-			}
-
-			hetznerCluster.Status.Ready = true
-		}
-	} else if hetznerCluster.Spec.ControlPlaneEndpoint != nil {
-		hetznerCluster.Status.Ready = true
-	}
-
-	// update control plane endpoint
-	hetznerCluster.Status.Ready = setControlPlaneEndpoint(hetznerCluster)
+	processControlPlaneEndpoint(hetznerCluster)
 
 	// delete deprecated conditions of old clusters
 	conditions.Delete(clusterScope.HetznerCluster, infrav1.DeprecatedHetznerClusterTargetClusterReadyCondition)
@@ -267,8 +241,7 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	return reconcile.Result{}, nil
 }
 
-// setControlPlaneEndpoint updates hetznerCluster.Spec.ControlPlaneEndpoint and returns whether the hetznerCluster is ready or not.
-func setControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) bool {
+func processControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) {
 	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
 		if hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
 			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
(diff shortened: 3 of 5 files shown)


api/v1beta1/hetznercluster_types.go
```diff
@@ -43,7 +46,7 @@ type HetznerClusterSpec struct {
 	SSHKeys HetznerSSHKeys `json:"sshKeys"`
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
-	ControlPlaneEndpoint *clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
+	ControlPlaneEndpoint *clusterv1.APIEndpoint `json:"controlPlaneEndpoint,omitempty"`
```
janiskemper, line 49: because of your API change you probably have to run "make generate"

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -73,7 +73,12 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
 	hetznerclusterlog.V(1).Info("validate create", "name", r.Name)
 	var allErrs field.ErrorList
 
-	if len(r.Spec.ControlPlaneRegions) == 0 {
+	allowEmptyControlPlaneAddress := "false"
```
janiskemper, line 76: better: make allowEmptyControlPlaneAddress a boolean and set it on true if the annotation is set

controllers/hetznercluster_controller.go
```diff
@@ -203,7 +203,9 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 		return reconcile.Result{}, fmt.Errorf("failed to reconcile placement groups for HetznerCluster %s/%s: %w", hetznerCluster.Namespace, hetznerCluster.Name, err)
 	}
 
-	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
+	if !hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled && (hetznerCluster.Spec.ControlPlaneEndpoint == nil || hetznerCluster.Spec.ControlPlaneEndpoint.Host == "") {
+		clusterScope.Logger.Info("disabled LoadBalancer and not yet provided ControlPlane endpoint")
```
janiskemper, line 207: can you add a new condition and set it on false here? 
Maybe something like "ControlPlaneEndpointSet". You can then set it on true or false in these cases below. 
If the cluster is usable, it should be always on true AFAIK.

janiskemper (changes requested):

janiskemper:
Thanks for the PR! Looking forward to this!

api/v1beta1/hetznercluster_types.go
janiskemper: okay! Never mind then!

rpahli:
@janiskemper I updated the PR, can you pls have a look.

janiskemper (commented):
looks good to me!
This new annotation together with
a) the webhook changes and
b) the flow about the new condition being set to true or false
can be tested very well with unit tests. 

Are you able to contribute to the tests as well?

rpahli:
ok I added some tests :-)

controllers/hetznercluster_controller_test.go
```diff
@@ -443,6 +443,44 @@ var _ = Describe("Hetzner ClusterReconciler", func() {
 					return isPresentAndFalseWithReason(key, instance, infrav1.LoadBalancerReadyCondition, infrav1.LoadBalancerFailedToOwnReason)
 				}, timeout, time.Second).Should(BeTrue())
 			})
+
+			It("should work with capi.syself.com/allow-empty-control-plane-address annotation error condition", func() {
+				instance.Annotations = make(map[string]string)
+				instance.Annotations[infrav1.AllowEmptyControlPlaneAddressAnnotation] = "true"
+				instance.Spec.ControlPlaneLoadBalancer.Enabled = false
+				instance.Spec.ControlPlaneEndpoint = &clusterv1.APIEndpoint{
```
janiskemper, line 451: I believe that this should be nil right? :D

controllers/hetznercluster_controller_test.go
```diff
@@ -778,6 +816,16 @@ var _ = Describe("HetznerCluster validation", func() {
 			Expect(testEnv.Create(ctx, hetznerCluster)).To(Succeed())
 		})
 
+		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation", func() {
```
janiskemper, line 819: I think we should have two tests here, which reflect the two changes you made in the webhooks!

janiskemper (changes requested):
looks good so far. We need to do some changes, but this definitely goes into the right direction!

controllers/hetznercluster_controller_test.go
rpahli: yes and no it should work in both cases because the Kamaji operator uses the port to configure the loadbalancer.

controllers/hetznercluster_controller_test.go
rpahli: ok I'll extend the tests

controllers/hetznercluster_controller_test.go
rpahli: added both tests

janiskemper: I'm sure that it should work, but this current setup works even without the flag you introduced.. What you want to test is that both Loadbalancer and ControlPlaneEndpoint are not set. Then it should only work with the flag (and the condition should be set on false accordingly).
I believe currently, your condition would be also on true, no? Maybe I'm missing something but that's what I understand from https://github.com/syself/cluster-api-provider-hetzner/pull/1106/files#diff-6ec10baeb9341142d245628229b63ac327258d210a8cf290fb19771006046aafR236

rpahli: ok I refactored the code a bit because to me it looked that there are some duplicated code snippets. There is one tests case which I can't create because the controller and the fake client are maybe to fast.

Loadbalancer enabled but no loadbalancer ready yet should create a false condition

rpahli: nevermind found a way

controllers/hetznercluster_controller_test.go
janiskemper: I still think that we should have two or three tests here:
a) check that it works for empty regions if you specify the annotation
b) check that it works without load balancer and without controlplaneendpoint
c) check that it works without load balancer and without controlplaneendpoint.host

controllers/hetznercluster_controller_test.go
janiskemper: oh that's a good catch! there was indeed duplicate code! thanks a lot for finding this!

controllers/hetznercluster_controller_test.go
rpahli: ok I added the test cases

janiskemper: Sorry, I think there was a misunderstanding... I was talking about tests of the webhook server, which is what we do in the "HetznerCluster validation" section. I thought the other tests in the other sections (which you touched in your latest commit) were already completed...

rpahli: ok I reverted the last commit and added the new tests :-)

janiskemper: nice!

controllers/hetznercluster_controller_test.go
```diff
@@ -443,6 +443,78 @@ var _ = Describe("Hetzner ClusterReconciler", func() {
...
+				Eventually(func() bool {
+					if err := testEnv.Get(ctx, key, instance); err != nil {
+						return false
+					}
+
+					return isPresentAndFalseWithReason(key, instance, infrav1.ControlPlaneEndpointSetCondition, infrav1.ControlPlaneEndpointNotSetReason)
+				}, timeout, time.Second).Should(BeTrue())
+			})
+
+			It("should work with capi.syself.com/allow-empty-control-plane-address annotation error condition custom port", func() {
```
janiskemper, line 463:
```suggestion
			It("should set condition on false with capi.syself.com/allow-empty-control-plane-address annotation, disabled load balancer and empty controlPlaneEndpoint.host", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -443,6 +443,78 @@ var _ = Describe("Hetzner ClusterReconciler", func() {
 					return isPresentAndFalseWithReason(key, instance, infrav1.LoadBalancerReadyCondition, infrav1.LoadBalancerFailedToOwnReason)
 				}, timeout, time.Second).Should(BeTrue())
 			})
+
+			It("should work with capi.syself.com/allow-empty-control-plane-address annotation error condition", func() {
```
janiskemper, line 447:
```suggestion
			It("should set condition on false with capi.syself.com/allow-empty-control-plane-address annotation, disabled load balancer and nil controlPlaneEndpoint", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -443,6 +443,78 @@ var _ = Describe("Hetzner ClusterReconciler", func() {
...
+				Eventually(func() bool {
+					if err := testEnv.Get(ctx, key, instance); err != nil {
+						return false
+					}
+
+					return isPresentAndFalseWithReason(key, instance, infrav1.ControlPlaneEndpointSetCondition, infrav1.ControlPlaneEndpointNotSetReason)
+				}, timeout, time.Second).Should(BeTrue())
+			})
+
+			It("should work with capi.syself.com/allow-empty-control-plane-address annotation success condition", func() {
```
janiskemper, line 482:
```suggestion
			It("should set condition on true with capi.syself.com/allow-empty-control-plane-address, disabled load balancer and specified controlPlaneEndpoint", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -443,6 +443,78 @@ var _ = Describe("Hetzner ClusterReconciler", func() {
...
+				Eventually(func() bool {
+					if err := testEnv.Get(ctx, key, instance); err != nil {
+						return false
+					}
+
+					return isPresentAndTrue(key, instance, infrav1.ControlPlaneEndpointSetCondition)
+				}, timeout, time.Second).Should(BeTrue())
+			})
+
+			It("should work with enabled load balancer success", func() {
```
janiskemper, line 501:
```suggestion
			It("should set condition on true with capi.syself.com/allow-empty-control-plane-address, enabled load balancer and specified controlPlaneEndpoint", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -778,6 +850,35 @@ var _ = Describe("HetznerCluster validation", func() {
...
+			hetznerCluster.Annotations = make(map[string]string)
+			hetznerCluster.Annotations[infrav1.AllowEmptyControlPlaneAddressAnnotation] = "true"
+			hetznerCluster.Spec.ControlPlaneRegions = []infrav1.Region{}
+			hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled = false
+			hetznerCluster.Spec.ControlPlaneEndpoint.Port = 443
+			hetznerCluster.Spec.ControlPlaneEndpoint.Host = "localhost"
+			Expect(testEnv.Create(ctx, hetznerCluster)).To(Succeed())
+		})
+
+		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation empty host", func() {
```
janiskemper, line 863:
```suggestion
		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation, disabled load balancer and empty host", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -778,6 +850,35 @@ var _ = Describe("HetznerCluster validation", func() {
 			Expect(testEnv.Create(ctx, hetznerCluster)).To(Succeed())
 		})
 
+		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation", func() {
```
janiskemper, line 853:
```suggestion
		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation, disabled load balancer and empty controlPlaneRegions", func() {
```

controllers/hetznercluster_controller_test.go
```diff
@@ -778,6 +850,35 @@ var _ = Describe("HetznerCluster validation", func() {
...
+			hetznerCluster.Annotations = make(map[string]string)
+			hetznerCluster.Annotations[infrav1.AllowEmptyControlPlaneAddressAnnotation] = "true"
+			hetznerCluster.Spec.ControlPlaneRegions = []infrav1.Region{}
+			hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled = false
+			hetznerCluster.Spec.ControlPlaneEndpoint.Port = 443
+			hetznerCluster.Spec.ControlPlaneEndpoint.Host = ""
+			Expect(testEnv.Create(ctx, hetznerCluster)).To(Succeed())
+		})
+
+		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation empty ControlPlaneEndpoint", func() {
```
janiskemper, line 873:
```suggestion
		It("should succeed with capi.syself.com/allow-empty-control-plane-address annotation, disabled load balancer and nil ControlPlaneEndpoint", func() {
```

janiskemper (changes requested):

janiskemper:
Cool! Can you run `make verify` and `make test-unit` to see whether everything works? 

Then we can squash and rebase.

After that, this will be good to merge!

rpahli:
ok I run `make verify` and `make test-unit` and all looks good. 

One question from my side some python scripts use `#!/usr/bin/env python`. I changed it to `#!/usr/bin/env python3` locally. Is there a plan to migrate als scripts to `python3`?

janiskemper:
@kranurag7 what about python3?

controllers/hetznercluster_controller.go
```diff
@@ -267,9 +241,9 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	return reconcile.Result{}, nil
 }
 
-// setControlPlaneEndpoint updates hetznerCluster.Spec.ControlPlaneEndpoint and returns whether the hetznerCluster is ready or not.
-func setControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) bool {
+func processControlPlaneEndpoint(hetznerCluster *infrav1.HetznerCluster) {
 	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
+		// setControlPlaneEndpoint updates hetznerCluster.Spec.ControlPlaneEndpoint and returns whether the hetznerCluster is ready or not.
```
janiskemper, line 246: I think this comment got to the wrong line. you can probably remove it

janiskemper (changes requested):

controllers/hetznercluster_controller.go
rpahli: done

janiskemper (approved):

janiskemper:
thanks a lot, @rpahli :)
I really appreciate your work!

prometherion:
Thanks for the efforts put in place by @rpahli and all the reviewers, I'm happy to see the use-case considered relevant and implemented! 🤗

guettli:
@rpahli @prometherion We merged your PR some time ago. Today we want to document all our Annotations. To be honest, I am currently unable to create a good documentation for the annotation introduced by your PR. Do you have docs for tools (afaik Kamaji) which use that annotation? How could we document the annotation?

BTW: I am happy that the solution works for you. But I am thinking: Why not create a field in CAPI which toggles that? Or do you think it should be the default for infra-providers to make that field optional?

prometherion:
The idea of the annotation was to preserve the current behaviour, such as a mandatory Control Plane, and the annotation has been picked up to avoid changing the specification.

We documented the required steps [here](https://github.com/clastix/cluster-api-control-plane-provider-kamaji/blob/master/docs/providers-hetzner.md), not sure if this could be helpful for you, but the idea is pretty straightforward.

By enabling that annotation, the Cluster API infrastructure provider doesn't complain about a missing Load Balancer and thus it waits for an eventual `controlPlaneEndpoint` assignment, by a Control Plane provider (e.g.: Kamaji) or by a human.

Not sure if I helped here, happy to review your changes tho to document it as clearly as possible!

guettli:
@prometherion how do other capi  infra providers handle that? Do they need such an annotation, too? Or do they allow the field to be empty by default? 

Maybe it's the best solution to remove the need for that annotation (making that field optimal in caph)

I guess you would welcome that.

rpahli:
@guettli unfortunately there is no common way to handle this behavior. But having an optional field would be the better way I think. For capz we use a field named `ControlPlaneEnabled`. Maybe we can use the same name for caph as well.

guettli:
@rpahli this should be solved at capi level, not at infra provider level. Please ask the capi people. If you do, please leave a link here. Thank you

PR #1110, opened 2024-01-14
:seedling: Create Condition if DownloadImage failed.

guettli:

If downloading the image on a bare-metal server failed, it was up to now only visible in the logs.

This PR creates a status.condition.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -1021,7 +1021,15 @@ func (s *Service) createAutoSetupInput(sshClient sshclient.Client) (autoSetupInp
 	if needsDownload {
 		out := sshClient.DownloadImage(imagePath, image.URL)
 		if err := handleSSHError(out); err != nil {
-			return autoSetupInput{}, actionError{err: fmt.Errorf("failed to download image: %s %s %w", out.StdOut, out.StdErr, err)}
+			err := fmt.Errorf("failed to download image: %s %s %w", out.StdOut, out.StdErr, err)
+			conditions.MarkFalse(
+				s.scope.HetznerBareMetalHost,
+				infrav1.ProvisionSucceededCondition,
+				infrav1.ImageDownloadFailedReason,
+				clusterv1.ConditionSeverityError,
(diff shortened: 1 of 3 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -1021,7 +1021,15 @@ func (s *Service) createAutoSetupInput(sshClient sshclient.Client) (autoSetupInp
 	if needsDownload {
 		out := sshClient.DownloadImage(imagePath, image.URL)
 		if err := handleSSHError(out); err != nil {
-			return autoSetupInput{}, actionError{err: fmt.Errorf("failed to download image: %s %s %w", out.StdOut, out.StdErr, err)}
+			err := fmt.Errorf("failed to download image: %s %s %w", out.StdOut, out.StdErr, err)
+			conditions.MarkFalse(
+				s.scope.HetznerBareMetalHost,
+				infrav1.ProvisionSucceededCondition,
+				infrav1.ImageSpecInvalidReason,
```
janiskemper, line 1028: this reason probably is not very precise looking at the function "downloadImage". Can you maybe list the reasons you know of for this to fail? Does it maybe even make sense to analyze the error?

pkg/services/baremetal/host/host.go
```diff
@@ -1021,7 +1021,15 @@ func (s *Service) createAutoSetupInput(sshClient sshclient.Client) (autoSetupInp
...
+				clusterv1.ConditionSeverityError,
```
janiskemper, line 1029: this is only used up to now when there is an error that cannot be recovered. Is that the case here?

pkg/services/baremetal/host/host.go
guettli: I created ImageDownloadFailedReason for this condition.

pkg/services/baremetal/host/host.go
guettli: Let's do the distinction between permanent and temporary error later.

janiskemper (approved):

PR #1111, opened 2024-01-14
:seedling: Get BM machine image via oci token

guettli:

Get BM machine image via oci token.

Before the PR $OCI_REGISTRY_AUTH_TOKEN needed to be "username:password"

How to test:

extract the shell script from the go code and use it to download an image.

for example: ghcr.io/syself/autopilot/node-images/staging/hetzner-apalla-1-27-workeramd64baremetal:v0-sha.2ae2207

### DIFF
--- a/.lychee.toml
+++ b/.lychee.toml
@@ -77,3 +77,7 @@ exclude_loopback = true
 
 # Exclude all mail addresses from checking
 exclude_mail = false
+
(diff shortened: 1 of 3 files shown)


janiskemper:
can you get a review from someone else? I am not the expert here, unfortunately!

lucasrattz (approved):
Looks good to me

PR #1130, opened 2024-02-01
:bug: Fix wrong names of storage devices

janiskemper:
Names of storage devices might change after reboot. We stored the names in the host object and didn't update them anymore. However, these names might change after a reboot and might even switch between devices.

This led to the issue that the wwn that the users give as reference to the storage device that is used in provisioning did not correspond anymore to the name of the device that we stored in the host object.

This led to the operator using the wrong storage device in provisioning.

Now we get the latest information of wwn and storage device names before installimage and thus always take the correct storage device for provisioning.

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (approved):
lgtm.

batistein (approved):

PR #1154, opened 2024-02-06
🐛 Fix server deletion

lucasrattz:
This PR fixes the issue of deleting servers introduced by [PR 1145](https://github.com/syself/cluster-api-provider-hetzner/pull/1145).

The server was being removed as a target from the load balancer during the load balancer reconciliation, and the server deletion function was trying to do that removal again and failing. Now there is a check, and only servers registered as targets are removed.

**TODOs**:

- [x] squash commits

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -202,8 +202,13 @@ func (s *Service) Delete(ctx context.Context) (res reconcile.Result, err error) 
 
 	// control planes have to be deleted as targets of server
 	if s.scope.IsControlPlane() && s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
-		if err := s.deleteServerOfLoadBalancer(ctx, server); err != nil {
-			return reconcile.Result{}, fmt.Errorf("failed to delete attached server of loadbalancer: %w", err)
+		for _, target := range s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.Target {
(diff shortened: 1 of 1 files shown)


guettli (approved):
lgtm.

janiskemper:
this is a valid improvement because we only call the Hetzner API if necessary. However, I know that you did it to fix the deletion process and I would say by looking at the code that there shouldn't be anything to fix. We have already a check for the error code that a target is not existing (https://github.com/syself/cluster-api-provider-hetzner/blob/a5540262345ea1ca441825f5850b7c74aba24257/pkg/services/hcloud/server/server.go#L616).

Which error did you get so that deletion was blocked?

lucasrattz:
> this is a valid improvement because we only call the Hetzner API if necessary. However, I know that you did it to fix the deletion process and I would say by looking at the code that there shouldn't be anything to fix. We have already a check for the error code that a target is not existing (
> 
> https://github.com/syself/cluster-api-provider-hetzner/blob/a5540262345ea1ca441825f5850b7c74aba24257/pkg/services/hcloud/server/server.go#L616
> 
> ).
> Which error did you get so that deletion was blocked?

The machine was hanging in Deleting state. The node was already drained but registered, and it was already removed as a target from the LB.

janiskemper:
@batistein @lucasrattz if you were not able to reproduce the bug, can you at least rename the PR title and description? AFAIK this doesn't fix any bug because there is none but rather saves some unnecessary API calls

PR #1162, opened 2024-02-13
:seedling: Skip machines with one WWN, if RAID is desired.

### LINKED ISSUE #1122: Controller should not pick hbmh with one disk, if swraid=true
/kind bug

if the hetzner-bare-metal-machine-template has swraid=true, then the controller should not pick a hbmh which has only one disk.

Up to now the controller picks the hbmh and after several minutes the process fails because installimage fails. 

You can see the error message of installimage in the Kubernetes event, but it is not immediately visible.

I expect the controller to ignore machines which have only one WWN in the hbmh, if the hetzner-bare-metal-machine-template has swraid=true.

We should do the opposite check, too:

If the  hetzner-bare-metal-machine-template has swraid=false, then the controller should not pick a hbmh which has two WWNs.

[hbmh docs](https://github.com/syself/cluster-api-provider-hetzner/blob/ad6342c684236dda19d85fd12a06066dbb46ed76/docs/reference/hetzner-bare-metal-host.md?plain=1#L33) 

[hbmmt docs](https://github.com/syself/cluster-api-provider-hetzner/blob/ad6342c684236dda19d85fd12a06066dbb46ed76/docs/reference/hetzner-bare-metal-machine-template.md?plain=1#L47)

Implementation: 

There are two lists of WWNs: rootDeviceHints.raid.wwn and rootDeviceHints.wwn.

The controller should only take a machine wher swraid (from the hbmmt) matches the WWN of hbmh.

guettli:

Skip machines with one WWN, if RAID is desired.

Fixes #1122 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -29,6 +29,8 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -283,19 +285,19 @@ func (s *Service) associate(ctx context.Context) error {
 	}
 
 	// choose new host
-	host, helper, err := s.chooseHost(ctx)
+	host, helper, reason, err := s.chooseHost(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to choose host: %w", err)
 	}
 	if host == nil {
 		s.scope.BareMetalMachine.Status.Phase = clusterv1.MachinePhasePending
-		s.scope.V(1).Info("No available host found. Requeuing.")
+		s.scope.Info("No available host found. Requeuing.", "reason", reason)
 		conditions.MarkFalse(
 			s.scope.BareMetalMachine,
 			infrav1.HostAssociateSucceededCondition,
 			infrav1.NoAvailableHostReason,
 			clusterv1.ConditionSeverityWarning,
-			"no available host",
+			fmt.Sprintf("no available host (%s)", reason),
 		)
 		return &scope.RequeueAfterError{RequeueAfter: requeueAfter}
 	}
@@ -363,7 +365,12 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
 	return &host, helper, nil
 }
 
-func (s *Service) chooseHost(ctx context.Context) (*infrav1.HetznerBareMetalHost, *patch.Helper, error) {
+// chooseHost tries to find a free hbmh.
+// If no hbmh was found, then hbmh and err are nil, and the string
+// "reason" contains human readable details.
+func (s *Service) chooseHost(ctx context.Context) (
+	hbmh *infrav1.HetznerBareMetalHost, patchHelper *patch.Helper, reason string, err error,
+) {
 	// get list of hosts scoped to namespace of machine
 	hosts := infrav1.HetznerBareMetalHostList{}
 	opts := &client.ListOptions{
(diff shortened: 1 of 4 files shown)


janiskemper (commented):
Is this something that happens often? 
This should be documented accordingly. Right now people probably try to do this and wonder why the host does not get chosen. 

Should we also include a log or event for this?

guettli:
> Is this something that happens often? This should be documented accordingly. Right now people probably try to do this and wonder why the host does not get chosen.
> 
> Should we also include a log or event for this?

in 88bd7188 i added a "reason", so that users understand why hosts were not picked.

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,77 +366,121 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
 		}
+		if host.Spec.Status.ProvisioningState != infrav1.StateNone {
+			continue
+		}
+
+		// from now on each "continue" should add an entry
+		// to mapOfSkipReasons.
+		usableDevices++
+
+		if host.Spec.RootDeviceHints != nil &&
```
janiskemper, line 411: this code block has to come last. If the host is in maintenance mode, then getting this reason is misleading

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,77 +366,121 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
+					continue
+				}
+			}
+		}
 		if host.Spec.MaintenanceMode != nil && *host.Spec.MaintenanceMode {
+			mapOfSkipReasons["hbmh-in-maintenance-mode"]++
 			continue
 		}
 		if host.GetDeletionTimestamp() != nil {
+			mapOfSkipReasons["hbmh-has-deletion-timestamp"]++
```
janiskemper, line 439: put this on top, this is a very strong reason (after "usableDevices")

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,77 +366,121 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
 			continue
 		}
 		if host.Spec.Status.ErrorMessage != "" {
+			mapOfSkipReasons["hbmh-has-error-message-in-status"]++
```
janiskemper, line 443: put this after deletion timestamp

janiskemper (changes requested):

pkg/services/baremetal/baremetal/baremetal.go
guettli: makes sense. Done in 40f22534

pkg/services/baremetal/baremetal/baremetal.go
guettli: @janiskemper please take a look at the new code. Does it make sense now? I moved "label" up, because I think we should not look too deep in to machines which are not in our scope.

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,71 +365,126 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
 	}
 
 	labelSelector := s.getLabelSelector()
 
-	availableHosts := make([]*infrav1.HetznerBareMetalHost, 0, len(hosts.Items))
+	usableHosts := make([]*infrav1.HetznerBareMetalHost, 0, len(hosts.Items))
 
+	mapOfSkipReasons := make(map[string]int)
+
+	availableDevices := 0
```
janiskemper, line 390: hosts not devices, and don't mix available and usable.. You renamed something from "available" to "usable" but here you name your variable "available" again. We should stick to one!

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,71 +365,126 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
 			continue
 		}
 
-		if !labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
-			continue
+		// This comes last, because now we would choose the machine, but we check
+		// if the config is correct.
+		if host.Spec.RootDeviceHints != nil &&
+			host.Spec.RootDeviceHints.IsValid() {
+			// If there are no RootDeviceHints, then the it should be ok that the hbmh gets picked.
```
janiskemper, line 434: better text:
Even if there are no rootDeviceHints specified, the host is usable. After the phase registering, the process to provision the server stops and waits for the user to specify the rootDeviceHints. 
Here we want to check whether the specified rootDeviceHints fit with the InstallImage configuration of the HetznerBareMetalMachine. If not, it is not usable.

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -363,71 +365,126 @@ func (s *Service) getAssociatedHost(ctx context.Context) (*infrav1.HetznerBareMe
...
+					continue
+				}
+			}
 		}
 
-		if host.Spec.Status.ProvisioningState != infrav1.StateNone {
+		usableHosts = append(usableHosts, &hosts.Items[i])
+	}
+
+	reasons := make([]string, 0, len(mapOfSkipReasons))
```
janiskemper, line 459: can you put this in a utility function that outputs a string ("reasons")?

pkg/services/baremetal/baremetal/baremetal_test.go
```diff
@@ -208,6 +234,9 @@ var _ = Describe("chooseHost", func() {
 		Hosts            []client.Object
 		HostSelector     infrav1.HostSelector
 		ExpectedHostName string
+		ExpectedReason   string
+		Swraid           int
```
janiskemper, line 238: can you define a separate struct for the test case below? There is no overlap in the properties you need  expect for "Hosts" and "ExpectedHostName". I think it is cleaner if we have two structs

pkg/services/baremetal/baremetal/baremetal_test.go
```diff
@@ -276,6 +306,42 @@ var _ = Describe("chooseHost", func() {
 				ExpectedHostName: "hostWithLabel",
 			}),
 	)
+	DescribeTable("chooseHost(): Test empty result becaues RAID config does not match.",
```
janiskemper, line 309:
```suggestion
	DescribeTable("chooseHost(): Test empty result because RAID config does not match.",
```

janiskemper (changes requested):

pkg/services/baremetal/baremetal/baremetal.go
guettli: We have three sets:

set 1: the big list which contains all devices.

set 2: Subset of set1. we removed the devices which are in use. This is called "available"

set 3: Subset of set2. We removed the one that don't match (like raid is wrong, labels don't match). This is called "usable".

I changed the name because I think it makes sense that there are more available devices than usable devices.

Does this make sense to you? Should I add a comment to the code?

pkg/services/baremetal/baremetal/baremetal.go
guettli: yes, makes sense. Does the text of 1106b11b fit?

pkg/services/baremetal/baremetal/baremetal.go
janiskemper: Ah I actually did not understand it. Maybe you can define "available" then before "usable" and quickly comment on it. 
However, it is still not "devices" but "hosts" ;)

pkg/services/baremetal/baremetal/baremetal.go
guettli: yes, makes sense. done in ab75bb93

pkg/services/baremetal/baremetal/baremetal_test.go
guettli: done in d286906e. I use unexported field names in the new struct, since there is no need to export them. ok?

pkg/services/baremetal/baremetal/baremetal_test.go
guettli: fixed.

guettli:
@janiskemper please review

janiskemper:
did you do more than fixing the test? If not, I'll approve

batistein (approved):

janiskemper (approved):

PR #1164, opened 2024-02-14
:bug: Script to detect invalid WWN/RAID configuration of hbmh.

guettli:
If the WWN configuration of a hbmh is invalid, this PR detects that and creates a corresponding Event and Condition.

We think about doing that in Go with [go-diskfs](https://github.com/diskfs/go-diskfs)

Update: We will stick to the shell script. It works.

### DIFF
--- a/pkg/services/baremetal/client/ssh/ssh_client.go
+++ b/pkg/services/baremetal/client/ssh/ssh_client.go
@@ -20,6 +20,7 @@ package sshclient
 import (
 	"bufio"
 	"bytes"
+	_ "embed"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -34,6 +35,9 @@ import (
 const (
 	sshTimeOut time.Duration = 5 * time.Second
 )
+
+//go:embed detect-linux-on-another-disk.sh
+var detectLinuxOnAnotherDiskShellScript string
 
 var downloadFromOciShellScript = `#!/bin/bash
 
@@ -200,6 +204,7 @@ type Client interface {
 	CleanCloudInitInstances() Output
 	ResetKubeadm() Output
 	UntarTGZ() Output
+	DetectLinuxOnAnotherDisk(sliceOfWwns []string) Output
 }
 
 // Factory is the interface for creating new Client objects.
@@ -480,6 +485,13 @@ func (c *sshClient) ResetKubeadm() Output {
 	echo ========= done =========
 `)
 	return output
+}
+
+func (c *sshClient) DetectLinuxOnAnotherDisk(sliceOfWwns []string) Output {
+	return c.runSSH(fmt.Sprintf(`cat <<'EOF' | bash -s -- %s
+%s
+EOF
+`, strings.Join(sliceOfWwns, " "), detectLinuxOnAnotherDiskShellScript))
 }
 
 func (c *sshClient) UntarTGZ() Output {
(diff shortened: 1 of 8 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -951,14 +953,54 @@ func (s *Service) actionImageInstalling() actionResult {
...
+			infrav1.SSHToRescueSystemFailedReason,
+			clusterv1.ConditionSeverityInfo,
+			msg,
+		)
+		record.Event(s.scope.HetznerBareMetalHost, infrav1.SSHToRescueSystemFailedReason, msg)
+		return actionContinue{
+			delay: 10 * time.Second,
+		}
+	}
+	conditions.MarkTrue(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition)
```
janiskemper, line 994: did the provisioning really succeed here?

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
guettli: @janiskemper 

no, but if you don't set it to OK here, then the error might be visibly longer than it is there. I removed that line. I think creating a custom condition is not needed, and if the err is visibly longer than it exists is not a big issue.

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -200,6 +204,7 @@ type Client interface {
 	CleanCloudInitInstances() Output
 	ResetKubeadm() Output
 	UntarTGZ() Output
+	DetectLinuxOnOtherDisk(sliceOfWwns []string) Output
```
janiskemper, line 207: grammatically better: DetectLinuxOnAnotherDisk

janiskemper (commented):
small grammar improvement. Otherwise looks good

janiskemper:
there are linting issues. I think you accidentally exported your variable

guettli:
> there are linting issues. I think you accidentally exported your variable

fixed

test/e2e/Makefile
```diff
@@ -37,7 +37,7 @@ export GOBIN := $(abspath $(TOOLS_BIN_DIR))
 
 ginkgo: $(GINKGO)
 $(GINKGO): # Build ginkgo from tools folder.
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.11.0
+	go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo
```
janiskemper, line 40: it this intended? (I don't know whether it makes sense, I'm literally asking)

guettli: strange. This is from a PR which was merged by Anurag. The new (green) line is the one which is correct. I did a `git pull origin/main` to get the latest changes. Don't ask me why github shows it in the diff.

janiskemper: Maybe because you merged the main branch in here or something? Can you clean up the PR and squash everything, then we can see again. But it shouldn't show like this if all git stuff worked correctly

guettli: @janiskemper I rebase on main again, and now the change at `-mod=mod` is gone. I think there is no need to wait for the e2e tests to finish. Ready for merge?

janiskemper: I would like them to finish, yes, in case that this is messing up the provisioning process, they are very valid!

janiskemper: problem is again that this was not in draft, so you triggered with e2e tests multiple times.. they don't get canceled if you push again on the same PR

guettli: e2e tests were successful.

aniruddha2000 (approved):

PR #1196, opened 2024-03-06
:seedling: Several small changes to make debugging bare-metal provisioning easier

### LINKED ISSUE #1174: Condition for missing `rootDeviceHints` of `HetznerBareMetalHost`
/kind feature

The user has to specify `rootDeviceHints` of `HetznerBareMetalHosts`. Currently, no condition shows the user that `rootDeviceHints` are still missing. 

I propose to add a condition that shows whether rootDeviceHints are existing (and valid) or not. This condition can be set on true or false independent of the provisioning of a machine. It is maybe necessary to distinguish `rootDeviceHints` that are set from those that are validated, because if the controller hasn't gotten the information about the storage devices from the machine yet, then the `rootDeviceHints` cannot be validated yet.

guettli:

Add Condition for missing rootDeviceHints of HetznerBareMetalHost

Better logging: Show URL of machine image (with password of URL redacted).

Update filter-caph-controller-manager-logs.py to cut very long output.

Adapt detect-linux-on-another-disk.sh to work on Ubuntu 20.04 and on machines with empty disks (no partitions yet).

Create events for InstallingMachineImageStarted to provide better hints about what is going on (and easier way to see how long these steps take).

### DIFF
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -57,11 +57,19 @@ type RootDeviceHints struct {
 
 // IsValid checks whether rootDeviceHint is valid.
 func (rdh *RootDeviceHints) IsValid() bool {
-	if rdh.WWN == "" && len(rdh.Raid.WWN) == 0 ||
-		rdh.WWN != "" && len(rdh.Raid.WWN) > 0 {
-		return false
+	return rdh.IsValidWithMessage() == ""
+}
+
+// IsValidWithMessage checks whether rootDeviceHint is valid.
+// If valid, an empty string gets returned.
+func (rdh *RootDeviceHints) IsValidWithMessage() string {
+	if rdh.WWN == "" && len(rdh.Raid.WWN) == 0 {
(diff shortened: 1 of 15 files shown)


guettli:
@janiskemper I started with a PR. This gets called for hbmhs which are in-use. But a hbmh which is not in-use seems to be ignored. I guess Reconcile() does not get called.

janiskemper:
@guettli we usually create draft PRs. Since your linting still fails because of spelling mistakes, I would have thought that a draft PR is more appropriate.

guettli:
> This gets called for hbmhs which are in-use. But a hbmh which is not in-use seems to be ignored. I guess Reconcile() does not get called.

guettli:
> @guettli we usually create draft PRs. Since your linting still fails because of spelling mistakes, I would have thought that a draft PR is more appropriate.

The PR is a draft now.

@janiskemper what about this:

The conditions gets called for hbmhs which are in-use. But a hbmh which is not in-use seems to be ignored. I guess Reconcile() does not get called.

Is it ok like this, or should the new method be called for hbmhs which are not in use, too?

janiskemper:
yes, definitely and especially for the ones that are not in use!

janiskemper (approved):

PR #1217, opened 2024-03-18
:seedling: detect degraded mdraid

### LINKED ISSUE #1214: Create Event+Condition if RAID has issues
/kind feature

If there are issues with the raid, then the controller should create an Event and a Condition.

Example: Both (md0 and md1) are missing one drive (`[2/1]`).

The controller should tell us about this.

```
Personalities : [raid1] [linear] [multipath] [raid0] [raid6] [raid5] [raid4] [raid10] 
md0 : active raid1 nvme1n1p2[1]
      1046528 blocks super 1.2 [2/1] [_U]
      
md1 : active raid1 nvme1n1p3[1]
      498400576 blocks super 1.2 [2/1] [_U]
      bitmap: 4/4 pages [16KB], 65536KB chunk

unused devices: <none>
```

guettli:

detect degraded mdraid

Fixes #1214

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -184,6 +184,13 @@ const (
 )
 
 const (
+	// HealthCheckSucceededCondition indicates that a host is healthy.
+	HealthCheckSucceededCondition clusterv1.ConditionType = "HealthCheckSucceeded"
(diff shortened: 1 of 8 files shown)


guettli:
Closing this, we want a solution which is more generic (which works for other bare-metal providers, too).

PR #1220, opened 2024-03-19
:seedling: handle case where rescue system can't be reached.

guettli:

Provide better err msg if rescue system can't be reached. Handle this case better, because currently it reaches and endless loop of hardware reboots if a bare metal server's rescue system is not reachable.

Additionally, the controller emits an Event for every reboot now. This makes it more clear what is happening.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -225,9 +225,11 @@ func (s *Service) actionPreparing() actionResult {
 		if err := handleSSHError(sshClient.Reboot()); err != nil {
 			return actionError{err: fmt.Errorf("failed to reboot server via ssh: %w", err)}
 		}
-
+		msg := fmt.Sprintf("Phase %s: Rebooting via ssh into rescue mode.",
+			s.scope.HetznerBareMetalHost.Spec.Status.ProvisioningState)
+		record.Eventf(s.scope.HetznerBareMetalHost, "RebootBMServer", msg)
 		// we immediately set an error message in the host status to track the reboot we just performed
-		s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeSSHRebootTriggered, "ssh reboot triggered")
+		s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeSSHRebootTriggered, msg)
 		return actionComplete{}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -679,10 +712,6 @@ func (s *Service) analyzeSSHErrorRegistering(sshErr error) (isSSHTimeoutError, i
 		}
 		reterr = fmt.Errorf("wrong ssh key: %w", sshErr)
 	case sshclient.IsConnectionRefusedError(sshErr):
-		if !rebootTriggered {
-			// Reboot did not trigger
-			return false, false, nil
-		}
```
guettli, line 685: @janiskemper This is the only "real" change in the PR.  All other changes just improve error messages or create events.

AFAIK we have been both looking at that change. Does that look good to you?

janiskemper: I'd like to spend some time thinking about possible issues. I have this on my list!

janiskemper: I think we might have the following problem: 
In case we trigger a reboot but that reboot fails somehow (either API issue of Hetzner or ssh command didn't properly work), then we catch this here right now. This works because a connection refused error comes when the port is wrong. If the reboot didn't even start and we think it did, we try e.g. to go into the rescue mode with port 22, but we are currently not in rescue mode and the port is different. Then we get connection refused. This case is handled here, because if we realize that the reboot is not even triggered, then we can trigger it again.

If we just remove it, then we will not catch this case and I am not sure how important it is. We can either remove this now completely and then just see what happens, or we can add something to the current code. This addition would ensure that we have tried our best to trigger the reboot and it still didn't work, therefore we probably have a permament error. 

We can solve this by adding `if !rebootTriggered && s.scope.HetznerBareMetalHost.Spec.Status.ErrorType != infrav1.ErrorTypeHardwareRebootTriggered {`. 

If we do that, we make sure that we actually triggered the "strongest" form of reboot and it still doesn't work. 

At the same time, we will not ignore it forever and only until we actually triggered a hardware reboot. If this hardware reboot still didn't work, then we will go into the connection refused error and catch it properly

guettli: @janiskemper thank you. Makes sense. I added your if-statement.

guettli:
Janis will merge this PR

janiskemper (approved):

PR #1242, opened 2024-04-02
🌱 log name of server and lb

kranurag7:
till now we only used numeric ID in logs which becomes sometimes hard during debugging as things needs to cross checked.


guettli (changes requested):
I would avoid %v and use the formatting which is matching to the type.

The %v was in the old code. But if you change the line, I would change that, too.

janiskemper (approved):

PR #1264, opened 2024-04-12
:seedling: machinedeployment with RAID, reject host if not enough WWNs.

guettli:

If a machine does not have root device hints, then the host gets chosen ignoring the config of the machinedeployment (RAID yes/no).

This is needed, so that the hmbh gets the rootDeviceHints populated.

Up to now this could lead to provisioning a machine for read, although not two WWNs were set.

Zulip [topic](https://chat.syself.org/#narrow/stream/20-project.2Fhetzner/topic/Check.20HBMH.20for.20WWN.20config)

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -407,59 +407,8 @@ func (s *Service) chooseHost(ctx context.Context) (
 		// from now on each "continue" should add an entry
 		// to mapOfSkipReasons.
 		unusedHostsCounter++
-
-		// This comes first, because we should not look too deep into machines
-		// which are not in our scope.
-		if !labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
-			mapOfSkipReasons["label-selector-does-not-match"]++
+		if s.skipHost(labelSelector, host, mapOfSkipReasons) {
 			continue
-		}
-
-		if host.GetDeletionTimestamp() != nil {
-			mapOfSkipReasons["hbmh-has-deletion-timestamp"]++
-			continue
-		}
-
-		if host.Spec.MaintenanceMode != nil && *host.Spec.MaintenanceMode {
-			mapOfSkipReasons["hbmh-in-maintenance-mode"]++
-			continue
-		}
-		if host.Spec.Status.ErrorMessage != "" {
-			mapOfSkipReasons["hbmh-has-error-message-in-status"]++
-			continue
-		}
-
-		if host.Spec.Status.ProvisioningState != infrav1.StateNone {
-			mapOfSkipReasons["hbmh-in-wrong-provisioning-state"]++
-			continue
-		}
-
-		// This comes last, because now we would choose the machine, but we check
-		// if the config is correct.
-		if host.Spec.RootDeviceHints != nil &&
-			host.Spec.RootDeviceHints.IsValid() {
-			// Even if there are no rootDeviceHints specified, the host should be picked.
-			// After the phase registering, the process to provision the server stops and
-			// waits for the user to specify the rootDeviceHints.
-			// Here (RootDeviceHints exists and is valid) we want to check whether
-			// the specified rootDeviceHints fit with the InstallImage configuration
-			// of the HetznerBareMetalMachine. If not, it is not valid.
-			// Doing that without first choosing the hbmh would be nice, there is a feature request:
-			// https://github.com/syself/cluster-api-provider-hetzner/issues/1166
-			if s.scope.BareMetalMachine.Spec.InstallImage.Swraid == 1 {
-				// Machine should have RAID. Skip machines which have less than two WWNs
-				lenOfWwnSlice := len(host.Spec.RootDeviceHints.Raid.WWN)
-				if lenOfWwnSlice < 2 {
-					mapOfSkipReasons[fmt.Sprintf("machine-should-use-swraid-but-only-%d-RAID-WWN-in-hbmh", lenOfWwnSlice)]++
-					continue
-				}
-			} else { //nolint:gocritic
-				// Machine should have no RAID.
-				if host.Spec.RootDeviceHints.WWN == "" {
-					mapOfSkipReasons["machine-should-use-no-swraid-and-no-non-raid-WWN-in-hbmh"]++
-					continue
-				}
-			}
 		}
 
 		availableHosts = append(availableHosts, &hosts.Items[i])
@@ -502,6 +451,70 @@ func (s *Service) chooseHost(ctx context.Context) (
 	}
 
 	return chosenHost, helper, "", nil
+}
+
+func (s *Service) skipHost(labelSelector labels.Selector, host infrav1.HetznerBareMetalHost, mapOfSkipReasons map[string]int) bool {
+	// This comes first, because we should not look too deep into machines
+	// which are not in our scope.
+	if !labelSelector.Matches(labels.Set(host.ObjectMeta.Labels)) {
+		mapOfSkipReasons["label-selector-does-not-match"]++
+		return true
+	}
+
+	if host.GetDeletionTimestamp() != nil {
+		mapOfSkipReasons["hbmh-has-deletion-timestamp"]++
+		return true
+	}
+
+	if host.Spec.MaintenanceMode != nil && *host.Spec.MaintenanceMode {
+		mapOfSkipReasons["hbmh-in-maintenance-mode"]++
+		return true
+	}
+	if host.Spec.Status.ErrorMessage != "" {
+		mapOfSkipReasons["hbmh-has-error-message-in-status"]++
+		return true
+	}
+
+	if host.Spec.Status.ProvisioningState != infrav1.StateNone {
+		mapOfSkipReasons["hbmh-in-wrong-provisioning-state"]++
+		return true
+	}
+
+	if host.Spec.RootDeviceHints == nil ||
+		(host.Spec.RootDeviceHints.WWN == "" && len(host.Spec.RootDeviceHints.Raid.WWN) == 0) {
+		// Even if there are no rootDeviceHints specified, the host should be picked.
+		// After the phase registering, the process to provision the server stops and
+		// waits for the user to specify the rootDeviceHints.
+		// Here (RootDeviceHints exists and is valid) we want to check whether
+		// the specified rootDeviceHints fit with the InstallImage configuration
+		// of the HetznerBareMetalMachine. If not, it is not valid.
+		// Doing that without first choosing the hbmh would be nice, there is a feature request:
+		// https://github.com/syself/cluster-api-provider-hetzner/issues/1166
+		// See "tworaidchecks" for the other place.
+		return false
+	}
+
+	// IsValid returns false if rootDeviceHints are empty, but this case was handled before
+	if !host.Spec.RootDeviceHints.IsValid() {
+		mapOfSkipReasons["hbmh-has-invalid-rootDeviceHints"]++
+		return true
+	}
+
+	if s.scope.BareMetalMachine.Spec.InstallImage.Swraid == 1 {
+		// Machine should have RAID. Skip machines which have less than two WWNs
+		lenOfWwnSlice := len(host.Spec.RootDeviceHints.Raid.WWN)
+		if lenOfWwnSlice < 2 {
+			mapOfSkipReasons[fmt.Sprintf("machine-should-use-swraid-but-only-%d-RAID-WWN-in-hbmh", lenOfWwnSlice)]++
+			return true
+		}
+		return false
+	}
+	// Machine should have no RAID.
+	if host.Spec.RootDeviceHints.WWN == "" {
+		mapOfSkipReasons["machine-should-use-no-swraid-and-no-non-raid-WWN-in-hbmh"]++
+		return true
+	}
+	return false
 }
 
 func reasonString(mapOfSkipReasons map[string]int, unusedHostsCounter int) string {
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -524,6 +525,8 @@ func (s *Service) actionRegistering() actionResult {
 
 		isSSHTimeoutError, isSSHConnectionRefusedError, err := s.analyzeSSHOutputRegistering(out)
 		if err != nil {
+			// This can happen if the bare-metal server was taken by another mgt-cluster.
+			// Check in https://robot.hetzner.com/server for the "History" of the server.
 			return actionError{err: fmt.Errorf("failed to handle incomplete boot - registering: %w", err)}
 		}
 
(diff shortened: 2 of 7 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -588,6 +588,31 @@ func (s *Service) actionRegistering() actionResult {
 		return s.recordActionFailure(infrav1.RegistrationError, err.Error())
 	}
 
+	// Check RAID for the second time.
+	// This happens for new machines which don't have rootDeviceHints.
```
janiskemper, line 592: this comment is wrong. If you reach this check, then rootDeviceHints are actually set. They were not set at the time when the host was chosen to provision it.

pkg/services/baremetal/host/host.go
```diff
@@ -588,6 +588,31 @@ func (s *Service) actionRegistering() actionResult {
...
+	// See "tworaidchecks" for the other place.
+	msg = ""
+	if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage == nil {
```
janiskemper, line 595: do you have a reason for this check? Because the host takes installimage != nil as a reason to start provisioning. IF this is not set, we would never reach this state. 

Reference: "NeedsProvisioning()" function

pkg/services/baremetal/host/host.go
```diff
@@ -588,6 +588,31 @@ func (s *Service) actionRegistering() actionResult {
...
+		msg = "Internal Error: HetznerBareMetalHost.Spec.Status.InstallImage is nil. setHostSpec() not called yet?"
+	} else if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Swraid != 0 &&
+		len(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.Raid.WWN) < 2 {
+		msg = "HetznerBareMetalHost.Spec.Status.InstallImage.Swraid is active. List spec.rootDevideHints.raid.wwn must contain at least two entries."
+	} else if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Swraid == 0 &&
+		len(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.Raid.WWN) > 0 {
```
janiskemper, line 601: what we actually want to check here is that rootDeviceHints.wwn is not empty, don't we?

pkg/services/baremetal/host/host.go
```diff
@@ -588,6 +588,31 @@ func (s *Service) actionRegistering() actionResult {
...
+		msg = "HetznerBareMetalHost.Spec.Status.InstallImage.Swraid is not active. List spec.rootDevideHints.raid.wwn must be empty."
+	}
+	if msg != "" {
+		s.scope.Logger.Info(msg)
+		conditions.MarkFalse(
+			s.scope.HetznerBareMetalHost,
+			infrav1.RootDeviceHintsValidatedCondition,
+			infrav1.StorageDeviceNotFoundReason,
```
janiskemper, line 609: this is not the appropriate reason I guess

janiskemper (changes requested):
If you don't want to trigger the e2e tests when you push during development, it makes sense to mark the PR as "draft"

pkg/services/baremetal/host/host.go
guettli: I removed that line:

```
This happens for new machines which don't have rootDeviceHints.
```

pkg/services/baremetal/host/host.go
guettli: > do you have a reason for this check? Because the host takes installimage != nil as a reason to start provisioning. IF this is not set, we would never reach this state.
> 
> Reference: "NeedsProvisioning()" function

Some tests failed, and I got null-pointer exceptions.

If you have a better solution, then please tell me.

pkg/services/baremetal/host/host.go
guettli: Some lines above `IsValidWithMessage()` gets called. There we check that single-WWN  XOR  raid-WWNs are specified. So it does not matter if I check the Raid.WWN list or the single WWN string.

But I can change that, if you prefer the other.

pkg/services/baremetal/host/host.go
guettli: I use `infrav1.ValidationFailedReason,` now

pkg/services/baremetal/host/host.go
janiskemper: yes I think that makes it easier to read. I am aware that it comes down to the same thing

pkg/services/baremetal/host/host.go
janiskemper: I checked the code again and I see no reason why InstallImage shouldn't be specified at this point. If this is needed, I believe we have a bug somewhere. Is this easily reproducible or happening sometimes?

pkg/services/baremetal/host/host_test.go
```diff
@@ -1192,6 +1188,62 @@ var _ = Describe("actionRegistering", func() {
 	)
 })
 
+func registringSSHMock(storageStdOut string) *sshmock.Client {
```
janiskemper, line 1191: typo

pkg/services/baremetal/host/host.go
janiskemper: @guettli I believe the reason is that you have added this here to this function and it wasn't there before. This probably fails in the unit tests (I guess you meant that) because we did not "properly" define the host objects with InstallImage specified, as it wasn't necessary before. 

The solution is to define the host objects in the unit tests appropriately to this change in the function.

guettli: @janiskemper what do you want:

Image for some strange reason in the future the value is nil in production.

What do you want to happen?

option 1: a null pointer exception

option 2: My current code

option 3: We do a panic("Internal Error: HetznerBareMetalHost.Spec.Status.InstallImage is nil. setHostSpec() not called yet?")

... or something else?

guettli: > The solution is to define the host objects in the unit tests appropriately to this change in the function.

That is already fixed.

janiskemper: so you say we don't need this check "installimage != nil" anymore?

If it is still needed, we should rather change the tests instead of changing the functions based on our tests

janiskemper: we update the tests so that we don't have to change the actual code to comply to wrong tests! If you don't know how that works, let's talk about it

pkg/services/baremetal/host/host.go
janiskemper: @guettli I think you now did it the other way around and also made the other case indirect. I would prefer the direct one for both! 

If you are supposed to specifiy a raid, it should be "specify a raid" and not "specify not a wwn". And the other way around

pkg/services/baremetal/host/host.go
guettli: > we update the tests so that we don't have to change the actual code to comply to wrong tests! If you don't know how that works, let's talk about it

reading between the lines, I guess you want to say: "Please choose option 1". Is this correct?

janiskemper: the tests should be configured in a way that there is no issue here with the code or nil pointers.. If we have always InstallImage configured for tests of this function, this never occurs.

pkg/services/baremetal/host/host.go
guettli: @janiskemper the two error messages are now:

> Invalid HetznerBareMetalHost: spec.status.installImage.swraid is active and wwn="wwn1". Use spec.rootDevideHints.raid.wwn instead of spec.rootDevideHints.wwn.

> Invalid HetznerBareMetalHost: spec.rootDevideHints.raid.wwn ([wwn1 wwn2]) contains several WWNs, but spec.status.installImage.swraid is not active.

If you want it different, then please tell me exactly your desired err msg.

guettli: @janiskemper I implemented option-1, please speak up if you want something else.

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -434,10 +434,16 @@ func (s *Service) chooseHost(ctx context.Context) (
 			continue
 		}
 
-		// This comes last, because now we would choose the machine, but we check
-		// if the config is correct.
+		if host.Spec.RootDeviceHints != nil &&
```
janiskemper, line 437: can you maybe put this to another PR? Because this is a bugfix that should be visible to users through PR description, commit messages, and release notes. Right now it would not be visible anywhere. 

Additionally, can you add a comment to this that describes the != nil check? Because I actually thought you did the wrong thing here because you don't allow empty rootDeviceHints. You do - but you do it in a hidden way that is not explained. 

However, is this != nil check actually what we want? What about people using a template where rootDeviceHints is actually specified but left empty? Those people get an error here right now without having an idea that they should remove the "rootDeviceHints" key from their host template...

pkg/services/baremetal/host/host.go
janiskemper: I described it above, that I think you should do the "direct" check and not the "indirect" one. I understand that it comes down to the same thing because we have the XOR operator, but I do think that it makes the code and the error message better.

If you don't know what I mean, then let's talk about it

pkg/services/baremetal/host/host.go
janiskemper: I see that you removed the InstallImage != nil check without changing the unit tests. So there was no problem in the first place? I'm confused. If you wanna explain me why you added it first and why you removed it now without changing anything else, then let's talk about it

guettli: @janiskemper 

I changed code. Look for `host.Spec.Status.InstallImage = &infrav1.InstallImage{}`

And the second change was that: `helpers.WithRootDeviceHintWWN(),`

Both changes were done before I requested the review from you.

pkg/services/baremetal/host/host.go
guettli: @janiskemper it feels exhausting for me to guess what you want. Please take the PR and adapt the error message to the way you want it to be. Or post a code snippet here, and I will update the PR.

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -504,6 +453,69 @@ func (s *Service) chooseHost(ctx context.Context) (
...
+		// Here (RootDeviceHints exists and is valid) we want to check whether
+		// the specified rootDeviceHints fit with the InstallImage configuration
+		// of the HetznerBareMetalMachine. If not, it is not valid.
+		// Doing that without first choosing the hbmh would be nice, there is a feature request:
+		// https://github.com/syself/cluster-api-provider-hetzner/issues/1166
+		// See "tworaidchecks" for the other place.
+		return false
+	}
+
+	if !host.Spec.RootDeviceHints.IsValid() {
```
janiskemper, line 497: you should add one small comment to make this understandable together with the previous check:

 // IsValid returns false if rootDeviceHints are empty, but this case was handled before

janiskemper (commented):
one small comment. Otherwise looks fine and you can squash

janiskemper:
thanks, plz squash

pkg/services/baremetal/host/host_test.go
```diff
@@ -1192,6 +1188,62 @@ var _ = Describe("actionRegistering", func() {
...
+	sshMock.On("GetHardwareDetailsCPUFlags").Return(sshclient.Output{StdOut: "flag1 flag2 flag3"})
+	sshMock.On("GetHardwareDetailsCPUThreads").Return(sshclient.Output{StdOut: "123"})
+	sshMock.On("GetHardwareDetailsCPUCores").Return(sshclient.Output{StdOut: "12"})
+	sshMock.On("GetHardwareDetailsDebug").Return(sshclient.Output{StdOut: "Dummy outupt"})
+	return sshMock
+}
+
+var _ = Describe("actionRegistering check RAID", func() {
+	It("check RAID", func() {
+		// 		NAME="loop0" LABEL="" FSTYPE="ext2" TYPE="loop" HCTL="" MODEL="" VENDOR="" SERIAL=""  WWN="dummyWWN" ROTA="0"`)
```
janiskemper, line 1214: did you forget this comment?

pkg/services/baremetal/host/host_test.go
```diff
@@ -1192,6 +1188,62 @@ var _ = Describe("actionRegistering", func() {
...
+		StdOut: `name="eth0" model="Realtek Semiconductor Co., Ltd. RTL8111/8168/8411 PCI Express Gigabit Ethernet Controller (rev 15)" mac="a8:a1:59:94:19:42" ipv4="23.88.6.239/26" speedMbps="1000"
+name="eth0" model="Realtek Semiconductor Co., Ltd. RTL8111/8168/8411 PCI Express Gigabit Ethernet Controller (rev 15)" mac="a8:a1:59:94:19:42" ipv6="2a01:4f8:272:3e0f::2/64" speedMbps="1000"`,
+	})
+	sshMock.On("GetHardwareDetailsCPUArch").Return(sshclient.Output{StdOut: "myarch"})
+	sshMock.On("GetHardwareDetailsCPUModel").Return(sshclient.Output{StdOut: "mymodel"})
+	sshMock.On("GetHardwareDetailsCPUClockGigahertz").Return(sshclient.Output{StdOut: "42654"})
+	sshMock.On("GetHardwareDetailsCPUFlags").Return(sshclient.Output{StdOut: "flag1 flag2 flag3"})
+	sshMock.On("GetHardwareDetailsCPUThreads").Return(sshclient.Output{StdOut: "123"})
+	sshMock.On("GetHardwareDetailsCPUCores").Return(sshclient.Output{StdOut: "12"})
+	sshMock.On("GetHardwareDetailsDebug").Return(sshclient.Output{StdOut: "Dummy outupt"})
```
janiskemper, line 1208: typo

guettli:
@guettli I tested it, and while doing that, I realized that the err msg like this:

> missing storage device for root device hint eui.1234

Could be improved. It is now:

> missing storage device for root device hint eui.1234. Known WWNs: [eui.002538b411b2cee2 eui.0025388801b4dff2]

pkg/services/baremetal/host/host.go
```diff
@@ -617,17 +618,16 @@ func (s *Service) actionRegistering() actionResult {
 }
 
 func validateRootDeviceWwnsAreSubsetOfExistingWwns(rootDeviceHints *infrav1.RootDeviceHints, storageDevices []infrav1.Storage) error {
+	knownWWNs := make([]string, 0, len(storageDevices))
```
janiskemper, line 621: if you do this already, you should use a map and save some computing time compared to looping over slices in slices.Contains()

guettli: @janiskemper 

> if you do this already, you should use a map and save some computing time compared to looping over slices in slices.Contains()

Are you sure a map is faster? A slice is sequential memory access, which is very fast. 

Here this gets explained: https://stackoverflow.com/a/31707228/633961

janiskemper: not faster apparently, as you tried it, just fewer steps to do. But it is not very significant anyway

guettli: @janiskemper what is missing for this PR to get your approval?

guettli:
@janiskemper 

I tested the code in a cluster with Tilt.

It worked. Except that CAPI seemed to ignore that the machine had a failure set (FatalError in caph). The machine was not deleted.

pkg/services/baremetal/host/host.go
```diff
@@ -577,34 +580,56 @@ func (s *Service) actionRegistering() actionResult {
...
+	msg = ""
+	if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Swraid != 0 &&
+		len(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.Raid.WWN) < 2 {
+		msg = "Invalid HetznerBareMetalHost: spec.status.installImage.swraid is active. Use at least two WWNs in spec.rootDevideHints.raid.wwn."
+	} else if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Swraid == 0 &&
+		s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.WWN == "" {
+		msg = "Invalid HetznerBareMetalHost: spec.status.installImage.swraid is not active. Use spec.rootDevideHints.wwn and leave raid.wwn empty."
+	}
+	if msg != "" {
+		s.scope.Logger.Info(msg)
```
janiskemper, line 606: Add a comment that this triggers a FailureMessage on the HetznerBareMetalMachine and CAPI machine and will lead to this Machine to be deleted. 
Another machine will not take the same host anymore because the rootDeviceHints don't fit

janiskemper (commented):
one small comment. Otherwise we can squash this. Is it possible to keep Yash as one of the authors?

pkg/services/baremetal/host/host.go
janiskemper: thanks!

janiskemper (commented):
looks good

janiskemper (approved):

PR #1284, opened 2024-04-30
🌱Add condition indicating deletion timestamp is set

### LINKED ISSUE #1229: Condition indicating that HetznerBareMetalHost is deleting
/kind feature

The deletion process of a HetznerBareMetalHost needs some time. We can introduce a condition / a reason that indicates this and as such is printed in the printcolumns of the object. This makes it easier for users because they have an immediate feedback without having to check the deletion timestamp of the object.

yrs147:
Add condition indicating deletion timestamp is set
Fixes #1229

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -200,6 +200,11 @@ const (
 	HostAssociateFailedReason = "HostAssociateFailed"
 )
 
+const (
+	// DeletionInProgressReason indicates that a host is being deleted.
+	DeletionInProgressReason = "DeletionInProgress"
+)
+
 // deprecated conditions.
 
 const (
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -97,6 +97,16 @@ func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err e
 
 	oldHost := s.scope.HetznerBareMetalHost.DeepCopy()
 
+	if !s.scope.HetznerBareMetalHost.DeletionTimestamp.IsZero() {
+		conditions.MarkFalse(
+			s.scope.HetznerBareMetalHost,
+			infrav1.HostReadyCondition,
+			infrav1.DeletionInProgressReason,
+			clusterv1.ConditionSeverityWarning,
+			"Host is not ready because it is being deleted",
+		)
+	}
+
 	hostStateMachine := newHostStateMachine(s.scope.HetznerBareMetalHost, s, s.scope.Logger)
 
 	defer func() {
@@ -104,7 +114,6 @@ func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err e
 		conditions.Delete(s.scope.HetznerBareMetalHost, infrav1.DeprecatedHetznerBareMetalHostReadyCondition)
 		conditions.Delete(s.scope.HetznerBareMetalHost, infrav1.DeprecatedHostProvisionSucceededCondition)
 		conditions.Delete(s.scope.HetznerBareMetalHost, infrav1.DeprecatedRateLimitExceededCondition)
-
 		conditions.SetSummary(s.scope.HetznerBareMetalHost)
 
 		// save host if it changed during reconciliation


api/v1beta1/conditions_const.go
```diff
@@ -200,6 +200,13 @@ const (
 	HostAssociateFailedReason = "HostAssociateFailed"
 )
 
+const (
```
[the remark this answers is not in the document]
janiskemper: what about "HostReadyCondition"? https://github.com/syself/cluster-api-provider-hetzner/blob/main/api/v1beta1/conditions_const.go#L103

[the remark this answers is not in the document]
yrs147: A host can be  `Available` but not `Ready` , in which case marking `HostReadyCondition` to false won't be correct imo , thats why I suggested `HostAvailiableCondition`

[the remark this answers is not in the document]
janiskemper: sorry I don't get what you mean tbh :D If you don't want to elaborate via text we can just talk in the next days

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -170,6 +170,18 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(ctx context.Con
...
 		var needsUpdate bool
 		if !bmHost.DeletionTimestamp.IsZero() && bmHost.Spec.ConsumerRef == nil {
+			conditions.MarkTrue(
+				bmHost,
+				infrav1.DeletionInProgressCondition,
+			)
+			conditions.MarkFalse(
+				bmHost,
+				infrav1.CredentialsAvailableCondition,
+				infrav1.HetznerSecretUnreachableReason,
```
guettli, line 180: is `HetznerSecretUnreachableReason` the correct reason?

yrs147: not sure , what do would you suggest it to be ?

guettli: I would set it like below:

```
			conditions.MarkFalse(
				bmHost,
				infrav1.DeletionInProgressCondition,
				infrav1.DeletionInProgressReason,
				"Deletion in progress",
			)
```

yrs147: should I create a `CredentialsAvailableReason` for this then ?
cc @janiskemper

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -188,6 +196,15 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(ctx context.Con
 		// Handle StateDeleting
 	case infrav1.StateDeleting:
 		if !utils.StringInList(bmHost.Finalizers, infrav1.BareMetalHostFinalizer) {
+			conditions.MarkFalse(
```
janiskemper, line 199: I think you can remove this. Right now it doesn't have any effect (see that you don't update the host after this). You could, if anything, add it to the code below, but also there you remove the finalizer already. After the finalizer is removed, the object will be deleted very quickly. Therefore, we probably don't need a condition here at all.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -170,6 +170,14 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(ctx context.Con
 	case infrav1.StateNone:
 		var needsUpdate bool
 		if !bmHost.DeletionTimestamp.IsZero() && bmHost.Spec.ConsumerRef == nil {
+			conditions.MarkFalse(
```
janiskemper, line 173: you should set this condition independent of Spec.ConsumerRef. So just split up the if statements to set the condition always when the deletion timestamp is not nil

janiskemper (changes requested):

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -180,19 +180,26 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(ctx context.Con
 	switch bmHost.Spec.Status.ProvisioningState {
 	// Handle StateNone: check whether needs to be provisioned or deleted.
 	case infrav1.StateNone:
-		var needsUpdate bool
-		if !bmHost.DeletionTimestamp.IsZero() && bmHost.Spec.ConsumerRef == nil {
-			bmHost.Spec.Status.ProvisioningState = infrav1.StateDeleting
-			needsUpdate = true
+		if !bmHost.DeletionTimestamp.IsZero() {
+			if !conditions.Has(bmHost, infrav1.HostReadyCondition) {
```
janiskemper, line 184: you shouldn't do this check here. You should set it independently of whether it exists or not. You even have to overwrite it. 

What I meant earlier was that you might still keep needsUpdate and then need to check whether the condition you set actually changed something in the object or not. But for that you need to check the reason as well, not only that the condition exists. 

I would just remove this if statement and always set the condition on false and you are good to go!

janiskemper (changes requested):

janiskemper (commented):
LGTM. Thank you! Please squash your commits now

pkg/services/baremetal/host/host.go
```diff
@@ -91,6 +91,29 @@ func NewService(s *scope.BareMetalHostScope) *Service {
 
 // Reconcile implements reconcilement of HetznerBareMetalHosts.
 func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err error) {
+	// Fetch the Hetzner bare metal host instance.
```
janiskemper, line 94: the host is in the scope (s.scope)

pkg/services/baremetal/host/host.go
```diff
@@ -91,6 +91,29 @@ func NewService(s *scope.BareMetalHostScope) *Service {
...
+	}
+	if !bmHost.DeletionTimestamp.IsZero() {
+		conditions.MarkFalse(
+			bmHost,
+			infrav1.HostReadyCondition,
+			infrav1.DeletionInProgressReason,
+			clusterv1.ConditionSeverityInfo,
+			"Host is not ready because it is being deleted",
+		)
+		conditions.SetSummary(bmHost)
```
janiskemper, line 114: "setsummary" should be done anyway in the defer below

pkg/services/baremetal/host/host.go
```diff
@@ -91,6 +91,29 @@ func NewService(s *scope.BareMetalHostScope) *Service {
...
+		Namespace: s.scope.HetznerBareMetalHost.Namespace,
+		Name:      s.scope.HetznerBareMetalHost.Name,
+	}, bmHost)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get HetznerBareMetalHost : %w", err)
+	}
+	if !bmHost.DeletionTimestamp.IsZero() {
```
janiskemper, line 106: you need to put this code after the "oldHost" is defined. The oldHost is the "original" state of the host and all changes thereafter are updated via the "Update" API call below

janiskemper (changes requested):

janiskemper (commented):
can you revert all changes to hetznerbaremetalhost_controller? I think those changes are not relevant anymore

janiskemper (commented):
thanks! This looks good now. You can squash all commits and do one small change:
use the severity "warning" instead of "info". I'm a bit afraid that it wouldn't be shown to users otherwise

janiskemper (approved):

PR #1285, opened 2024-04-30
:sparkles: Add constant hostnames to bare metal servers

janiskemper:
Adding the feature to keep hostnames of bare metal servers constant. This can be triggered by a feature flag on the Cluster object. It takes the usual "bm-" prefix together with the cluster name as well as the serverID of the bare metal server.

This feature can be used by adding the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"` to `Cluster` objects.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/scope/baremetalhost.go
+++ b/pkg/scope/baremetalhost.go
@@ -148,8 +148,5 @@ func (s *BareMetalHostScope) Hostname() (hostname string) {
 }
 
 func (s *BareMetalHostScope) hasConstantHostname() bool {
-	if s.Cluster.Annotations != nil {
-		return s.Cluster.Annotations[infrav1.ConstantBareMetalHostnameAnnotation] == "true"
-	}
-	return false
+	return s.Cluster.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true"
 }
--- a/pkg/scope/baremetalhost.go
+++ b/pkg/scope/baremetalhost.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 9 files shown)


pkg/scope/baremetalhost.go
```diff
@@ -128,3 +135,21 @@ func (s *BareMetalHostScope) GetRawBootstrapData(ctx context.Context) ([]byte, e
...
+	if s.hasConstantHostname() {
+		hostname = fmt.Sprintf("%s%s-%v", infrav1.BareMetalHostNamePrefix, s.Cluster.Name, s.HetznerBareMetalHost.Spec.ServerID)
+	} else {
+		hostname = infrav1.BareMetalHostNamePrefix + s.HetznerBareMetalHost.Spec.ConsumerRef.Name
+	}
+
+	return hostname
+}
+
+func (s *BareMetalHostScope) hasConstantHostname() bool {
```
guettli, line 150: You could use this, which is a bit shorter:
```
return s.Cluster.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true"
```
But it does not matter much.

guettli (changes requested):
lgtm, except that docs about the new feature are missing.

pkg/scope/baremetalhost.go
janiskemper: I don't think that this method is safe to use in the case where annotations are nil. Probably that will never happen, but that's the only reason for my if statement. Should I remove the check for nil annotations?

guettli: Getting a value from a nil map just returns the zero value. Setting on a nil map panics.

Yes, I would remove the "if" since it is not needed

guettli (approved):

PR #1287, opened 2024-04-30
🐛Fix flaky reach state image installing test

### LINKED ISSUE #1228: Flaky test: "reaches state image installing"
/kind bug

There is a flaky unit test (based on envtest) that tests whether a bare metal host reaches the state "image-installing". Probably this is due to the fact that image-installing is an intermittent state and the host will continue provisioning and reach further states. We can include a check that says it should reach AT LEAST state image-installing. This should fix the flakiness that comes because sometimes the controller finds the state image-installing and sometimes it is too slow and the provisioning process went too far.

The test should not be designed based on luck in checking at the right time and should be stable.

yrs147:

Fixes #1228

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hetznerbaremetalhost_controller_test.go
+++ b/controllers/hetznerbaremetalhost_controller_test.go
@@ -340,7 +340,7 @@ var _ = Describe("HetznerBareMetalHostReconciler", func() {
 						testEnv.GetLogger().Info("reaches the state image installing. Get failed", "err", err)
 						return false
 					}
-					if host.Spec.Status.ProvisioningState == infrav1.StateImageInstalling {
+					if host.Spec.Status.ProvisioningState == infrav1.StateImageInstalling || host.Spec.Status.ProvisioningState == infrav1.StateProvisioned {
 						return true
 					}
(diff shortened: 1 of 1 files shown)


guettli (approved):

yrs147:
> looks good! Were you able to validate that sometimes it actually reaches the state provisioned, or is this more a good guess (also one that I made)?
> 
> just asking out of curiousity.
> 
> Other than that plz rename commit and PR. this is a bugfix, therefore you can use 🐛



> looks good! Were you able to validate that sometimes it actually reaches the state provisioned, or is this more a good guess (also one that I made)?
> 
> just asking out of curiousity.
> 
> Other than that plz rename commit and PR. this is a bugfix, therefore you can use 🐛

I did come up with this based on what you said to me earlier :)

janiskemper:
alright so I guess we haven't 100% validated that this fixes it, it's a good hypthosis. An issue might be if "provisioned" is not actually reached but it gets stuck in another state beforehand. If you change the title as I suggested above, we can merge this

PR #1294, opened 2024-05-06
🌱 avoid redacting logs when running e2e tests

kranurag7:
We were running a script to redact the token in _artifacts directory which gets produced when we run our e2e tests. Since, ginkgo spec are run one after another, once first spec runs and we redact the logs, the manifest (kind: Secret) gets invalid and we get an error in our e2e tests.

`Error from server (BadRequest): error when creating "secret.yaml": Secret in version "v1" cannot be handled as a Secret: illegal base64 data at input byte 0`

This commit avoids redacting logs after each spec. We still do it after the whole suite completes.


janiskemper:
can you explain this change?

janiskemper (approved):

PR #1295, opened 2024-05-06
:bug: Fix CSR for ConstantBareMetalHostname

guettli:

If a cluster uses ConstantBareMetalHostname, then the CSRs were not approved, since the node-name
did not match the expected pattern.

This PR extends the CSR approving, so that CSRs for constant bm hostnames get approved.

### DIFF
--- a/controllers/csr_controller.go
+++ b/controllers/csr_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


janiskemper:
@kranurag7 can you test this and check whether
a) hcloud servers get csrs approved
b) bare metal servers without constant node name get csrs approved
c) bare metal servers with constant node name get csrs approved

janiskemper (commented):
LGTM from my side, but didn't test it

controllers/csr_controller.go
```diff
@@ -228,7 +232,18 @@ func (r *GuestCSRReconciler) getMachineAddresses(
 			}
 			providerID := "hcloud://bm-" + matches[2]
 			hList := &infrav1.HetznerBareMetalMachineList{}
-			if err := r.mCluster.List(ctx, hList, &client.ListOptions{}); err != nil {
+			selector := labels.NewSelector()
+			key := "cluster.x-k8s.io/cluster-name"
```
janiskemper, line 235: there is a constant for this (clusterv1.ClusterLabel or something)

guettli: @janiskemper, ok I use that constant now.

kranurag7 (approved):
LGTM, thank you Thomas.

PR #1314, opened 2024-05-20
:seedling: Allow constant hostname annotation on machines

janiskemper:
Allowing the annotation also on HetznerBaremetalMachine to achieve more flexibility (e.g. with regards to ClusterClass approaches).

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/docs/topics/hetzner-baremetal.md
+++ b/docs/topics/hetzner-baremetal.md
@@ -391,4 +391,6 @@ In some cases it has advantages to fix the hostname and with it the names of nod
 
 Therefore, there is the possibility to create a cluster that uses fixed node names for bare metal servers. Please note: this only applies to the bare metal servers and not to Hetzner Cloud servers.
 
-You can trigger this feature by creating a `Cluster` with the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"`. This is still an experimental feature but it should be save to use and to also update existing clusters with this annotation. All new machines will be created with this constant hostname.
+You can trigger this feature by creating a `Cluster` or `HetznerBareMetalMachine` (you can choose) with the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"`. Of course, `HetznerBareMetalMachines` are not created by the user. However, if you use the `ClusterClass`, then you can add the annotation to a `MachineDeployment`, so that all machines are created with this annotation. 
+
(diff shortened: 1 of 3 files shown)


docs/topics/hetzner-baremetal.md
```diff
@@ -391,4 +391,4 @@ In some cases it has advantages to fix the hostname and with it the names of nod
 
 Therefore, there is the possibility to create a cluster that uses fixed node names for bare metal servers. Please note: this only applies to the bare metal servers and not to Hetzner Cloud servers.
 
-You can trigger this feature by creating a `Cluster` with the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"`. This is still an experimental feature but it should be save to use and to also update existing clusters with this annotation. All new machines will be created with this constant hostname.
\ No newline at end of file
+You can trigger this feature by creating a `Cluster` or `HetznerCluster` (you can choose) with the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"`. This is still an experimental feature but it should be save to use and to also update existing clusters with this annotation. All new machines will be created with this constant hostname.
```
kranurag7, line 394:
```suggestion
You can trigger this feature by creating a `Cluster` or `HetznerCluster` (you can choose) with the annotation `"capi.syself.com/constant-bare-metal-hostname": "true"`. This is still an experimental feature but it should be safe to use and to also update existing clusters with this annotation. All new machines will be created with this constant hostname.
```

nit around docs and I think we can also add a line that `true` should be a string which is `"true"` and not `true` because then this will be a YAML keyword.

batistein (dismissed):

guettli (dismissed):

docs/topics/hetzner-baremetal.md
janiskemper: mmh that's what is written there already. Would it still be valid if users used `true` instead of `"true"`? I mean from yaml / CRD perspective?

kranurag7: I did it once by mistake. :( 
A note would be good but can be skipped as well given we already have the correct one. Not a blocker from my end.

kranurag7 (dismissed):

batistein (approved):

guettli (approved):

PR #1318, opened 2024-05-21
:seedling: Create more detailed events about rebooting BM.

guettli:

We have seen that machines get rebooted again and again.

We want to have more detailed events, so that we can debug that.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -1147,7 +1147,7 @@ func (s *Service) actionImageInstalling() actionResult {
 		record.Warn(s.scope.HetznerBareMetalHost, "RebootFailed", err.Error())
 		return actionError{err: fmt.Errorf("failed to reboot server: %w", err)}
 	}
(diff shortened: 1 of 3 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -1600,6 +1600,7 @@ func (s *Service) handleCloudInitNotStarted() actionResult {
 		if err := handleSSHError(out); err != nil {
 			return actionError{err: fmt.Errorf("failed to reboot: %w", err)}
 		}
+		createSSHRebootEvent(s.scope.HetznerBareMetalHost, "CheckCloudInitLogsForSigTerm failed")
```
janiskemper, line 1603: I don't get this. This is not about ssh reboot is it?

guettli: @janiskemper 

Some lines above: `out = oldSSHClient.Reboot()`. 

We could give that a better name, but overall the reboot gets done because the output of `CheckCloudInitLogsForSigTerm` is not the empty string.

janiskemper: I made a commit

PR #1327, opened 2024-05-31
:bug: Improve handling of bm server reboot timeouts

janiskemper:
Bare metal servers currently can time out while rebooting. However, the timeouts don't make too much sense right now.

- The timeout is too big, so that usually MachineHealthChecks will trigger before the timeout is reached. We reduce this here.
- The server was just rebooted again if a timeout is reached. However, if the timeout is actually reached, we don't want to continue, as there is probably something wrong with the server. Now we set a permanent error.

Additionally, there are some code improvements:
- The check whether a reboot has been triggered requires an API call to Hetzner API. This check is now only done when necessary, so that we safe the unneccessary API calls.
- We set a permanent error if a reboot is marked as failed, so that we stop reconciling.
- The ProvisionSucceeded condition of a host is saved in case of a permanent error to give the user some feedback that the permanent error happened and why.

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -52,7 +52,7 @@ const (
 	rebootWaitTime       time.Duration = 15 * time.Second
 	sshResetTimeout      time.Duration = 5 * time.Minute
 	softwareResetTimeout time.Duration = 5 * time.Minute
-	hardwareResetTimeout time.Duration = 60 * time.Minute
+	hardwareResetTimeout time.Duration = 5 * time.Minute
 	rescue               string        = "rescue"
 	rescuePort           int           = 22
 	gbToMebiBytes        int           = 1000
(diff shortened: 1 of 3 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -500,13 +501,30 @@ func (s *Service) handleErrorTypeHardwareRebootFailed(isSSHTimeoutError, wantsRe
...
 		}
 		msg := fmt.Sprintf("Reboot via ssh into %s failed. Now using rebootType %q.",
 			rebootInto, infrav1.RebootTypeHardware)
 		createRebootEvent(s.scope.HetznerBareMetalHost, infrav1.RebootTypeHardware, msg)
 	}
-	return nil
+
+	// if hardware reboots time out, we should fail
+	if hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.LastUpdated, hardwareResetTimeout) {
+		msg := "reboot timed out - please check if server is working properly"
```
guettli, line 513: if `wantsRescue` is true, then we should extend the message like this:

> The rescue system could not be reached. Please ensure that the machine tries to boot from network before booting from disk. This setting needs to be enabled *permanently* in the BIOS.

guettli (approved):

PR #1330, opened 2024-06-06
:seedling: Add new CX server types

apricote:

Adds the new server types `cx22`, `cx32`, `cx42` and `cx52` to the CRD validation. (There is no `cx12`). These were just launched and you can find out more at https://www.hetzner.com/news/new-cx-plans/

The previous generation (`cx11` - `cx51`) was deprecated and will be removed on 06 September 2024. I saw that you reference them in some templates and you might want to replace them until then.

Fixes #

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/hcloudmachine_types.go
+++ b/api/v1beta1/hcloudmachine_types.go
@@ -37,7 +37,7 @@ type HCloudMachineSpec struct {
 	ProviderID *string `json:"providerID,omitempty"`
 
 	// Type is the HCloud Machine Type for this machine. It defines the desired server type of server in Hetzner's Cloud API. Example: cpx11.
-	// +kubebuilder:validation:Enum=cpx11;cx21;cpx21;cx31;cpx31;cx41;cpx41;cx51;cpx51;ccx11;ccx12;ccx13;ccx21;ccx22;ccx23;ccx31;ccx32;ccx33;ccx41;ccx42;ccx43;ccx51;ccx52;ccx53;ccx62;ccx63;cax11;cax21;cax31;cax41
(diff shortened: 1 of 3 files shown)


janiskemper (approved):
thanks! @kranurag7 can you have a look and update the templates?

janiskemper:
@apricote general question: What about the fact that certain server types are only available in some locations? Do you think it is feasible that we make an actual validation so that people can be 100% sure that the type exists? For example with the US regions?

kranurag7 (approved):
LGTM, thank you Julian for the patch.

apricote:
> What about the fact that certain server types are only available in some locations? Do you think it is feasible that we make an actual validation so that people can be 100% sure that the type exists? For example with the US regions?

Thats certainly a possibility to do in the validating webhook. This makes it dependent on the Cloud API, not sure if you want that, though a 1 hour cache or similar would be good enough. 

You would also need some kind of API Token to get the list of server types. As far as I can see, there is no direct link between the HcloudMachineTemplate & the tokens. There is also no guarantee, that every user/project has the same server types available.

janiskemper:
I actually thought about hardcoding it in CAPH :sweat_smile: 
So kinda having what we already have in little more complex with taking into account the different regions. 

What do you mean by this?
> There is also no guarantee, that every user/project has the same server types 

I'm just looking for a solution that makes it the most reliable that users don't specify server types that are not available

apricote:
> There is also no guarantee, that every user/project has the same server types

There might be users with pre-view access to new server types, or custom server types for their usecases. I think this is mostly used internally, for example before the CAX servers launched the employees already had access to develop & test for them.

This is already an issue with the current hardcoded enum in the CRD, so whatever you are going to change its not going to get worse :D

Hardcoding it somewhere is probably the easiest. If its in the CRD, editors/IDEs can get the json schema and provide the user with autocomplete, not sure if any editor works well with this.

janiskemper:
So what would be your ideal solution? Not validating? If validating then only with an API call to fetch the "real" and up-to-date list? Having a more complex logic per location? Not validation but some good documentation so that users don't do mistakes? 
Currently, I'm not even sure if it is visible in a good way if a type that is not available is specified and the server cannot create it. Need to check

PR #1346, opened 2024-06-17
:seedling: Avoid misleading error logs when deleting hbmh

guettli:

Requeue after Deleting, and better log-message if HetznerCluster name is empty.


janiskemper (approved):
I'll approve, but plz update your PR description. This sounds a bit too much right now.

"Avoid misleading error logs when deleting hbmh" sounds more appropriate IMO

PR #1351, opened 2024-06-17
:seedling: Set providerID in phase ensure-provisioned

janiskemper:
Currently, we set the providerID of a baremetal server only after we confirmed that cloud init has successfully completed. However, CAPI also reacts on the result of cloud init, so there might be a temporary state when CAPI already reconciles the node object, CSRs are created, etc., but we have not yet set the providerID.

This leads for example to errors in the csr-controller, where we rely on the providerID to be set.

This PR sets the providerID already in ensure-provisioned state, so that these temporary errors should not occur anymore.

I think we will prefer this approach: https://github.com/syself/cluster-api-provider-hetzner/pull/1357

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -673,17 +673,27 @@ func (s *Service) setProviderID(ctx context.Context) error {
 		return fmt.Errorf("host not found for machine %s: %w", s.scope.Machine.Name, err)
 	}
 
-	if host.Spec.Status.ProvisioningState != infrav1.StateProvisioned {
+	// set providerID in phase "ensure-provisioned" and "provisioned" as from this point on the node might join the cluster
+	switch host.Spec.Status.ProvisioningState {
+	case infrav1.StateProvisioned:
+		providerID := providerIDFromServerID(host.Spec.ServerID)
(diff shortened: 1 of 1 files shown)


pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -673,17 +673,27 @@ func (s *Service) setProviderID(ctx context.Context) error {
 		return fmt.Errorf("host not found for machine %s: %w", s.scope.Machine.Name, err)
 	}
 
-	if host.Spec.Status.ProvisioningState != infrav1.StateProvisioned {
+	// set providerID in phase ensure provisioned and provisioned as from this point on the node might join the cluster
```
guettli, line 676: @janiskemper I suggest to use quotes to make it more readable: 

> set providerID in phase "ensure provisioned" and "provisioned" ...

janiskemper: fixed

guettli:
Just for the records, we cherry-picked this PR, and then the bm machines failed to join the cluster.

guettli:
This was merged instead: https://github.com/syself/cluster-api-provider-hetzner/pull/1357

PR #1359, opened 2024-06-21
:seedling: Reduce reconcilements

janiskemper:
- Filter HCloudMachine events so that status updates of the resource don't trigger another reconcilement of HCloudMachine
- Filter HetznerCluster events so that status updates of the resource don't trigger another reconcilement of HetznerCluster
- Filter HetznerCluster events so that some status updates that are not important don't trigger reconcilement of HCloudMachine
- Filter Machine events so that status updates of the resource don't trigger another reconcilement of HCloudMachine a
- Filter Cluster events so that status updates of the resource don't trigger another reconcilement of HetznerCluster
- Increase the time period to requeue if a server is starting
- Increase the time period to requeue if a server has not status yet
- Increase the time period to requeue if the kubeapi server is not yet reachable before adding it as target to the load balancer
- Don't trigger multiple shutdown calls for one server and increase the timeout to requeue after a shutdown is triggered

**TODOs**:
- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

Side note:
I will add an issue for unit testing the functions I have added after this is getting merged.

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -220,14 +220,16 @@ func (r *HCloudMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl
 		WithOptions(options).
 		For(&infrav1.HCloudMachine{}).
 		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(log, r.WatchFilterValue)).
+		WithEventFilter(IgnoreInsignificantHCloudMachineStatusUpdates(log)).
 		Watches(
 			&clusterv1.Machine{},
 			handler.EnqueueRequestsFromMapFunc(util.MachineToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("HCloudMachine"))),
+			builder.WithPredicates(IgnoreInsignificantMachineStatusUpdates(log)),
 		).
 		Watches(
 			&infrav1.HetznerCluster{},
 			handler.EnqueueRequestsFromMapFunc(r.HetznerClusterToHCloudMachines(ctx)),
-			builder.WithPredicates(IgnoreHetznerClusterConditionUpdates(log)),
+			builder.WithPredicates(IgnoreInsignificantHetznerClusterUpdates(log)),
 		).
 		Build(r)
 	if err != nil {
@@ -303,8 +305,8 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
 	}
 }
 
-// IgnoreHetznerClusterConditionUpdates is a predicate used for ignoring HetznerCluster condition updates.
-func IgnoreHetznerClusterConditionUpdates(logger logr.Logger) predicate.Funcs {
+// IgnoreInsignificantHetznerClusterUpdates is a predicate used for ignoring insignificant HetznerCluster.Status updates.
+func IgnoreInsignificantHetznerClusterUpdates(logger logr.Logger) predicate.Funcs {
 	return predicate.Funcs{
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			log := logger.WithValues(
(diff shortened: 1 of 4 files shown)


batistein (dismissed):

controllers/hcloudmachine_controller.go
```diff
@@ -355,3 +374,103 @@ func IgnoreHetznerClusterConditionUpdates(logger logr.Logger) predicate.Funcs {
...
+
+			// We should not modify the original objects, this causes issues with code that relies on the original object.
+			oldMachine = oldMachine.DeepCopy()
+			newMachine = newMachine.DeepCopy()
+
+			oldMachine.ManagedFields = nil
+			newMachine.ManagedFields = nil
+
+			oldMachine.ResourceVersion = ""
+			newMachine.ResourceVersion = ""
```
apricote, line 409: Is this actually a problem or is this filtered out by controller-runtime beforehand? Feels like it _should_ be ignored by controller-runtime.

controllers/hcloudmachine_controller.go
```diff
@@ -355,3 +374,103 @@ func IgnoreHetznerClusterConditionUpdates(logger logr.Logger) predicate.Funcs {
...
+			newHCloudMachine = newHCloudMachine.DeepCopy()
+
+			oldHCloudMachine.ManagedFields = nil
+			newHCloudMachine.ManagedFields = nil
+
+			oldHCloudMachine.ResourceVersion = ""
+			newHCloudMachine.ResourceVersion = ""
+
+			oldHCloudMachine.Status = infrav1.HCloudMachineStatus{}
+			newHCloudMachine.Status = infrav1.HCloudMachineStatus{}
```
apricote, line 462: if something wipes the status you usually want the controller to restore it. This way, no user-modifications of the status cause a reconcile.

apricote (commented):
Two notes about the predicates, otherwise this looks good and a step in the right direction. Would love to have some better options for finding/debugging this from controller-runtime.

controllers/hcloudmachine_controller.go
janiskemper: I just copy pasted the code from you :D didn't think more about it

controllers/hcloudmachine_controller.go
janiskemper: added this check

controllers/hcloudmachine_controller.go
apricote: Did a quick look through the controller-runtime code, but did not see any explicit handling of these fields, so we probably need to consider them here.

For managed fields there might be a solution in `controller-runtime` 0.18.0 (currently on 0.17.x). Someone added a cache transformer to remove these fields from the object cache. IIUC this cache is also used to evaluate the reconcile enqueing, so objects would be passed to the predicates without the managed fields, but this should be tested: https://github.com/kubernetes-sigs/controller-runtime/pull/2791


```patch
diff --git a/main.go b/main.go
index 4e8269ab..7a2a0321 100644
--- a/main.go
+++ b/main.go
@@ -123,6 +123,7 @@ func main() {
 			ByObject:          secretutil.AddSecretSelector(),
 			SyncPeriod:        &syncPeriod,
 			DefaultNamespaces: watchNamespaces,
+			DefaultTransform:  cache.TransformStripManagedFields(),
 		},
 	})
 	if err != nil {
```

apricote (approved):

controllers/hcloudmachine_controller.go
janiskemper: sounds interesting! We should keep this in mind for the CAPI 1.8 upgrade

batistein (approved):

guettli:
> Two notes about the predicates, otherwise this looks good and a step in the right direction. Would love to have some better options for finding/debugging this from controller-runtime.

@apricote what kind of debugging do you mean here? BTW, you can get the total reconcile counts per CRD:

```
k port-forward caph-controller-manager-XXX 8080:8080
```
```
curl localhost:8080/metrics | grep reconcile_total

controller_runtime_reconcile_total{controller="hcloudmachine",result="error"} 0
controller_runtime_reconcile_total{controller="hcloudmachine",result="requeue"} 0
controller_runtime_reconcile_total{controller="hcloudmachine",result="requeue_after"} 162
controller_runtime_reconcile_total{controller="hcloudmachine",result="success"} 59
...
```

apricote:
I would love to have a flag in controller-runtime `--debug-controller-reconciles` that show why a reconcile was triggered:

- Was it from an event? Was it because of the reconcile interval?
- If event, which event (kind, namespacedname)?
- Which predicates allowed this event? Which denied?

Some transparent insights into `handler.EnqueueRequestsFromMapFunc()`, `builder.WithPredicates()`, `WithEventFilter()` and friends.

Not something that you want running in our prod clusters, but something that can be enabled locally (or through some diagnostics API) to figure out why it reconciles more often then expected.

apricote:
One alternative would be distributed tracing for the Kubernetes API that includes any actions taken by controllers because of watches. But I guess that would be way harder and spans many more components.

guettli:
@apricote good questions. I have no answer to them, so I asked here to get more insights: https://kubernetes.slack.com/archives/C02MRBMN00Z/p1720524329698439

PR #1394, opened 2024-07-25
🌱 Set condition to false if server resource unavailable

Dhairya-Arora01:
**What this PR does / why we need it**: mark hcloud machine's "serverCreateSucceeded" condition to false if it throws server resource unavailable error

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


pkg/services/hcloud/server/server.go
```diff
@@ -465,8 +465,30 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
 				s.scope.Name(),
 				err,
 			)
-			return nil, errServerCreateNotPossible
+
+			err = errors.Join(errServerCreateNotPossible, err)
```
janiskemper, line 469: ah now I saw what this error is used for. I think that in both cases that we have there, we should return this error as well to not immediately reconcile. WDYT?

Dhairya-Arora01: seems logical ... then it would reque after 5 min...

janiskemper: so it this correctly implemented right now?

janiskemper: I guess it is. then we can resolve it

Dhairya-Arora01: Yes

janiskemper (approved):

PR #1407, opened 2024-07-29
:seedling: Install Cloud-Init-Data via post-install.sh

guettli:

We can reduce the number of reboots when provisioning bm server.

We install cloud-init-data via post-install.sh


pkg/services/baremetal/host/host.go
```diff
@@ -1140,10 +1140,35 @@ func (s *Service) actionImageInstalling() actionResult {
...
 
-	if postInstallScript != "" {
-		if err := handleSSHError(sshClient.CreatePostInstallScript(postInstallScript)); err != nil {
-			return actionError{err: fmt.Errorf("failed to create post install script %s: %w", postInstallScript, err)}
-		}
+	if postInstallScript == "" {
+		postInstallScript = "#!/bin/bash\n"
+	}
+
+	cloudInitData, err := s.scope.GetRawBootstrapData(context.TODO())
```
janiskemper, line 1147: we probably have to give the context now to the actionImageInstalling function right? Other option would be to store the bootstrap data somewhere, but I don't think that is valid

pkg/services/baremetal/host/state_machine.go
```diff
@@ -249,19 +257,6 @@ func (hsm *hostStateMachine) handleImageInstalling() actionResult {
...
 
 	actResult := hsm.reconciler.actionImageInstalling()
-	if _, ok := actResult.(actionComplete); ok {
-		hsm.nextState = infrav1.StateProvisioning
-	}
-	return actResult
-}
-
-func (hsm *hostStateMachine) handleProvisioning() actionResult {
-	if hsm.provisioningCancelled() {
```
janiskemper, line 259: this check might still be relevant. Should we really remove it?

guettli: @janiskemper 

This is handled here:

```
func (hsm *hostStateMachine) handleEnsureProvisioned() actionResult {
	if hsm.provisioningCancelled() {
		hsm.nextState = infrav1.StateDeprovisioning
		return actionComplete{}
	}
```

and here:

```
func (hsm *hostStateMachine) handleImageInstalling() actionResult {
	if hsm.provisioningCancelled() {
		hsm.nextState = infrav1.StateDeprovisioning
		return actionComplete{}
	}
```

I think that is enough. Or am I missing something?

pkg/services/baremetal/host/host.go
guettli: @janiskemper yes, good point. I will have a look if I can get the context there somehow.

pkg/services/baremetal/host/state_machine.go
janiskemper: maybe it just looked weird in the changes of the PR. but yes, this is what is necessary!

pkg/services/baremetal/host/host.go
guettli: @janiskemper I saw two ways to get the context. I asked the experts, and did it accordingly: https://www.reddit.com/r/golang/comments/1efotre/store_ctx_in_a_struct/

guettli: @janiskemper I updated the code. OK for you?

pkg/services/baremetal/host/state_machine.go
guettli: ok for you?

pkg/services/baremetal/host/host.go
janiskemper: yes

janiskemper (dismissed):

pkg/services/baremetal/host/host.go
```diff
@@ -1188,7 +1193,12 @@ EOF_POST_INSTALL_SCRIPT
 		record.Warnf(s.scope.HetznerBareMetalHost, "ExecuteInstallImageFailed", out.StdOut)
 		return actionError{err: fmt.Errorf("failed to execute installimage: %w", out.Err)}
 	}
-
+	if !strings.Contains(out.StdOut, postInstallScriptFinished) {
+		record.Warnf(s.scope.HetznerBareMetalHost, "ExecuteInstallImageFailed", out.StdOut)
+		return actionError{err: fmt.Errorf("did not find marker %q in stdout. Failed to execute installimage: %w",
+			postInstallScriptFinished,
+			out.Err)}
```
janiskemper, line 1206: out. Err is nil at this point. This will create a nil pointer error afaik with fmt.Errorf

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
guettli: @janiskemper thank you for your sharp eyes! Yes, the err-case is handled a line above with "return".

pkg/services/baremetal/host/host.go
```diff
@@ -1511,29 +1431,35 @@ func (s *Service) actionEnsureProvisioned() (ar actionResult) {
 		if err != nil {
 			return actionError{err: fmt.Errorf(errMsgFailedHandlingIncompleteBoot, err)}
 		}
+		// Often: err: 'failed to dial ssh. Error message: dial tcp 94.130.206.211:22: i/o timeout. DialErr: failed to dial ssh'
```
guettli, line 1434: @janiskemper I noticed that this line gets reached if the ssh port is not reachable yet. This is not a big problem. But I think this means that the above line to detect "ssh not reachable" don't work reliably.

janiskemper: i/o timeout should be a timeout error. Is it not declared as such?

guettli: all I can say is that the code reaches this lines, and that it creates the output I added as a comment.

Overall I don't think it makes sense to distinguish between connection-refused and a timeout. A different Linux distribution could return different errors.

I think we should leave it like this in this PR. If we want to, we can create a second PR which handles that better.

@janiskemper what do you think?

pkg/services/baremetal/host/host.go
```diff
@@ -1112,11 +1122,9 @@ func (s *Service) actionImageInstalling() actionResult {
...
 	}
 	if out.StdOut != "" {
-		record.Warnf(s.scope.HetznerBareMetalHost, "InstallImageAlreadyRunning",
-			"installimage is already running:\n%s", out.StdOut)
-		return actionContinue{
-			delay: 10 * time.Second,
-		}
+		record.Warn(s.scope.HetznerBareMetalHost, "InstallImageAlreadyRunning",
+			"installimage is already running.")
+		return actionError{err: fmt.Errorf("installimage is already running")}
```
janiskemper, line 1127: why this change?

pkg/services/baremetal/host/host.go
```diff
@@ -1160,10 +1196,14 @@ func (s *Service) actionImageInstalling() actionResult {
 	// Execute install image
 	out = sshClient.ExecuteInstallImage(postInstallScript != "")
 	if out.Err != nil {
-		record.Warnf(s.scope.HetznerBareMetalHost, "ExecuteInstallImageFailed", out.StdOut)
+		record.Warnf(s.scope.HetznerBareMetalHost, "ExecuteInstallImageFailed", fmt.Sprintf("Error: %s. %s", out.Err.Error(), out.StdOut))
```
janiskemper, line 1199: just asking: are you sure that there is nothing in StdErr?

pkg/services/baremetal/host/host.go
janiskemper: sorry I don't get it.. This error just means that we need to wait a bit more because the boot process didn't finish.. We will obviously get errors if the server is not reachable. Doesn't really matter whether it's i/o timeout or connection refused. 

All we do, if we haven't reached any time limits, which I believe are handled in handleIncompleteBoot, is to requeue and try later. Which is correct in all cases. 

What you suggest regarding connection refused and timeout errors and to not look at them in such detail, is completely independent of this.

pkg/services/baremetal/host/host.go
```diff
@@ -1511,29 +1433,35 @@ func (s *Service) actionEnsureProvisioned() (ar actionResult) {
...
 
 	// Check the status of cloud init
-	actResult, _ := s.checkCloudInitStatus(sshClient)
+	actResult, msg, _ := s.checkCloudInitStatus(sshClient)
 	if _, complete := actResult.(actionComplete); !complete {
+		record.Event(s.scope.HetznerBareMetalHost, "CloudInitStillRunning", msg)
 		return actResult
 	}
-
+	s.scope.Logger.Info("ensureProvisioned: completed.", "msg", msg)
```
janiskemper, line 1449: plz remove

pkg/services/baremetal/host/host.go
```diff
@@ -1548,10 +1476,11 @@ func (s *Service) handleConnectionRefused() actionResult {
 		Port:       s.scope.HetznerBareMetalHost.Spec.Status.SSHSpec.PortAfterInstallImage,
 		IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
 	})
-	actResult, err := s.checkCloudInitStatus(oldSSHClient)
+	actResult, msg, err := s.checkCloudInitStatus(oldSSHClient)
 	// If this ssh client also gives an error, then we go back to analyzing the error of the first ssh call
 	// This happens in the statement below this one.
 	if err == nil {
+		s.scope.Logger.Info("handleConnectionRefused", "msg", msg, "err", err)
```
janiskemper, line 1483: plz remove

janiskemper (changes requested):
plz only keep log statements that are very relevant for debugging. I have marked probably all that are not, because they just indicate that things go well.

pkg/services/baremetal/host/host.go
guettli: If InstallImage is already running, then there is something strange. Running installimage again can work, but if the first installimage run failed, then the second try will fail with `ERROR unmounting device(s):`. See https://github.com/syself/cluster-api-provider-hetzner/pull/1407/files#diff-d1bdf47f722af9b77810d5839d1927cf67585d79b9f74cb5ce2c7586f4e9937eR265 

That's why I think it is better to start with getting into the rescue system again.

janiskemper: but is it necessary at all, if you always return back to the registering state, to check for running install image processes at all? is there a case where this can still happen?

guettli: @janiskemper if the controller restarts, then there can be an old install-image process still running.

pkg/services/baremetal/host/host.go
guettli: @janiskemper good point. Pretty printing out.Stdout, out.Stderr, out.Err is "hard". I created a method: Output.String(). Please have a look: https://github.com/syself/cluster-api-provider-hetzner/pull/1407/commits/0c86f783f905d82667159029923fcbc997093bfd

If I like it, then we can use that method in several places. But maybe in a second PR.

pkg/services/baremetal/host/host.go
guettli: removed.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: I removed the comment and the logger.Info().

pkg/services/baremetal/host/host.go
janiskemper: so if a running comand via ssh gets interrupted by a container restart? Or is it normal to start the operation (in the background) , then finish the control loop and start it again?

pkg/services/baremetal/host/host.go
janiskemper: yes I like it!

pkg/services/baremetal/host/host.go
guettli: > so if a running comand via ssh gets interrupted by a container restart?

Yes, then the install-image process on the bm-server will keep on running. The remote process won't notice that the ssh client has terminated. I added a comment to make this edge-case more obvious. It often happens via Tilt, but in production it can happen, too.

The Reconcile method call should terminate quickly. Installimage takes roughly two minutes. That's ok.

But if a process takes much longer, then we should handle it differently. Maybe via a Kubernetes-Job.

janiskemper (commented):
LGTM. Do you have anything else you wanna add to this PR?

guettli:
> LGTM. Do you have anything else you wanna add to this PR?

I added a warning, so that reconciles which take too long are visible (already pushed).

And I want to see that the e2e are fine. They failed some minutes ago. I guess this was flakiness.

pkg/services/baremetal/host/host.go
guettli: Doing install-image async makes sense, I created a new issue for that: https://github.com/syself/cluster-api-provider-hetzner/issues/1408

janiskemper (approved):

PR #1413, opened 2024-08-01
🌱Add check to disallow hcloud servers without ssh key

### LINKED ISSUE #1402: Disallow HCloud servers without SSH key
/kind feature

No SSH key should be usually (otherwise please comment) a configuration issue. We can easily catch that and make users aware of this.

This will introduce a breaking change and will affect everyone provisioning HCloud servers without ssh key.

However, I think the drawbacks are not as severe and the advantage will be that no-one can misconfigure it.

We will add a condition to HCloudMachine that there is no ssh key.

The ssh key can be provided by the feature of adding it to the Hetzner secret even on the fly, without changing any templates.

yrs147:

Fixes #1402 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -415,6 +415,17 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
 		}
 	}
 
+	if len(sshKeySpecs) == 0 {
+		conditions.MarkFalse(
+			s.scope.HCloudMachine,
+			infrav1.ServerCreateSucceededCondition,
+			infrav1.SSHKeyNotFoundReason,
+			clusterv1.ConditionSeverityError,
+			"No SSH key configured for the server. Please add an SSH key",
+		)
+		return nil, errServerCreateNotPossible
+	}
+
 	// get all ssh keys that are stored in HCloud API
 	sshKeysAPI, err := s.scope.HCloudClient.ListSSHKeys(ctx, hcloud.SSHKeyListOpts{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -435,6 +435,17 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
 		return nil, errServerCreateNotPossible
 	}
 
+	if len(sshKeySpecs) == 0 {
+		conditions.MarkFalse(
+			s.scope.HCloudMachine,
+			infrav1.ServerCreateSucceededCondition,
+			"NoSSHKeyFound",
```
janiskemper, line 442: can you add the reason to the api folder where the other reasons are defined?

pkg/services/hcloud/server/server.go
```diff
@@ -435,6 +435,17 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
 		return nil, errServerCreateNotPossible
 	}
 
+	if len(sshKeySpecs) == 0 {
```
janiskemper, line 438: this check can be done before you fetch the ssh keys from the API

pkg/services/hcloud/server/server.go
```diff
@@ -435,6 +435,17 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
...
+		conditions.MarkFalse(
+			s.scope.HCloudMachine,
+			infrav1.ServerCreateSucceededCondition,
+			"NoSSHKeyFound",
+			clusterv1.ConditionSeverityError,
+			"No SSH key configured for the server. Please add an SSH key ",
+		)
+		return nil, fmt.Errorf("SSHKeySpec is empty")
```
janiskemper, line 446: return errServerCreateNotPossible

janiskemper (changes requested):

pkg/services/hcloud/server/server.go
yrs147: I saw there is already a reason for that `infrav1.SSHKeyNotFoundReason`

janiskemper (approved):

janiskemper (commented):
Sorry, forgot the unit tests. I believe that we can test this. There should be existing tests probably for related cases. Can you check? Should be about that we set the condition when needed

PR #1419, opened 2024-08-05
:seedling: Avoid an Event GetCloudInitOutputFailed. Wait until machine is reachable.

guettli:
Up to now you always got a "useless" warning event "GetCloudInitOutputFailed", just because the machine was not immediately reachable via ssh.


pkg/services/baremetal/host/host.go
```diff
@@ -1437,6 +1409,26 @@ func (s *Service) actionEnsureProvisioned(_ context.Context) (ar actionResult) {
...
+		case actionContinue:
+			return
+		case actionError:
+			if errors.Is(v.err, errGettingCloudInitOutput) {
+				return
+			}
+			errMsg = v.err.Error()
+		}
+		out := sshClient.GetCloudInitOutput()
+		record.Warnf(s.scope.HetznerBareMetalHost, "CloudInitOutput", "cloud init output (%s):\n%s",
```
janiskemper, line 1427: why do you use "warnf"?

and why do we use the defer statement at all btw? Why cannot we create it in the normal flow? The defer makes it harder to understand. Right now I also don't understand why we call the function once inside the defer and once outside

guettli: > why do you use "warnf"?

I use Warnf() because I use formatting in the string (`%s`).

> and why do we use the defer statement at all btw? Why cannot we create it in the normal flow? The defer makes it harder to understand. Right now I also don't understand why we call the function once inside the defer and once outside

Good point. I did it like this, because it was done like that before.

I look if this can be done without defer.

guettli: @janiskemper I updated the PR.

janiskemper (approved):

guettli:
Just for the records, I provisioned two machines with the current state and it worked. No Warning Event was created.

PR #1427, opened 2024-08-08
:seedling: Set PermanentErrorAnnotation, if removed by user, then remove the error state.

### LINKED ISSUE #1335: Annotation to control permanent error of HetznerBareMetalHost
/kind feature

There are multiple reasons why a HetznerBareMetalHost object has a "permanent error" set. This means that user intervention is required.

The current work flow for users that have taken action on the server so that it can be used again to delete the custom resource and to recreate it. This is no issue at all from the workflow. However, it is not very elegant to delete an object and to manually re-create it. 

Therefore, we can use an annotation that is set on the HetznerBareMetalHost by the controller and that can be removed from the user to indicate that the reason for the permanent error is gone and that we again want to use and provision the server. 

So instead of deleting the HetznerBareMetalHost, the user can just delete the annotation and the controller would take care of the rest.

guettli:
Fixes #1335

> There are multiple reasons why a HetznerBareMetalHost object has a "permanent error" set. This means that user intervention is required.

> The current work flow for users that have taken action on the server so that it can be used again to delete the custom resource and to recreate it. This is no issue at all from the workflow. However, it is not very elegant to delete an object and to manually re-create it.

### DIFF
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -27,6 +27,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/record"
 )
 
 const (
@@ -550,6 +551,8 @@ func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string)
 	host.Spec.Status.ErrorMessage = errMessage
 	if errType == PermanentError {
 		host.Annotations[PermanentErrorAnnotation] = time.Now().Format(time.RFC3339)
+		record.Warnf(host, "PermanentErrorSet", "Remove annotation %q, if you want the controller to use the hbmh again.",
+			PermanentErrorAnnotation)
 	}
 }
 
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -547,6 +548,9 @@ func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string)
 	}
 	host.Spec.Status.ErrorType = errType
 	host.Spec.Status.ErrorMessage = errMessage
+	if errType == PermanentError {
+		host.Annotations[PermanentErrorAnnotation] = time.Now().Format(time.RFC3339)
```
janiskemper, line 552: is it worth using a utility or checking if annotations are empty?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -97,6 +97,18 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// Remove permanent error, if the corresponding annotation was removed by the user.
+	requeue := removePermanentErrorIfAnnotationIsGone(bmHost)
+	if requeue {
```
janiskemper, line 102: this part looks a bit weird. Why do we update the host after requeue is true?
Maybe we can rename the variables and functions to improve the flow. "removed" instead of "requeue" for example would make it a lot better IMO

guettli: @janiskemper changed in 5bf44503 - better now?

api/v1beta1/hetznerbaremetalhost_types.go
guettli: @janiskemper good point, I updated the code to check for nil. See 72fbd26f - ok now?

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -17,11 +17,20 @@ limitations under the License.
...
 
 import (
+	"testing"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
```
janiskemper, line 27: this will mess with the linting. YOu need to put the caph dependencies below. Maybe make format or something should solve this?

guettli: I fixed that via `make format-golang`.

janiskemper (approved):

PR #1428, opened 2024-08-09
:seedling: Always create a Condition, if createServer failed (hcloud)

### LINKED ISSUE #1414: Improve error handling of HCloudClient.CreateServer  call
/kind feature

First, we can always return the error "errServerCreateNotPossible" if the CreateServer API call fails. In this way, we might have some edge cases where we can immediately retry, but it is safer in general to return the error. This error will lead us to wait for 5 minutes. 

Additionally, we can set the condition ServerCreateSucceeded on false with the error as message. We do that already in certain cases for specific errors, but it also makes sense to show unknown / not explicitly handled errors in the conditions, because that's where users look. 

This will make explicit handling of error codes, as we do it now, redundant, because we will handle all errors in the same way. 

Only rate limit exceeded errors should be handled separately, because in that case we set another condition.

guettli:

If this error happens (events):

image

Up to now, no Condition gets set:

image

This PR updates to code, so that always a Condition get set (if createServer failed)

fixes #1414


janiskemper (changes requested):
this is related to https://github.com/syself/cluster-api-provider-hetzner/issues/1414. Please consider that issue description

controllers/hetznerbaremetalremediation_controller_test.go
```diff
@@ -257,7 +257,7 @@ var _ = Describe("HetznerBareMetalRemediationReconciler", func() {
 					Expect(testEnv.Cleanup(ctx, hetznerBareMetalRemediation, hetznerBaremetalMachine)).To(Succeed())
 				})
 
-				It("should not remediate if no annotations is present in the hetznerBaremetalMachine", func() {
+				It("should not remediate if no annotations is present in the hetznerBaremetalMachine (flaky)", func() {
```
guettli, line 260: @janiskemper 
 This test was flaky, so I marked it accordingly

guettli:
> this is related to #1414. Please consider that issue description

nice, this makes it simpler. I updated the PR.

janiskemper (approved):

PR #1429, opened 2024-08-09
:seedling: WipeDisk annotation

guettli:
Introduced annotation "wipedisk.hetznerbaremetalhost.infrastructure.cluster.x-k8s.io"

Set it to a space seperated list of WWNs or "all".

The given disks will be erased with `wipefs -af` before provisioning.

Once wipefs was executed the annotation gets removed.


controllers/hcloudmachine_controller_test.go
```diff
@@ -377,7 +377,7 @@ var _ = Describe("HCloudMachineReconciler", func() {
 					}
 
 					return len(servers) == 0
-				}, timeout, interval).Should(BeTrue())
+				}, 2*timeout, interval).Should(BeTrue())
```
guettli, line 380: @janiskemper the test fais here often:

>   [FAIL] HCloudMachineReconciler Basic test correct server [It] creates the HCloud machine in Hetzner 1 (flaky)


[job](https://github.com/syself/cluster-api-provider-hetzner/actions/runs/10316952420/job/28560210828#step:7:1840)

I hope that the increased timeout will help.

janiskemper: It is about this: By("checking that no servers exist")

Therefore, an increase in timeout wouldn't lead to anything. Even the eventually is probably wrong in the first place

guettli: The error message says line 380, which is exactly the line where I added  `2*":

```go
				Eventually(func() bool {
					servers, err := hcloudClient.ListServers(ctx, hcloud.ServerListOpts{
						ListOpts: hcloud.ListOpts{
							LabelSelector: utils.LabelsToLabelSelector(map[string]string{hetznerCluster.ClusterTagKey(): "owned"}),
						},
					})
					if err != nil {
						return false
					}

					return len(servers) == 0
				}, 2*timeout, interval).Should(BeTrue())
```

guettli: Since I added `2*`, I have not seen the test failing again.

janiskemper: then you explain why this should be the case

guettli: Maybe it is because github runners run on shared CPUs, and a delay of some seconds is always possible. 

All I can say: I have seen this error several times during the last weeks. I have not seen it after increasing the timeout.

janiskemper (approved):

PR #1442, opened 2024-08-14
:seedling: Extend releasing docs: Check for assets

guettli:

Extend releasing docs: Check for assets


controllers/csr_controller.go
```diff
@@ -310,8 +310,6 @@ func getServerIDFromConstantHostname(ctx context.Context, csrUsername string, cl
 	// example csrUsername: system:node:bm-my-cluster-1234567
 	matches := constantBareMetalHostnameRegex.FindStringSubmatch(strings.TrimPrefix(csrUsername, nodePrefix))
 	if len(matches) != 3 {
-		log.V(1).Info("No constant baremetal hostname - regex does not match CSR username",
```
batistein, line 313: @guettli i think this was added by mistake?

guettli: I would like to merge this tiny change. When I see the line in the logs, I often think that there is something wrong. I think we can remove it. Should I create a new PR for that?

guettli: I created a new pr for that. https://github.com/syself/cluster-api-provider-hetzner/pull/1449

batistein (approved):

PR #1459, opened 2024-08-21
:seedling: fix flaky unit-tests

guettli:
Avoid global hcloudClient. The hcloudClient got created once, and stored in a global variable. If the previous test left a fake hcloud-server in the client, then the next test sees a different result from hcloudClient.ListServers().

### DIFF
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -34,7 +34,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
-	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
 	"github.com/syself/cluster-api-provider-hetzner/test/helpers"
 )
 
@@ -45,10 +44,9 @@ const (
 )
 
 var (
-	testEnv      *helpers.TestEnvironment
-	hcloudClient hcloudclient.Client
-	ctx          = ctrl.SetupSignalHandler()
-	wg           sync.WaitGroup
+	testEnv *helpers.TestEnvironment
+	ctx     = ctrl.SetupSignalHandler()
+	wg      sync.WaitGroup
 
 	defaultPlacementGroupName = "caph-placement-group"
 	defaultFailureDomain      = "fsn1"
@@ -64,7 +62,6 @@ var _ = BeforeSuite(func() {
 	utilruntime.Must(clusterv1.AddToScheme(scheme.Scheme))
 
 	testEnv = helpers.NewTestEnvironment()
-	hcloudClient = testEnv.HCloudClientFactory.NewClient("")
 
 	wg.Add(1)
 
--- a/controllers/hcloudmachine_controller_test.go
+++ b/controllers/hcloudmachine_controller_test.go
@@ -21,6 +21,8 @@ import (
 	"testing"
 	"time"
 
+	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
+
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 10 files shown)


janiskemper (dismissed):
this initializing won't make a difference. The difference will be made if you call "close": https://github.com/syself/cluster-api-provider-hetzner/blob/main/pkg/services/hcloud/client/fake/hcloud_client.go#L59

guettli:
> extended explanation: this unfortunately won't help. What would help is initializing one factory per test, but that is almost impossible, since you need that factory shared across all controllers, i.e. your whole envtest environment.

Thank you for your hint. 

Since I avoid global variables as much as I can, I was not thinking that the global hcloudClient used a second global variable.

Afaik there is no way to avoid a global variable, if you want to initialize something in BeforeSuite.

I changed the PR. The former Close() method is not called Reset() and called before the tests get executed.

Example:

```go
			BeforeEach(func() {
				hcloudClient = testEnv.ResetAndGetHCloudClient()
```

This way we are sure that every test resets the cache.

Is that ok for you?

janiskemper (commented):
This will work. However, I don't really like the naming. "Reset" is, as you write, not a general usable method, but only related to the fact that we clear a cache that we have in the fake client. 

Therefore, I would prefer to have the "close" function, which is a general pattern, and use it in the "aftereach" function instead of the "beforeeach". 

If you don't want to change that, then okay. 

Are you sure though that we cannot use the global hcloudclient? I think you always get a seemingly new one in the individual tests, even though they aren't actually new, because they share a same global variable. Therefore, I think it would be less misleading if we actually do the same and keep the hcloudclient a global variable in the controllers/ tests. If I understand it correctly, it shouldn't change anything. 

Please just respond to the above points and then we can get this merged. It is anyway a good conclusion that you had with investigating the flaky tests!

guettli:
> This will work. However, I don't really like the naming. "Reset" is, as you write, not a general usable method, but only related to the fact that we clear a cache that we have in the fake client.
> 
> Therefore, I would prefer to have the "close" function, which is a general pattern, and use it in the "aftereach" function instead of the "beforeeach".

> If you don't want to change that, then okay.

I have reasons for that :-)

I have seen these conversations in previous jobs. I think it is better if a test ensures during beforeEach (in other frameworks called SetUp) that everything is the way the tests want it to be.
A different test could fail badly, so that the test was not able to clean up after itself. Then maybe 20 tests will fail because the precondition does not fit. Then it is hard to detect what was wrong. That's why I prefer to ensure a clean test environment in beforeEach.

> Are you sure though that we cannot use the global hcloudclient? I think you always get a seemingly new one in the individual tests, even though they aren't actually new, because they share a same global variable. Therefore, I think it would be less misleading if we actually do the same and keep the hcloudclient a global variable in the controllers/ tests. If I understand it correctly, it shouldn't change anything.

I like to avoid a global client, because the explit way makes it is less likely that a developer forgets to call Reset().

> Please just respond to the above points and then we can get this merged. It is anyway a good conclusion that you had with investigating the flaky tests!

@janiskemper Was I able to change your mind?

janiskemper:
I think you didn't get my point :D Because I agree with you in general, however:
- we have a global client whether you want it or not. By defining new clients constantly you just always overwrite the same global client. You just pretend that you define new clients. That's what IMO confuses people and makes them believe (just as you did some days ago) that there is no global client and we can just redefine it however we want. I don't want to pretend that there is no global client.
- I also in theory agree with this BeforeEach is better than AfterEach. However, also doesn't matter here, as it is a global client that is shared among all. In theory, which is not the case for us, if you run these in parallel, they would reset the global client during the test of another one. But there I have no strong feelings. Here I am not happy with the naming, which is better when we do it at the end. But that is okay. The above point is the more relevant one

guettli:
> I think you didn't get my point :D Because I agree with you in general, however:
> 

> * we have a global client whether you want it or not. By defining new clients constantly you just always overwrite the same global client. You just pretend that you define new clients. That's what IMO confuses people and makes them believe (just as you did some days ago) that there is no global client and we can just redefine it however we want. I don't want to pretend that there is no global client.

This variable is no longer global:

```
hcloudClient hcloudclient.Client
```

And the name of the method (`testEnv.ResetAndGetHCloudClient()`) make it very clear that something gets re-used.

@janiskemper I renamed the method to ResetAndGetGlobalHCloudClient. This makes it more explicit. What do you think?

> * I also in theory agree with this BeforeEach is better than AfterEach. However, also doesn't matter here, as it is a global client that is shared among all. In theory, which is not the case for us, if you run these in parallel, they would reset the global client during the test of another one. But there I have no strong feelings. Here I am not happy with the naming, which is better when we do it at the end. But that is okay. The above point is the more relevant one

> ...if you run these in parallel,...

My changes are not related to that. This is a different topic. If we want parallel execution, we need to call the tests via the ginkgo binary. It uses different processes, so that a global variable would be ok.

Dhairya-Arora01 (approved):

guettli:
I was not aware how the nested BeforeEach() work. IIRC it is enough to do the Reset once in the outer-most BeforeEach. Then it is ok if someone forgets this and he/she adds a new block. Which means it is ok to have the global variable again. I am happy that you like it.

PR #1471, opened 2024-08-31
🐛 Make e2e scripts also work on mac

johannesfrey:
I had problems running `make test-e2e` as described in the [dev docs](https://syself.com/docs/caph/developers/development-guide#running-local-e2e-test) on MacOS mostly because certain essential command binaries used in the bash scripts work differently there. This PR should make them work on both platforms.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper:
thanks @johannesfrey ! 

I'm a bit worried about "mac support" that we won't sustain. 

@batistein @guettli do you think this is feasible? We may not be able to test it, but we can change to commands working on both mac and linux, as Johannes has done. 

I see the following options:
a) we merge this and if there are further issues in the future, we expect further contributions from the community
b) we add guidelines to use commands that work for both in general and try following them
c) we don't merge this as we don't even want to indicate that we support mac. 

I don't think c) is necessary, but I'm looking forward to your thoughts!

guettli:
@johannesfrey I see that you work for Loft. That's cool. Now why do you not just start a devpod, and then you have full Linux environment. Or did I misunderstand what this tool of Loft is all about?

PR #1472, opened 2024-08-31
✨ Bring your own network

### LINKED ISSUE #762: Make it possible to use a pre-created private network
/kind feature

I'm attempting to spin up a cluster that is entirely on a private network that has already been created. Unfortunately, I'm met with a uniqueness error:
```
"error":"failed to reconcile network for HetznerCluster rbjorklin-com/rbjorklin-com: failed to create network: error creating network: name is already used (uniqueness_error)"
```

I have created a network with a single VM inside it. From this VM I'm attempting to bootstrap a cluster with a private-only network. Does this make sense? Do you foresee any issues doing this?

**Environment:**

- cluster-api-provider-hetzner version: v1.0.0-beta.15
- Kubernetes version: v1.26.3
- OS (e.g. from `/etc/os-release`): Fedora 37

johannesfrey:
This PR makes it possible to "adopt" a pre-existing network by passing its ID to `hetznerCluster.spec.hcloudNetwork.id` instead of the network being created during cluster creation. Furthermore, during cluster deletion it only deletes the attached network if it does not have the `owned` label attached to it (currently the only way here to discriminate between a CAPH-managed network and an unmanaged one).

Fixes #762

This has been lingering around for a while untouched on my fork and I decided to rebase it onto the current main branch. Please consider this as a first attempt to approach this topic as a whole. I also tried to already add some unit tests. I guess it also might require some e2e tests!? No idea if this is the desired way to do this and about other side-effects I did not see. So looking forward for feedback or any pointers. And also feel free to push changes to the PR, as I'll be pretty occupied with other things almost the whole September. Just wanted to push this out there already for you to take a look at 🙂 

The most controversial changes so far:
- Making `hcloudNetwork.id` mutually exclusive with `cidrBlock`, `subnetCidrBlock` and `networkZone`
- Replacing kubebuilder defaulting/validation special tags with custom defaulting/validation, which makes the webhook more complex
- Changing `cidrBlock`, `subnetCidrBlock` and `networkZone` to be pointers (I guess this could also be done with empty strings, but pointers make it possible to be not shown at all, when not provided)
- Labels are shown in the NetworkStatus in order to check if it's managed or unmanaged

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [x] add unit tests


api/v1beta1/hetznercluster_webhook.go
```diff
@@ -109,6 +149,55 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
 		if err := isNetworkZoneSameForAllRegions(r.Spec.ControlPlaneRegions, nil); err != nil {
 			allErrs = append(allErrs, err)
 		}
+	} else {
+		// If ID is given check that all other network settings are empty.
+		if r.Spec.HCloudNetwork.ID != nil {
```
janiskemper, line 154: what if we do the following: We ignore all other properties if the ID is specified and overwrite it with the correct value from the actual network? Seems a bit unnecessary to me to have all this validation, because in the end "ignoring and overwriting" would be also fine I guess. Then we also don't have to change to pointers

I'm okay with both ways though!

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -109,6 +149,55 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
...
+			if errs := areCIDRsAndNetworkZoneEmpty(r.Spec.HCloudNetwork); errs != nil {
+				allErrs = append(allErrs, errs...)
+			}
+		} else {
+			// If no ID is given check the other network settings for valid entries.
+			if r.Spec.HCloudNetwork.NetworkZone != nil {
```
janiskemper, line 160: this is a validation that we don't have right now I think. It seems to me as if we should add a separate PR for that, as this part is relevant independent of the purpose of this PR!

janiskemper (commented):
In general this approach seems good to me. Thanks a lot for this contribution!

@guettli @batistein what's your opinion?

api/v1beta1/hetznercluster_webhook.go
guettli: @janiskemper 

> what if we do the following: We ignore all other properties if the ID is specified and overwrite it with the correct value from the actual network?

I would like to have a clear separation between Spec and Status. We can write the actual values into the Status, but I think writing them to the Spec can create confusion (and is against the api guidelines of Kubernetes).

janiskemper: happens for example with the ProviderID as well @guettli

guettli: > happens for example with the ProviderID as well @guettli

Yes, I know that there are exceptions to the api guidelines. Nevertheless, the guidelines make sense, and we should break them (except there are good reasons).

janiskemper: alright, it happens in multiple places (providerID, ControlPlaneEndpoint, etc.) but I am fine to not do it in this case, because we currently don't do it for the network.

controllers/hetznercluster_controller.go
```diff
@@ -344,9 +344,15 @@ func (r *HetznerClusterReconciler) reconcileDelete(ctx context.Context, clusterS
 		return reconcile.Result{}, fmt.Errorf("failed to delete load balancers for HetznerCluster %s/%s: %w", hetznerCluster.Namespace, hetznerCluster.Name, err)
 	}
 
-	// delete the network
-	if err := network.NewService(clusterScope).Delete(ctx); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to delete network for HetznerCluster %s/%s: %w", hetznerCluster.Namespace, hetznerCluster.Name, err)
+	// delete the network only if it is owned by us
```
janiskemper, line 347: this logic should be integrated into the Delete function that we call from the "network" package. Here we should just do a simple function call

pkg/services/hcloud/network/network.go
```diff
@@ -154,6 +159,24 @@ func (s *Service) Delete(ctx context.Context) error {
...
+	if id != nil {
+		network, err := s.scope.HCloudClient.GetNetwork(ctx, *id)
+		if err != nil {
+			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "GetNetwork")
+			return nil, fmt.Errorf("failed to get network %d: %w", *id, err)
+		}
+
+		if network != nil {
+			if len(network.Subnets) > 1 {
+				return nil, fmt.Errorf("multiple subnets not allowed")
```
janiskemper, line 173: this should not be an error, but we rather should set an appropriate condition on the HetznerCluster object, create an event, and return nil here.

api/v1beta1/types.go
```diff
@@ -217,32 +217,39 @@ type LoadBalancerTarget struct {
...
+	// +optional
+	ID *int64 `json:"id,omitempty"`
+
 	// Enabled defines whether the network should be enabled or not.
 	Enabled bool `json:"enabled"`
 
-	// CIDRBlock defines the cidrBlock of the HCloud Network. If omitted, default "10.0.0.0/16" will be used.
-	// +kubebuilder:default="10.0.0.0/16"
+	// CIDRBlock defines the cidrBlock of the HCloud Network.
+	// Defaults to "10.0.0.0/16".
```
janiskemper, line 229: let's maybe add here "webhook defaults ...". I was a bit confused right now before realizing that this is no kubebuilder command as well

janiskemper (changes requested):
Thanks a lot again for this PR @johannesfrey. I think we can merge it if you follow the suggestions I gave. It's really good work!

api/v1beta1/hetznercluster_webhook.go
janiskemper: I think it's good to have it.

johannesfrey:
> Thanks a lot again for this PR @johannesfrey. I think we can merge it if you follow the suggestions I gave. It's really good work!

Sorry for the long delay. I hope I addressed your suggestions correctly. PTAL. Thx!

pkg/services/hcloud/network/network.go
johannesfrey: I hope I understood you correctly here. I set the NetworkReadyCondition to false in that case. Also what do you mean with "create an event"?

api/v1beta1/hetznercluster_webhook.go
johannesfrey: I assume there was validation for this before, because of following kubebuilder markers, which would enforce the API server to perform the validations for the resource:
```
	// +kubebuilder:validation:Enum=eu-central;us-east;us-west;ap-southeast
	// +kubebuilder:default=eu-central
```

I think that the manual validation is now necessary because of the validation now being dependent on the new ID field.

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -60,10 +75,35 @@ func (r *HetznerClusterList) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+// Default implements webhook.CustomDefaulter so a webhook will be registered for the type.
+func (r *HetznerCluster) Default(_ context.Context, obj runtime.Object) error {
+	hetznerclusterlog.V(1).Info("default", "name", r.Name)
+
+	cluster, ok := obj.(*HetznerCluster)
+	if !ok {
+		return apierrors.NewBadRequest(fmt.Sprintf("expected an HetznerCluster but got a %T", obj))
+	}
+
+	if cluster.Spec.HCloudNetwork.Enabled {
```
janiskemper, line 89: if not enabled then return nil

-> in this way we omit the nested if statement

api/v1beta1/hetznercluster_webhook.go
janiskemper, line 160: what is supposed to happen if this is nil? 

I just wanted to say that we should again avoid the nested if statements if possible and that we can do that here. 

Currently, we don't handle the case where NetworkZone is nil at all. Shouldn't we though? Isn't that a problem? 

Anyway, let's avoid the nested if statement and if you say we should, then let's handle the case where NetworkZone is nil

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -109,6 +149,55 @@ func (r *HetznerCluster) ValidateCreate() (admission.Warnings, error) {
...
+				if !validNetworkZone {
+					allErrs = append(allErrs, field.Invalid(
+						field.NewPath("spec", "hcloudNetwork", "networkZone"),
+						r.Spec.HCloudNetwork.NetworkZone,
+						"wrong network zone. Should be eu-central, us-east, us-west or ap-southeast"),
+					)
+				}
+			}
+
+			if r.Spec.HCloudNetwork.CIDRBlock != nil {
```
janiskemper, line 179: again here and below it should never be nil afaik. Should we return errors?

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -161,13 +250,46 @@ func (r *HetznerCluster) ValidateUpdate(old runtime.Object) (admission.Warnings,
...
-	// Network settings are immutable
-	if !reflect.DeepEqual(oldC.Spec.HCloudNetwork, r.Spec.HCloudNetwork) {
+	if !reflect.DeepEqual(oldC.Spec.HCloudNetwork.Enabled, r.Spec.HCloudNetwork.Enabled) {
 		allErrs = append(allErrs,
-			field.Invalid(field.NewPath("spec", "hcloudNetwork"), r.Spec.HCloudNetwork, "field is immutable"),
+			field.Invalid(field.NewPath("spec", "hcloudNetwork", "enabled"), r.Spec.HCloudNetwork.Enabled, "field is immutable"),
 		)
 	}
 
+	if oldC.Spec.HCloudNetwork.Enabled {
```
janiskemper, line 259: if this was not enabled, then we are not allowed to update anything, right? This should be added to these checks

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -161,13 +250,46 @@ func (r *HetznerCluster) ValidateUpdate(old runtime.Object) (admission.Warnings,
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected an HetznerCluster but got a %T", old))
 	}
 
-	// Network settings are immutable
-	if !reflect.DeepEqual(oldC.Spec.HCloudNetwork, r.Spec.HCloudNetwork) {
+	if !reflect.DeepEqual(oldC.Spec.HCloudNetwork.Enabled, r.Spec.HCloudNetwork.Enabled) {
```
janiskemper, line 253: we only check equality of booleans here.. I think we don't need DeepEqual ;)

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -161,13 +250,46 @@ func (r *HetznerCluster) ValidateUpdate(old runtime.Object) (admission.Warnings,
...
 		allErrs = append(allErrs,
-			field.Invalid(field.NewPath("spec", "hcloudNetwork"), r.Spec.HCloudNetwork, "field is immutable"),
+			field.Invalid(field.NewPath("spec", "hcloudNetwork", "enabled"), r.Spec.HCloudNetwork.Enabled, "field is immutable"),
 		)
 	}
 
+	if oldC.Spec.HCloudNetwork.Enabled {
+		// Only allow updating the network ID when it was not set previously. This makes it possible to e.g. adopt the
+		// network that was created initially by CAPH.
+		if oldC.Spec.HCloudNetwork.ID != nil && !reflect.DeepEqual(oldC.Spec.HCloudNetwork.ID, r.Spec.HCloudNetwork.ID) {
```
janiskemper, line 262: again, no need for deepequal I think

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -161,13 +250,46 @@ func (r *HetznerCluster) ValidateUpdate(old runtime.Object) (admission.Warnings,
...
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec", "hcloudNetwork", "id"), r.Spec.HCloudNetwork.ID, "field is immutable"),
+			)
+		}
+
+		if r.Spec.HCloudNetwork.ID != nil {
+			if errs := areCIDRsAndNetworkZoneEmpty(r.Spec.HCloudNetwork); errs != nil {
```
janiskemper, line 269: I think you can combine this check with the ones below. Here you check that it is (again) empty, and below you check that it is immutable. 
Essentially it is in both cases a check for immutability right?

api/v1beta1/types.go
```diff
@@ -217,32 +217,39 @@ type LoadBalancerTarget struct {
...
 	// +optional
-	SubnetCIDRBlock string `json:"subnetCidrBlock,omitempty"`
+	SubnetCIDRBlock *string `json:"subnetCidrBlock,omitempty"`
 
 	// NetworkZone specifies the HCloud network zone of the private network.
-	// The zones must be one of eu-central, us-east, or us-west. The default is eu-central.
-	// +kubebuilder:validation:Enum=eu-central;us-east;us-west;ap-southeast
-	// +kubebuilder:default=eu-central
+	// The zones must be one of eu-central, us-east, or us-west.
+	// Defaults to "eu-central".
```
janiskemper, line 243: can you add this "webhook defaults" to all of the comments?

pkg/services/hcloud/network/network.go
```diff
@@ -106,14 +118,18 @@ func (s *Service) createNetwork(ctx context.Context) (*hcloud.Network, error) {
 func (s *Service) createOpts() (hcloud.NetworkCreateOpts, error) {
 	spec := s.scope.HetznerCluster.Spec.HCloudNetwork
 
-	_, network, err := net.ParseCIDR(spec.CIDRBlock)
+	if spec.CIDRBlock == nil || spec.SubnetCIDRBlock == nil || spec.NetworkZone == nil {
+		return hcloud.NetworkCreateOpts{}, errors.New("nil CIDRs or NetworkZone given")
```
janiskemper, line 122: you can use "fmt.Errorf" for this as well, we use it everywhere

janiskemper (commented):
thanks @johannesfrey ! I went another time over the details and found a few things.

api/v1beta1/hetznercluster_webhook.go
johannesfrey: I agree that we should probably append an error when NetworkZone is nil. But still I'm struggling to see how to really circumvent the nested statements because the method's idea seems to be to aggregate all errors that might occur instead of returning early. An early return would then potentially miss other validation errors that might happen afterwards. But probably I'm just missing a detail.

api/v1beta1/hetznercluster_webhook.go
johannesfrey: Aligns with the comment above. Do you mean return with an error here? Because in this method we only aggregate and return at the bottom?!

api/v1beta1/hetznercluster_webhook.go
johannesfrey: As ID is a pointer, the check might be simpler using DeepEqual as it also handles the case where a user might remove the ID, which would result in nil.

pkg/services/hcloud/network/network.go
janiskemper: @johannesfrey  didn't see your response here sorry. "record.Warnf()" would be appropriate here. that's actually something we should still add!

api/v1beta1/hetznercluster_webhook.go
janiskemper: I wanted to avoid the large code block for "NetworkZone != nil". However, I think you are right that this is not really possible, as we cannot just return at that point. 

Therefore, I would say let's add a handling of "NetworkZone == nil" and otherwise don't change anything. Does that make sense to you?

api/v1beta1/hetznercluster_webhook.go
janiskemper: yeah I mean with "returning errors" to "add them". Sorry for being not precise

api/v1beta1/hetznercluster_webhook.go
johannesfrey: Sure! And yeah, it's not the most pretty thing 😅.
Regarding the "NetworkZone == nil". So do you mean to only handle the nil case and delete the validation of the network zones or handle both cases? In that vein, should I also handle the nil case for cidrBlock and subnetCIDRBlock as well?

api/v1beta1/hetznercluster_webhook.go
johannesfrey: No worries. Just wanted to make sure I understood you correctly 🙂

api/v1beta1/hetznercluster_webhook.go
janiskemper: you did that already right? Do you think we miss any important case right now, or is this done?

janiskemper:
thanks @johannesfrey! Do you think anything is missing right now? If not, I'd propose the following path:
- I will do a final review
- You squash and rebase
- Someone from our team will again do a functionality test (which I haven't done at all) that the actual behavior is as expected

api/v1beta1/hetznercluster_webhook.go
johannesfrey: Yep. Did this as well.

johannesfrey:
> thanks @johannesfrey! Do you think anything is missing right now? If not, I'd propose the following path:
> 
>     * I will do a final review
> 
>     * You squash and rebase
> 
>     * Someone from our team will again do a functionality test (which I haven't done at all) that the actual behavior is as expected

That sounds awesome. Thx!
One thing that would be great to take a look at. Is my (probably too naive) tests in `controllers/hetznercluster_controller_test.go`. There seems to be a data race, especially in the first test. Most of the time it succeeds but there are cases where the reconciler errors that the requested network cannot be found while passing in the id. I'm struggling to see why this happens because the line of execution should look like this:
`create network with fake client -> check that there is no error -> add the id to the hetznercluster spec -> trigger a create of the hetznercluster`
The reconciler should then use its internal client (which should be the shared fake one from before) to find the network. But it cannot find it, so there must be something in the line that deleted it or there is some race when using the fake client and probably the usage of the mutexes in there?! I tried some variations of changing the locks in there, but to no avail. So wasn't able to really deflake the test. So, would be really cool if you could take another look there. And if the test makes more harm than that it's helping, we could also think of removing/chaning it. WDYT?

janiskemper:
mmh that's an important observation, thanks @johannesfrey. We will have a look. I'm not able to see anything in the code right now.

bitnik:
Hello, what is the plan here? Is this PR still considered?

batistein:
we just moved this PR into testing.

tcldr:
Looking forward to trying this out. I'm planning to use an existing private network with a NAT Gateway/Bastion Host.

One big advantage of a private network is that it doesn't incur overage fees when you take the 10Gbps option.

batistein:
@tcldr this implementation only covers hcloud not vswitch. Also internal routing even via public IPs is not incurring any costs.

tcldr:
@batistein good to know, thanks.

Is there something particular blocking vswitch enabled networks and subnets?

batistein:
They are basically unstable, and we removed, therefore, for all our customers, the support of private networks 3 years ago. Since then, the instability on hetzner side are not resolved that's why we never invested time to support them via caph. As Syself we switched to a zero trust architecture which aligns also more with our future plans. See: https://syself.com/docs/hetzner/apalla/platform/zero-trust

batistein:
Yes, of course we are open to PRs, these need to be E2E tested and the logic needs to be separated so it doesn't affect the current code. 
But believe me, this is wasted time. I spoke to a customer yesterday who is migrating to Syself with over hundreds of servers, and they have network outages on a daily basis because they are using vswitch in their self-managed Kubernetes environment... I don't think this is just a coincidence, without vswitch and a private network there is no noticeable problem at all. Also, the server limitations of the Hetzner network make it unsuitable for enterprise use. I've seen the team do the craziest things with connecting multiple networks to get around the limitations, ending up spending months on a topology only to realise it doesn't work. 
If you're worried about privacy, use mTLS and strict network policies.

bitnik:
We were also waiting for this PR, but meanwhile we tried out hetzner private network and experienced similar issues as @batistein told about. So we decided not to use it neither and go zero-trust way.

lkt82:
Any news on this request?

johannesfrey:
@batistein Thx for all the context around your past experiences with private networks.

At the time I initially created this PR I was not aware that you have been dropping support for those already 3 years ago (at least for the customers you have been interacting with). So I'm a bit unsure to "burden" you with even more private network functionality. So if the recommended way is zero-trust, I would be totally fine do close this PR. So I guess we would "just" need a decision how to proceed here (with the option to "close and forget" being totally fine form my side 😉). And I guess this would also relieve all the others waiting for this to be merged.

Also @janiskemper thx so much for your effort reviewing all of this.

janiskemper:
@johannesfrey IMO everyone should decide that on their own. We don't use Hetzner's private networks and have reasons to not do it. However, others might come to different conclusions. I'd very much like to get this merged. We are very busy right now internally though - that's why the testing didn't happen yet. I recognize the interest into this and put it up on our agenda. It will still take some time though, I'm afraid. However, that's a good thing because in this way we will be able to ensure the quality of CAPH, even if it is a bit slower.

PR #1502, opened 2024-11-04
:seedling: Fix CI of main branch.

guettli:

Fix CI of main branch.


guettli:
CI should be fixed now.

PR #1521, opened 2025-01-03
✨ Backwards compatible support for containerd 2

emolitor:
**What this PR does / why we need it**: Backwards compatible support for containerd 2

**Fixes**: https://github.com/syself/cluster-api-provider-hetzner/issues/1522

**Special notes for your reviewer**: I used a variation of this over the holidays while testing CAPH. The support for the 'batteries included' containerd release was deprecated a while ago and removed in containerd 2.0. This change updates the Ubuntu quickstart templates and packer templates to use the supported packaging for containerd and separately installs runc. This was tested with containerd 1.7.24 and containerd 2.0.1. 

One subtle change to make note of is moving the hardening of /etc/cni permissions into kubernetes.sh for the packer templates. Previously the containerd 'batteries included' package included kubernetes-cni as well as being installed as a dependency of kubelet (from the Kubernetes apt repo). Now that kubernetes-cni is solely being installed as part of the kubernetes apt packaging I've moved the permissions change out of cri.sh and into kubernetes.sh.

Lastly I simplified setting ARCH to use dpkg instead of uname as, at least for Ubuntu and Debian, it's a much cleaner approach.

Feedback is greatly appreciated. One question I had is what is the purpose of the structed comment on the existing CONTAINERD version line (see following). It seems like the RUNC version might want something similar?

`export CONTAINERD=1.7.24 # update: datasource=github-tags depName=containerd/containerd extractVersion=^v(?<version>.*)$ versioning=semver`

### DIFF
--- a/templates/node-image/1.28.9-ubuntu-22-04-containerd/scripts/cri.sh
+++ b/templates/node-image/1.28.9-ubuntu-22-04-containerd/scripts/cri.sh
@@ -47,25 +47,53 @@ kernel.panic=10
 kernel.panic_on_oops=1
 EOF
 
+# Create containerd systemd unit
+cat >/etc/systemd/system/containerd.service <<'EOF'
+[Unit]
+Description=containerd container runtime
+Documentation=https://containerd.io
+After=network.target local-fs.target dbus.service
+
+[Service]
+ExecStartPre=-/sbin/modprobe overlay
+ExecStart=/usr/local/bin/containerd
+Type=notify
+Delegate=yes
+KillMode=process
+Restart=always
+RestartSec=5
+LimitNPROC=infinity
+LimitCORE=infinity
+LimitNOFILE=infinity
+TasksMax=infinity
+OOMScoreAdjust=-999
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
 # Apply sysctl params without reboot
 sysctl --system
 
+ARCH="$(dpkg --print-architecture)"
 CONTAINERD=1.7.16 # https://github.com/containerd/containerd/releases
+RUNC=1.2.3 # https://github.com/opencontainers/runc/releases
+
+# Install runc
+wget https://github.com/opencontainers/runc/releases/download/v$RUNC/runc.$ARCH
+wget https://github.com/opencontainers/runc/releases/download/v$RUNC/runc.sha256sum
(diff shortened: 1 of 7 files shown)


batistein:
This comment will be tracked and updated by renovatebot. This is for demonstration purposes only, to show how to use automation while keeping components up to date. We haven't matched all the parts to this, so we can either keep it or drop it. Best case scenario, someone will align it with all the templates to make it easier to keep the components up to date.

templates/node-image/1.28.9-ubuntu-22-04-containerd/scripts/cri.sh
```diff
@@ -47,25 +47,53 @@ kernel.panic=10
...
+tar -zxf containerd-$CONTAINERD-linux-$ARCH.tar.gz -C /usr/local
 
 # Cleanup
-rm -f cri-containerd-cni-${CONTAINERD}-linux-${PACKER_ARCH}.tar.gz cri-containerd-cni-${CONTAINERD}-linux-${PACKER_ARCH}.tar.gz.sha256sum
-
-mkdir -p /etc/containerd
-
-# Sets permission accordingly to CIS Benchmark
-chmod -R 644 /etc/cni
-chown -R root:root /etc/cni
```
guettli, line 68: `chown -R root:root /etc/cni` gets lost in the PR. Do you think this should get removed?

lucasrattz: The CNI plugins are not included anymore in the containerd release artifact. It needs to be installed separately from https://github.com/containernetworking/plugins

@emolitor do you think this is not needed?

emolitor: I added the chown back in. I think it is redundant as the files are already owned by root but this is a separate change and should not be part of this PR.

emolitor (commented):
Updated per comments.

guettli (approved):
lgtm

guettli:
@emolitor just out of curiosity: Why do you want to use containerd 2.x? Which features do you need, and why?

emolitor:
> @emolitor just out of curiosity: Why do you want to use containerd 2.x? Which features do you need, and why?

Was playing around with the Transfer Service as a spike to prototype something and it is only stable in v2.

PR #1542, opened 2025-03-11
:seedling: Update Containerd, Runc, Kubernetes Version, Packages. Remove unsupported parts.

guettli:
This PR updates dependencies and removes outdated parts that we did not maintain in the past.

Overall, we believe that it makes sense to remove parts that are no longer being maintained.

If this PR is causing you trouble, and you want something reverted, please contact us: create an issue and elaborate on your use case.

---

KUBERNETES_VERSION v1.29.4 --> v1.31.6

Ubuntu 22.04 --> 24.04

CONTAINERD 1.7.25 --> 1.7.26

RUNC 1.2.3 --> 1.2.5

cluster-api v1.8.4 --> v1.8.10

---

Fix warning:

> Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. 

A unit-test failed, and  this was the reason for that.

---

Make e2e work locally again. Fix docs for running e2e tests. Make templating in e2e tests more robust.

---

use ssh-key shared-2024-07-08 in e2e tests.

---

make hack/ci-e2e-get-token.sh more robust: provide an error message if you use an outdated TPS_TOKEN

---

Not supported parts got removed: fedora, packer  and csr-off. We never use or test these templates. The docs got updated, so that users can help themselves.

---

Hide the default `privateNetwork.enabled=false`. If you do not know the option (like me until yesterday), you might have the gut feeling that this setting is insecure. But that is not the case. It lets the ccm set up routes in hcloud, so that the CNI does not need to do the routing. See https://github.com/hetznercloud/hcloud-cloud-controller-manager#networks-support

---

Show BareMetalHost Ready Condition directly in the output continuously. No need to check the artifacts for that.


docs/caph/02-topics/03-node-image.md
```diff
@@ -18,26 +18,10 @@ Node-image is necessary for using CAPH in production and as a user, there can be
...
 There are several ways to achieve this. In the quick-start guide, we use `pre-kubeadm` commands in the `KubeadmControlPlane` and `KubeadmConfigTemplate` objects. These are propagated from the bootstrap-provider-kubeadm and the control-plane-provider-kubeadm to the node as cloud-init commands. This way is usable universally also in other infrastructure providers.
 
-For Hcloud, there is an alternative way of doing this using Packer. It creates a snapshot to boot from. This makes it easier to version the images, and creating new nodes using this image is faster. The same is possible for Hetzner BareMetal, as we could use installimage and a prepared tarball, which then gets installed as the OS for your nodes.
+For Hcloud, there is an alternative way of doing this my using Hcloud Snapshots. Snapshots are images you can boot from. This makes it easier to version the images, and creating new nodes using this image is faster. The same is possible for Hetzner BareMetal, as we could use installimage and a prepared tarball, which then gets installed as the OS for your nodes.
```
janiskemper, line 21: my -> by

janiskemper (commented):
found a typo. I believe you should explain what exactly you remove in this PR and why

docs/caph/02-topics/03-node-image.md
guettli: typo is fixed. I will update the PR description.

guettli:
All e2e are fine: https://github.com/syself/cluster-api-provider-hetzner/actions/runs/13863802258/job/38798330043?pr=1542

Git hash: 45b001d09056fa0f1959d7f6c4d6d35b8bf21176

batistein (approved):

PR #1548, opened 2025-03-13
:seedling: dummy PR to test CI of baremetal e2e tests.

guettli:
dummy PR to test CI of baremetal e2e tests.


guettli:
CI on main should be fine now.

PR #1551, opened 2025-03-15
:seedling: Fix flaky envTest and add Debug Output

guettli:
Fix flaky unit test for "LoadBalancer should take over ..." and add debug output which will help us in the future to debug tests.

There are many of our objects which get reconciled although the envTests which created these objects are already finished.

Additionally, the timeout of that test  "should take over an existing load balancer with correct name" was increased. 

---

Secrets got created in a Table-Test. These secrets were not deleted up to now. This created many continuous reconcile of resources which are not relevant anymore (because they belong to a test which already was finished).

Related:

[Configuring EnvTest - The Kubebuilder Book: Namespace Deletion](https://book.kubebuilder.io/reference/envtest#namespace-usage-limitation)

---

Additionally, the PR adds continuous output to the HCloud e2e test (like we already have it for bm):

```
  STEP: --------------------------------------------------- @ 03/15/25 13:03:23.773
  STEP: HCloudMachine: ci13872902245-ao7cav-control-plane-v94sm 0xc0013898d0 @ 03/15/25 13:03:53.779
  STEP:   ProvisioningState: starting @ 03/15/25 13:03:53.779
  STEP:   Ready Condition: False ServerStarting server is starting @ 03/15/25 13:03:53.779
  STEP: HCloudMachine: ci13872902245-ao7cav-control-plane-xfp55 0xc0013899e0 @ 03/15/25 13:03:53.779
  STEP:   ProvisioningState: running @ 03/15/25 13:03:53.779
  STEP:   Ready Condition: True   @ 03/15/25 13:03:53.779
  STEP: HCloudMachine: ci13872902245-ao7cav-control-plane-xq9pp 0xc001389af0 @ 03/15/25 13:03:53.779
  STEP:   ProvisioningState: running @ 03/15/25 13:03:53.779
  STEP:   Ready Condition: True   @ 03/15/25 13:03:53.78
  STEP: HCloudMachine: ci13872902245-ao7cav-md-0-hxxh6-jkpds 0xc001389c40 @ 03/15/25 13:03:53.78
  STEP:   ProvisioningState: running @ 03/15/25 13:03:53.78
  STEP:   Ready Condition: True   @ 03/15/25 13:03:53.78
  STEP: --------------------------------------------------- @ 03/15/25 13:03:53.78
```

---

upload artifacts (even on failure), dump controller runtime metrics to `.reports`.

Example: controller_suite_test-metrics.txt in [Test Code Artifact](https://github.com/syself/cluster-api-provider-hetzner/actions/runs/13882125398/artifacts/2759725809)

---

Github Actions which upload Artifacts use `if: ${{ !cancelled() }}` now. This makes them run if the test was ok, or it failed. But not when the test was canceled. This is the official recommendation of Github.

---

The slowest tests get written to .reports now. Unfortunately all Ginkgo controller tests are measured as one big test. But we can fix that later.


batistein (approved):

controllers/hetznercluster_controller.go
```diff
@@ -409,6 +409,10 @@ func getAndValidateHCloudToken(ctx context.Context, namespace string, hetznerClu
 		return "", nil, err
 	}
 
+	if !hetznerSecret.DeletionTimestamp.IsZero() {
```
janiskemper, line 412: I suppose this could be problematic  o?

janiskemper (commented):
This seems as if it can create problems in case the secret has a deletion timestamps but is blocked from deletion... the controller would stop working.

I didn't check the details of this continuous output you created so cannot judge

controllers/hetznercluster_controller.go
guettli: I removed that again. Afaik it is not needed any more. Now the root cause of the flaky test was fixed: Some tests created secrets without deleting them. This made the controller reconcile again and again.

guettli: @janiskemper grrr after removing that line, the test failed again:
 
 > [FAIL] Hetzner ClusterReconciler cluster tests load balancer [It] should take over an existing load balancer with correct name

main.go
```diff
@@ -110,6 +111,10 @@ func main() {
 
 	ctrl.SetLogger(utils.GetDefaultLogger(logLevel))
 
+	bInfo, _ := debug.ReadBuildInfo()
+	setupLog.Info(bInfo.Main.Path, "version", caphversion.Get().String(),
```
janiskemper, line 115: where is the advantage of this line compared to the "setupLog.Info("starting manager", "version", caphversion.Get().String())" that we already have?

janiskemper (commented):
@batistein has to review this for the CI part

main.go
guettli: The string of caphversion is empty. You can see that in the testing-cluster:

```
❯ k get -n mgt-system deployments.apps caph-controller-manager -o yaml | grep image:
        image: ghcr.io/syself/caph:v1.0.0
```

```
❯ k logs -n mgt-system deployments/caph-controller-manager   | grep 'starting manager'
Found 4 pods, using pod/caph-controller-manager-5cbd64b89c-4cj8t
{"level":"INFO","time":"2025-03-16T03:56:44.057Z","logger":"setup","file":"./main.go:237","message":"starting manager","version":""}

```

With Go1.24 the string will be very useful, because it contains the git-hash if the build is not clean.

batistein (approved):

main.go
janiskemper: I don't get why you added a second version of the same code @guettli .. Sorry. Can you explain it for dumb people? 
I would in general expect that if you write code that replaces old code, that you would also remove this old code and don't keep it...

janiskemper (approved):

main.go
guettli: I removed the part you mentioned. We will do that via #1553

controllers/hetznercluster_controller.go
guettli: solved now.

PR #1554, opened 2025-03-17
DO NOT MERGE. Check e2e caph image before PR1527.

guettli:
What image gets run in e2e tests?

PR which might introduced a change, so that CAPH_LATEST_VERSION gets used.

https://github.com/syself/cluster-api-provider-hetzner/pull/1527


guettli:
Will be fixed in #1561

PR #1585, opened 2025-04-17
:seedling: fixing hanging deprovisioning

guettli:
fixing hanging deprovisioning

Additionally:

Avoid

> Warning: Reconciler returned both a non-zero result and a non-nil error.
> The result will always be ignored if the error is non-nil and the non-nil error
> causes reqeueuing with exponential backoff. For more details,
> see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

### DIFF
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -25,15 +25,15 @@ jobs:
             exit 1
           fi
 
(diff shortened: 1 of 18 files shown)


guettli:
already merged in other PR.

PR #1590, opened 2025-04-23
:seedling: Increase requeueAfter No Available Host. And trigger Reconcile of hbmm when hbmh is free.

### LINKED ISSUE #1588: Improve handling of "no available host"
/kind feature

Currently, the handling is that we reconcile every 30 seconds until we find a host. However, this is usually unnecessary. We can increase the timeout to e.g. 5 minutes, and in this way reduce the load of the controller, and keep most of the functionality. 

We can additionally improve the reactivity by reacting on "HetznerBareMetalHost create" events in the hetznerbaremetalmachine-controller. In this way, the usual case that a new host is added and then used by one of the machines, will be leveraged with perfect reactivity.

guettli:
Fixes #1588

### DIFF
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -345,8 +345,7 @@ func BareMetalHostToBareMetalMachines(c client.Client, log logr.Logger) handler.
 			return nil
 		}
 
-		// If this host has a consumerRef (hbmm), then it is not free,
-		// and we don't need to reconcile it.
+		// If this host has a consumerRef (hbmm), then reconcile the corresponding hbmm.
 		if host.Spec.ConsumerRef != nil {
 			return []reconcile.Request{
 				{
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -368,6 +368,7 @@ func (r *HetznerBareMetalMachineReconciler) BareMetalHostToBareMetalMachines(log
 			return []reconcile.Request{}
 		}
 
+		// We have a free host. Trigger a matching HetznerBareMetalMachine to be reconciled.
 		hbmmList := infrav1.HetznerBareMetalMachineList{}
 		err := r.Client.List(context.TODO(), &hbmmList, client.InNamespace(host.Namespace))
 		if err != nil {
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -345,6 +345,8 @@ func BareMetalHostToBareMetalMachines(c client.Client, log logr.Logger) handler.
 			return nil
 		}
 
+		// If this host has a consumerRef (hbmm), then it is not free,
+		// and we don't need to reconcile it.
 		if host.Spec.ConsumerRef != nil {
 			return []reconcile.Request{
 				{
@@ -378,8 +380,7 @@ func BareMetalHostToBareMetalMachines(c client.Client, log logr.Logger) handler.
 				continue
 			}
 
-			hosts := make([]infrav1.HetznerBareMetalHost, 1)
-			hosts[0] = *host
+			hosts := []infrav1.HetznerBareMetalHost{*host}
 			chosenHost, _, err := baremetal.ChooseHost(hbmm, hosts)
 			if err != nil {
 				log.Error(err, "failed to choose host for HetznerBareMetalMachine")
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -345,9 +345,7 @@ func (r *HetznerBareMetalMachineReconciler) BareMetalHostToBareMetalMachines(log
 			return nil
 		}
 
-		if host.Spec.ConsumerRef != nil &&
-			host.Spec.ConsumerRef.Kind == "HetznerBareMetalMachine" &&
-			host.Spec.ConsumerRef.GroupVersionKind().Group == infrav1.GroupVersion.Group {
+		if host.Spec.ConsumerRef != nil {
 			return []reconcile.Request{
 				{
 					NamespacedName: types.NamespacedName{
@@ -356,12 +354,6 @@ func (r *HetznerBareMetalMachineReconciler) BareMetalHostToBareMetalMachines(log
 					},
 				},
 			}
-		}
-
-		if host.Spec.ConsumerRef != nil {
-			log.Info("unexpected hmbh.Spec.ConsumerRef: %q %q", host.Spec.ConsumerRef.Kind,
-				host.Spec.ConsumerRef.GroupVersionKind().Group)
-			return []reconcile.Request{}
 		}
 
 		if host.Spec.Status.ErrorType != "" {
@@ -375,23 +367,31 @@ func (r *HetznerBareMetalMachineReconciler) BareMetalHostToBareMetalMachines(log
 			log.Error(err, "failed to list HetznerBareMetalMachines")
 			return []reconcile.Request{}
 		}
+
+		// Search for a hbmm which would like to use this host.
 		for i := range hbmmList.Items {
 			hbmm := &hbmmList.Items[i]
+
+			// Skip if the hbmm is already in use.
 			if hbmm.HasHostAnnotation() {
 				continue
 			}
+
 			hosts := make([]infrav1.HetznerBareMetalHost, 1)
 			hosts[0] = *host
-			host, _, err := baremetal.ChooseHost(hbmm, hosts)
+			chosenHost, _, err := baremetal.ChooseHost(hbmm, hosts)
 			if err != nil {
 				log.Error(err, "failed to choose host for HetznerBareMetalMachine")
 				continue
 			}
-			if host == nil {
+
+			// this hbmm does not match the host
+			if chosenHost == nil {
 				continue
 			}
 
 			// We found a matching hbmm for the free host. Trigger Reconcile for the hbmm.
+			// Exit the loop, no need to look for more.
 			return []reconcile.Request{
 				{
 					NamespacedName: types.NamespacedName{
--- a/controllers/hetznerbaremetalmachine_controller.go
+++ b/controllers/hetznerbaremetalmachine_controller.go
@@ -337,22 +337,67 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
 // BareMetalHost and that BareMetalHost references a BareMetalMachine.
 func (r *HetznerBareMetalMachineReconciler) BareMetalHostToBareMetalMachines(log logr.Logger) handler.MapFunc {
 	return func(_ context.Context, obj client.Object) []reconcile.Request {
-		if host, ok := obj.(*infrav1.HetznerBareMetalHost); ok {
-			if host.Spec.ConsumerRef != nil &&
-				host.Spec.ConsumerRef.Kind == "HetznerBareMetalMachine" &&
-				host.Spec.ConsumerRef.GroupVersionKind().Group == infrav1.GroupVersion.Group {
-				return []reconcile.Request{
-					{
-						NamespacedName: types.NamespacedName{
-							Name:      host.Spec.ConsumerRef.Name,
-							Namespace: host.Spec.ConsumerRef.Namespace,
-						},
-					},
-				}
-			}
-		} else {
+		host, ok := obj.(*infrav1.HetznerBareMetalHost)
+		if !ok {
 			log.Error(fmt.Errorf("expected a BareMetalHost but got a %T", obj),
 				"failed to get BareMetalMachine for BareMetalHost")
+			return nil
+		}
+
+		if host.Spec.ConsumerRef != nil &&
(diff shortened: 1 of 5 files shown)


controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -338,22 +338,68 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
...
+				{
+					NamespacedName: types.NamespacedName{
+						Name:      host.Spec.ConsumerRef.Name,
+						Namespace: host.Spec.ConsumerRef.Namespace,
+					},
+				},
+			}
+		}
+
+		if host.Spec.ConsumerRef != nil {
```
janiskemper, line 361: is this really necessary? IMO not

controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -338,22 +338,68 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
...
+		}
+
+		// We have a free host. Trigger a matching HetznerBareMetalMachine to be reconciled.
+		hbmmList := infrav1.HetznerBareMetalMachineList{}
+		err := r.Client.List(context.TODO(), &hbmmList, client.InNamespace(host.Namespace))
+		if err != nil {
+			log.Error(err, "failed to list HetznerBareMetalMachines")
+			return []reconcile.Request{}
+		}
+		for i := range hbmmList.Items {
```
janiskemper, line 378: can you maybe explain in comments what you do in this for loop?

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -61,6 +61,8 @@ const (
 	// requeueAfter gives the duration of time until the next reconciliation should be performed.
 	requeueAfter = time.Second * 30
 
+	requeueAfterNoAvailableHost = time.Minute * 3
```
janiskemper, line 64: IMO this could be increased further, especially since we will be able to catch the events from the hosts

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -825,4 +825,154 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 			})
 		})
 	})
+
+	Context("Test EnqueueRequestsFromMapFunc BareMetalHostToBareMetalMachines", func() {
```
janiskemper, line 829: Why do you test this with envtest and with relying on the events? You can easily test this in the same way and via a stable way through unit tests that use the Kubernetes API.. Just call BareMetalHostToBareMetalMachines directly!

janiskemper (changes requested):

controllers/hetznerbaremetalmachine_controller_test.go
guettli: I could write a test for BareMetalHostToBareMetalMachines, but this would not show that this function would resolve the issue.

Test Driven Development means: First write a test which shows that the current code is broken or is missing a feature. Then write the code to make the test pass. I do not do that always. But in this case, I would really want to see that my fix is fixing something.

pkg/services/baremetal/baremetal/baremetal.go
guettli: set to 7 minutes.

controllers/hetznerbaremetalmachine_controller_test.go
janiskemper: @guettli I'm totally up for test-driven development. But you can do that as well with a normal unit test that would show you in a more consistent way that it works.

guettli: How to write a unit-test which shows that the hbmm waits "forever" and does not detect when a hbmh gets free?

guettli: @janiskemper the test was re-written to not use sleep(). Additionally, a unit-test was added. Please review again.

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -977,3 +985,87 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
...
+			},
+		},
+	}
+	c := fakeclient.NewClientBuilder().WithScheme(scheme).WithObjects(
+		host, hbmm, hbmmOtherNS, hbmmWithoutLabel, hbmmWithHostAnnotation).Build()
+
+	// host does not have a label.
+	f := BareMetalHostToBareMetalMachines(c, logr.Logger{})
+	result := f(ctx, host)
+	require.Equal(t, "[]", fmt.Sprintf("%v", result))
```
janiskemper, line 1061: why do you use this method to check for equality? Why not comparing the Go structs here and below?

controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -336,25 +336,69 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
...
+func BareMetalHostToBareMetalMachines(c client.Client, log logr.Logger) handler.MapFunc {
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		host, ok := obj.(*infrav1.HetznerBareMetalHost)
+		if !ok {
 			log.Error(fmt.Errorf("expected a BareMetalHost but got a %T", obj),
 				"failed to get BareMetalMachine for BareMetalHost")
+			return nil
+		}
+
+		if host.Spec.ConsumerRef != nil {
```
janiskemper, line 348: add a comment that you found the owner of the host and you reconcile the owner

controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -336,25 +336,69 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
...
+		var found []reconcile.Request
+		for i := range hbmmList.Items {
+			hbmm := &hbmmList.Items[i]
+
+			// Skip if the hbmm is already in use.
+			if hbmm.HasHostAnnotation() {
+				continue
+			}
+
+			hosts := make([]infrav1.HetznerBareMetalHost, 1)
```
janiskemper, line 381: you can define the slice in one row instead of two.. If you don't want this, then ignore it. But doing `[*host]` should be enough AFAIK

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -825,4 +833,239 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 			})
 		})
 	})
+
+	Context("Test EnqueueRequestsFromMapFunc BareMetalHostToBareMetalMachines", func() {
+		It("should enqueue the second HetznerBareMetalMachine, when the first gets deleted", func() {
+			// We test this part in BareMetalHostToBareMetalMachines:
+			// We have a free host. Trigger a matching HetznerBareMetalMachine to be reconciled.
+			// To ensure this test works, uncomment the lines below
```
janiskemper, line 841: what do you mean by this uncommenting?

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -825,4 +833,239 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
...
+			// "We have a free host" in the function. Then this test should fail.
```
janiskemper, line 842: I don't understand this line either

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -825,4 +833,239 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
...
+				}
+
+				if host.Spec.ConsumerRef.Name == bmMachineName {
+					pickedMachine = bmMachine
+					waitingMachine = bmMachine2
+				} else {
+					pickedMachine = bmMachine2
+					waitingMachine = bmMachine
+				}
+				osSSHClient.On("GetHostName").Return(sshclient.Output{
```
janiskemper, line 962: in case the machine is already provisioning and this mock is needed, then it comes a bit late right? This sounds pretty flaky tbh.

controllers/hetznerbaremetalmachine_controller_test.go
guettli: This way, nil and an empty list are considered equal. And if there is an item in the list, you see immediately the result. If I compared the length, then you only would see `0 != 1`. I can change that, if this is importanet for you.

controllers/hetznerbaremetalmachine_controller.go
guettli: done

controllers/hetznerbaremetalmachine_controller.go
guettli: yes, good point. One line is better than two.

controllers/hetznerbaremetalmachine_controller_test.go
guettli: I remove the lower part from the comment. It does not matter.

controllers/hetznerbaremetalmachine_controller_test.go
guettli: this is line was removed.

controllers/hetznerbaremetalmachine_controller_test.go
guettli: This is needed. I added this comment:

```go
				// We need to change the mock on-the-fly. There are two
				// machines, and we need to adapt the mock to the one that is
				// picked.
```

janiskemper: I am aware of the fact that you need it. I'm just saying it sounds very flaky because AFAIK if the machine reached that point in the state machine / provisioning process before that mock is defined, then the test would fail. Isn't this a possibility according to you?

guettli: I adapted the test, so that it uses WaitUntil() to make the mock wait. See ddd7e543

controllers/hetznerbaremetalmachine_controller_test.go
guettli: The code compares go types now: 7c188fde

controllers/hetznerbaremetalmachine_controller.go
```diff
@@ -336,25 +336,70 @@ func (r *HetznerBareMetalMachineReconciler) ClusterToBareMetalMachines(ctx conte
...
+	return func(ctx context.Context, obj client.Object) []reconcile.Request {
+		host, ok := obj.(*infrav1.HetznerBareMetalHost)
+		if !ok {
 			log.Error(fmt.Errorf("expected a BareMetalHost but got a %T", obj),
 				"failed to get BareMetalMachine for BareMetalHost")
+			return nil
+		}
+
+		// If this host has a consumerRef (hbmm), then it is not free,
+		// and we don't need to reconcile it.
```
janiskemper, line 349: I think you do add the hbmm to reconcile here... I think this comment doesn't make sense

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -825,4 +834,258 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
...
+					// machines, and we need to adapt the mock to the one that is
+					// picked.
+					osSSHClient.On("GetHostName").Return(sshclient.Output{
+						StdOut: infrav1.BareMetalHostNamePrefix + pickedMachine.Name,
+						StdErr: "",
+						Err:    nil,
+					})
+					// Close the channel, so that mock "GetHostName" does no longer block.
+					close(w)
+					w = nil
```
janiskemper, line 978: why is this necessary?

controllers/hetznerbaremetalmachine_controller.go
guettli: yes, you are right. I fixed the comment.

controllers/hetznerbaremetalmachine_controller_test.go
guettli: What do you mean? We first need to wait until the host has a consumerRef.

The mock of GetHostName blocks until a machine was picked. When it was picked we set the return value, and remove the blocking of the GetHostName mock. Unblocking gets done with the channel `w`.

Does this answer your question?

janiskemper: I mean only "w= nil"

guettli: the channel `w` needs to be closed. Closing a closed channel panics. So I set it to nil, so that `w` gets closed only once.

Does this answer your question?

janiskemper: can you maybe add this as a comment to the code? This is not understandable at all for someone looking at it and actually quite hacky. But of course necessary in your "Eventually".

guettli: @janiskemper I added that comment:

```go 
                    // We updated the mock, and closed the channel. Set channel to nil,
                    // so that we don't mock and close again.
                    // Closing a closed channel would panic.
```

ok?

janiskemper: yes thanks!

janiskemper (approved):

PR #1598, opened 2025-05-07
:seedling: Before using the ssh Rescue Client, ensure the rescue system is active. Log reboot events

guettli:
Before using the ssh Rescue Client, ensure the rescue system is active.

This uses ssh not an API. So rate-limit is not an issue.

It is better to double-check that.

Additionally, Events created when rebooting machines will be written to the logs, so that debugging failing e2e is easier. Unfortunately this means `ctx` needs to be added to a lot of functions.

### DIFF
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/conditions"
@@ -71,7 +72,6 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	// Fetch the Hetzner bare metal host instance.
 	bmHost := &infrav1.HetznerBareMetalHost{}
-	// Use uncached APIReader to avoid cache issues.
 	err := r.Get(ctx, req.NamespacedName, bmHost)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -146,6 +146,8 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 		return reconcile.Result{}, fmt.Errorf("failed to get HetznerCluster: %w", err)
 	}
+
+	log = log.WithValues("HetznerCluster", klog.KObj(hetznerCluster))
 
 	// Fetch the Cluster.
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, hetznerCluster.ObjectMeta)
@@ -239,6 +241,7 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(ctx context.Con
 				return reconcile.Result{}, fmt.Errorf("Update() failed after setting ProvisioningState: %w", err)
 			}
 		}
+
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 
 		// Handle StateDeleting
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -82,6 +82,8 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	log = log.WithValues("state", bmHost.Spec.Status.ProvisioningState)
 
+	log.Info("Reconciling HetznerBareMetalHost") // TODO: Remove this log line after debugging
+
 	initialProvisioningState := bmHost.Spec.Status.ProvisioningState
 	defer func() {
 		// check duration of reconcile. Warn if it took too long.
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -79,7 +79,8 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	// Fetch the Hetzner bare metal host instance.
 	bmHost := &infrav1.HetznerBareMetalHost{}
-	err := r.Get(ctx, req.NamespacedName, bmHost)
+	// Use uncached APIReader to avoid cache issues.
+	err := r.APIReader.Get(ctx, req.NamespacedName, bmHost)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return reconcile.Result{}, nil
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -81,6 +81,7 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 	}
 
 	log = log.WithValues("state", bmHost.Spec.Status.ProvisioningState)
+	ctx = ctrl.LoggerInto(ctx, log)
 
 	log.Info("Reconciling HetznerBareMetalHost") // TODO: Remove this log line after debugging
 
@@ -166,8 +167,6 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 			return reconcile.Result{}, fmt.Errorf("failed to get HetznerBareMetalMachine: %w", err)
 		}
 	}
-
-	ctx = ctrl.LoggerInto(ctx, log)
 
 	// Get Hetzner robot api credentials
 	secretManager := secretutil.NewSecretManager(log, r.Client, r.APIReader)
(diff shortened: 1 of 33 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -592,46 +592,18 @@ func (s *Service) actionRegistering(_ context.Context) actionResult {
...
-
 	output := sshClient.GetHardwareDetailsDebug()
 	if output.Err != nil {
 		return actionError{err: fmt.Errorf("failed to obtain hardware for debugging: %w", output.Err)}
 	}
+
 	msg := fmt.Sprintf("%s\n\n", output.StdOut)
-	if out.StdErr != "" {
-		msg += fmt.Sprintf("stderr:\n%s\n\n", out.StdErr)
+	if output.StdErr != "" {
```
guettli, line 605: This was a typo in the old code, which as revealed by extracting the new method.

janiskemper (dismissed):

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -140,21 +141,18 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 	}
 	if err := r.Client.Get(ctx, hetznerClusterName, hetznerCluster); err != nil {
 		if apierrors.IsNotFound(err) {
+			log.Info("HetznerCluster not found. Requeueing")
 			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 		}
 		return reconcile.Result{}, fmt.Errorf("failed to get HetznerCluster: %w", err)
 	}
 
-	log = log.WithValues("HetznerCluster", klog.KObj(hetznerCluster))
```
janiskemper, line 148: why do you remove this?

pkg/services/baremetal/host/host.go
```diff
@@ -1097,6 +1111,14 @@ func (s *Service) actionPreProvisioning(ctx context.Context) actionResult {
 	}
 	sshClient := s.scope.SSHClientFactory.NewClient(in)
 
+	out := sshClient.GetHostName()
+	hostName := trimLineBreak(out.StdOut)
+	if hostName != rescue {
+		// This is unexpected. We should be in rescue mode.
+		ctrl.LoggerFrom(ctx).Info("pre-provision: expected rescue system, but found different hostname", "hostname", hostName, "sshOutput", out.String())
+		return actionContinue{delay: 10 * time.Second}
```
janiskemper, line 1119: should we requeue in this case? Is this PR for testing or for merging? For merging, I would expect that we handle this situation in a way that can be done in production. I think requeueing wouldn't be it. Or do you think this might be temporary?

janiskemper (changes requested):

controllers/hetznerbaremetalhost_controller.go
guettli: I removed that, because the logs are hard to read with the human eyes. The signal-to-noise ratio is low.

I think the HetznerCluster does not provide value.

But I can revert that in this PR and create a new one, if you prefer that.

pkg/services/baremetal/host/host.go
guettli: I want to merge that PR :-)

I guess this could happen because of a stale cache read by the controller.

If we see that machine hang there in production (which I do not expect), then we should handle that differently.

controllers/hetznerbaremetalhost_controller.go
janiskemper: In they CAPI docs they say that this is how it should be done. If we have good reasons, then we can remove it. But I would like to have that in a separate PR, including a reasoning why we should NOT follow CAPI here.

pkg/services/baremetal/host/host.go
janiskemper: but what is the value of this PR then? I thought it is about handling these cases in production in a better way. Currently, we would make them hang forever (which might have happened anyway of course). 

And which cache would this read from? the hostname on that server?  

Tbh I don't understand the use case for this really.. Can you maybe explain why you wanna do this change?

guettli: I have seen logs on a customer system which looked like the underlaying system changed during provisioning.

That's why I want to check before connecting to the rescue-system that the rescue system is active.

guettli: @janiskemper here is a second reason:

failed to reconcile HetznerBareMetalHost ci14972060597-sn76rg/bm-e2e-1724024: action \"ensure-provisioned\" failed: failed to handle incomplete boot - provisioning: unexpected hostname: bm-ci14972069472-kvra9l-control-plane-xh4gj

In our CI we get unexpected hostname because a different CI run uses the same machine. That's why I think that double checking that the rescue system is running makes sense.

guettli: @janiskemper I updated the PR: 09795339 

If there is a connection error, the code will re-try. If the hostname is unexpected a permanent error gets set.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -168,10 +168,6 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 		}
 	}
 
-	log = log.WithValues("HetznerBareMetalMachine", klog.KObj(hetznerBareMetalMachine))
```
janiskemper, line 171: here the change still exists

pkg/services/baremetal/host/host.go
```diff
@@ -592,46 +595,18 @@ func (s *Service) actionRegistering(_ context.Context) actionResult {
...
-		}
-		timeSinceReboot := "unknown"
-		if s.scope.HetznerBareMetalHost.Spec.Status.LastUpdated != nil {
-			timeSinceReboot = time.Since(s.scope.HetznerBareMetalHost.Spec.Status.LastUpdated.Time).String()
-		}
-
-		s.scope.Logger.Info("Could not reach rescue system. Will retry some seconds later.", "out", out.String(), "hostName", hostName,
-			"isSSHTimeoutError", isSSHTimeoutError, "isSSHConnectionRefusedError", isSSHConnectionRefusedError, "timeSinceReboot", timeSinceReboot)
-		return actionContinue{delay: 10 * time.Second}
+	ok, res := s.validateRescueSystemIsActive(ctx, sshClient)
```
janiskemper, line 598: can you maybe keep the original code here? Because you don't re-use the new utility function anymore, as you probably intended originally. If you don't re-use, I'd keep it as is

guettli: ok, I inlined the function which was used only once: 3e6fd3e4

controllers/hetznerbaremetalhost_controller.go
janiskemper: @guettli the change still exists. Can you get it back to the original state with regards to the logger? Except of course the additional value that you write into the loggers values.

guettli: I reverted these changes concerning WithValues, and created a new PR: [:seedling: log less values in hbmh Reconcile. by guettli · Pull Request #1606 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1606)

pkg/services/baremetal/host/host.go
```diff
@@ -608,13 +612,22 @@ func (s *Service) actionRegistering(_ context.Context) actionResult {
...
-		failed, err := s.handleIncompleteBoot(true, isSSHTimeoutError, isSSHConnectionRefusedError)
+		failed, err := s.handleIncompleteBoot(ctx, true, isSSHTimeoutError, isSSHConnectionRefusedError)
 		if failed {
 			return s.recordActionFailure(infrav1.PermanentError, err.Error())
 		}
 		if err != nil {
 			return actionError{err: fmt.Errorf(errMsgFailedHandlingIncompleteBoot, err)}
 		}
+
+		if !isSSHTimeoutError && !isSSHConnectionRefusedError {
```
janiskemper, line 623: are you sure about this code block? We do handle these cases in the handleIncompleteBoot function. Also this case here. We handle it in a way that says "boot didn't happen so we trigger another one". If you want to update this logic, then you should change the handleIncompleteBoot function IMO

guettli: I would love to update that, so that it is easier to understand. But not in this PR, and not now.

Thank you for your hint. I removed that.

pkg/services/baremetal/host/host.go
guettli: @janiskemper can this thread be resolved?

janiskemper (approved):

guettli:
Strange error in e2e test:

https://github.com/syself/cluster-api-provider-hetzner/actions/runs/15015378781/job/42192031779?pr=1598

> W0514 08:22:37.884145       1 reflector.go:547] k8s.io/client-go@v0.30.3/tools/cache/reflector.go:232: failed to list *v1.CertificateSigningRequest: Get "https://162.55.157.123:443/apis/certificates.k8s.io/v1/certificatesigningrequests?resourceVersion=871": tls: failed to verify certificate: x509: certificate is valid for 10.96.0.1, 136.243.69.167, 167.235.106.155, not 162.55.157.123

136.243.69.167 is  severID=1670788

162.55.157.123 is controlPlaneEndpoint and apiServerEndpoint in KubeadmConfig

janiskemper (approved):

PR #1599, opened 2025-05-07
:bug: Dev env bug fixes

madbbb:
Hello! I wanted to run Tilt locally and noticed that it tries to download 1.35.6 image which is not exist here: https://github.com/fluxcd/flux-benchmark/pkgs/container/kindest%2Fnode/versions?filters%5Bversion_type%5D=tagged

and I noticed that there was a line deleted by accident in cillium install step.

### DIFF
--- a/hack/kind-dev.sh
+++ b/hack/kind-dev.sh
@@ -18,7 +18,7 @@ set -o errexit
 set -o pipefail
 set -x
 
-K8S_VERSION=v1.31.6
+K8S_VERSION=v1.31.5
 
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}" || exit 1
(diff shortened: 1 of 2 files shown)


hack/kind-dev.sh
```diff
@@ -18,7 +18,7 @@ set -o errexit
 set -o pipefail
 set -x
 
-K8S_VERSION=v1.31.6
+K8S_VERSION=v1.31.5
```
guettli, line 21: yes, you are right. Please add this comment above that line:

```
# See `crane ls ghcr.io/fluxcd/kindest/node` for available versions
```

guettli (dismissed):
Thank you very much for the PR!

Just two small changes are needed to get it merged.

guettli:
@madbbb merged, thank you!

PR #1602, opened 2025-05-08
:seedling: Use logger instead of by to avoid e2e test failure.

guettli:
Use logger instead of by to avoid e2e test failure.


janiskemper (approved):
@guettli I don't think that this caused the test to fail. It was just a warning. But still seems to be valid

PR #1607, opened 2025-05-15
🐛 Only wait for api server pod healthy condition if cluster control plane is kubeadm-based

### LINKED ISSUE #1496: #1150 breaks control plane LB if CAPI provider doesn't set the APIServerPodHealthy condition
/kind bug

After upgrading CAPH, new control pane nodes are not added as targets to the API load balancer anymore.

No change in behaviour in this regard.

Some control plane providers, at least the Talos one don't set this condition. According to https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200506-conditions.md I understand this would be compliant behaviour:

```A set of additional Machine.Status.Conditions should be created in order to provide an observation point on the operational state of kubeadm control-plane machine components like e.g. EtcdReady, ApiServerPodReady, etc. (the other control plane providers are expected to implement different conditions at this level)```

While I do consider #1150 to be a great improvement for the consistency of the cluster, the underlying assumption that this condition will be present once the apiserver is ready seems to go too far. Thus, this feature should be optional and/or configurable to use e.g. request probes instead.

**Environment:**

- cluster-api-provider-hetzner version: v1.0.0
- Kubernetes version: v1.31.1
- OS (e.g. from `/etc/os-release`): Talos Linux v1.8.1

### PR DESCRIPTION
**Pull request by xDev789**
…ne is kubeadm-based

This PR mitigates a bug which prevented non-kubeadm control planes from being added as load balancer targets since API server pod healthy condition is a kubeadm-specific condition and other providers might not propagate it.

Fixes #1496

### CODE REVIEW

**Review by guettli — APPROVED** [review]
lgtm

**Comment by guettli** [comment]
I created a PR to test your changes. The e2e tests must be run by an org member.

**Comment by xDev789** [comment]
@madbbb a new version was just released today. unfortunately, the PR was not merged before the release, so I assume it is going to be a while. but it is really easy to use the patched version today if you use the official Cluster API operator like we do!

Here is an example of what your manifest InfrastructureProvider might look like:
```
apiVersion: operator.cluster.x-k8s.io/v1alpha2
kind: InfrastructureProvider
metadata:
  name: hetzner
  namespace: capi-infrastructure
spec:
  deployment:
    containers:
      - name: manager
        imageUrl: "ghcr.io/syself/caph-staging:pr-1621"
```

caph-staging:pr-1621 is basically the upstream (v1.0.4) release with my fix.

PR #1621, opened 2025-06-12
:seedling: Only wait for api server pod healthy condition if cluster control plane is kubeadm-based (via Syself PR)

guettli:
#1607


guettli:
merged.

PR #1625, opened 2025-07-08
[DO-NOT-MERGE] Add some debug logs to verify deletion

janiskemper:
Some debug logs to verify deletion


guettli:
@janiskemper is that PR still needed?

PR #1626, opened 2025-07-09
:seedling: Delete server, even it is currently in "initializing".

guettli:
When a hcloud server was initialing, and got deleted, then up to now nothing happend: endless RequeueAfter.

This PR fixes this.


janiskemper (approved):

guettli:
Hint: hcloud e2e fails, because Hetzner has an outage of machines. Merging should be save.

PR #1631, opened 2025-07-11
:seedling: Look for different error code while deleting server of lb

### LINKED ISSUE #1630: Hetzner API change leads to control planes not being deleted
/kind bug

We realized today that [this change](https://docs.hetzner.cloud/changelog#2025-07-09-bug-fixes-and-new-call-specific-error-code-for-adding-load-balancer-targets) of the Hetzner Cloud API breaks the deletion process of control planes.  

We have to fix it.

As far as I know, a restart of CAPH solves the problem temporarily once there are stuck control planes. 

This affects all users of CAPH, so that everyone will have to update to the latest CAPH version we will release soon, after the fix is merged.

Dhairya-Arora01:
We no longer get the error `load_balancer_not_found` if the target server was not found, instead we should check for the error `invalid_input`.

Fixes #1630 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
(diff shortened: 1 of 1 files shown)


janiskemper (commented):
can you add a comment with the error we see in the hcloud api plz and ideally explain shortly why you  check this

batistein (approved):

pkg/services/hcloud/server/server.go
```diff
@@ -663,7 +663,10 @@ func (s *Service) deleteServerOfLoadBalancer(ctx context.Context, server *hcloud
 	lb := &hcloud.LoadBalancer{ID: s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ID}
 
 	if err := s.scope.HCloudClient.DeleteTargetServerOfLoadBalancer(ctx, lb, server); err != nil {
-		// do not return an error in case the target was not found
+		// Do not return an error in case the target was not found.
+		// In case the target server was not found we will get an error similar to
+		// "server with ID xxxxx not found (invalid_input, xxxxxxx)".
+		// Therefor, if error code is "invalid_input" or "load_balancer_not_found" don't do anything.
```
janiskemper: typo

Issue #1493, opened 2024-10-18
Support adding vSwitch subnets to the network


dannynodies:
/kind feature

**Describe the solution you'd like**
At present, if I add a vSwitch subnet to the created network, it causes `NetworkReconcileFailed : failed to find network: multiple subnets not allowed`

I am attempting to route traffic between the cluster and some baremetal servers over the private network, and allowing the addition of a vSwitch subnet would permit this configuration to be applied.

Related to #1472 (since the changes required would conflict)

**Anything else you would like to add:**
I have a pretty good idea how this can be done - will submit a PR

**Environment:**

- cluster-api-provider-hetzner version: v1.0.0
- Kubernetes version: 1.30.5
- OS (e.g. from `/etc/os-release`): Ubuntu 22.04

dobe:
hi @dannynodies I have exactly the same problem. Did you find a workaround for this use case?

suyash-811:
Oh great, i'll give that a try. We are also waiting for #1472 to be merged, as this could (potentially) break the changes made in the PR mentioned by you. However, when this PR gets merged is still a big unknown.

Issue #213, opened 2022-06-04
Private network for bare metal host / cloud hybrid


pucilpet:
/kind feature

**Describe the solution you'd like**
It is possible to link bare metal servers to the same private network as the cloud servers. I think it would be good to have an option for the private networking for hybrid clusters.

**Anything else you would like to add:**
More info: https://docs.hetzner.com/cloud/networks/connect-dedi-vswitch/

Adding the vswitch is possible with the robot web service:
https://robot.your-server.de/doc/webservice/en.html#post-vswitch

batistein:
This feature is not implemented yet. Since we had several problems with the private network in the past (very high ping, interruptions, etc.), we do not use it for our offers, so it has less prio from our side.
But I would love to see any contribution on it.

shurshun:
Hi there! I would like to say thank you for such a handy product! I would also like to agree with the topic starter on the issue of private network support: we plan to order a set-up with a private network between servers (without the cloud) and, unfortunately, this is the only factor limiting the use of this provider :(

batistein:
@mads-sogaard we didn't implemented the private network stuff on Hetzner Dedicated. In our production cluster we use the konnectivity-service, Cilium Host Firewall and run mTLS everywhere so we do not have the use case for private network. But of course anyone is welcome to contribute a PR.

mstarostik:
I also figured the private networks aren't of that much use unless and until they offer more advanced routing setup between different private networks, NAT to public or the likes.
Anyway, one thing using a private network turned out to be still very useful to me is so I can get a known (private) IP for the API Server LB to pass as `k8sServiceHost` to Cilium when enabling its kube-proxy replacement. I'd be happy to simplify the setup some but haven't figured a way to pull the public IP for the LB into the ArgoCD app that provisions Cilium on newly created clusters.

galiev:
> Cilium Host Firewall / Calico Global Network Policy are very convenient for host protection, but how to protect kube-apiserver, the main problem is that the api server does not support the proxy protocol and therefore I cannot restrict access. If you have any ideas I'll be very grateful.

the load in my cluster is minimal, so the following option suited me: TalosControlPlane + Cilium Host Firewall
```yaml
apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
kind: TalosControlPlane
metadata:
  name: tcp
spec:
  controlPlaneConfig:
    controlplane:
      configPatches:
      - op: add
        path: /machine/network
        value:
[...]

PR #1633, opened 2025-07-15
✨ allow multiple subnets if first matches the configured one on HetznerCluster

fgrutsch:

We are currently running a setup where we have a k8s cluster running on Hetnzer cloud with a public IP and a private network. We also have bare metal nodes (setup manually and not running k8s) which we are connecting via a private network (vSwitch) to the k8s nodes. The setup works in general, but the controller reports errors due to multiple subnets restriction.

The idea of the PR is that we could allow multiple subnets in a network as long as the first subnet matches with the one that is configured on the `HetznerCluster` resource. When attaching a server to the network it will use the first subnet by default.

Wdyt about that solution? I have seen in multiple issues that you don't use private networks in your product and I think therefore this would be an acceptable solution to connect bare metal nodes <-> cloud nodes without making major modifications to the project's code.

- https://github.com/syself/cluster-api-provider-hetzner/issues/213
- https://github.com/syself/cluster-api-provider-hetzner/issues/1493

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/hcloud/network/network.go
+++ b/pkg/services/hcloud/network/network.go
@@ -173,7 +173,14 @@ func (s *Service) findNetwork(ctx context.Context) (*hcloud.Network, error) {
 	}
 
 	if len(networks[0].Subnets) > 1 {
-		return nil, fmt.Errorf("multiple subnets not allowed")
+		configuredSubnet := s.scope.HetznerCluster.Spec.HCloudNetwork.SubnetCIDRBlock
+		firstSubnet := networks[0].Subnets[0]
+
+		// Allow multiple subnets only if the first subnet matches the configured one. On attaching a server to a
+		// network the first subnet is used.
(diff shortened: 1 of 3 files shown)


fgrutsch:
Hi @suyash-811,

our use case is that we only have 1 cloud subnet and 1 vswitch subnet, therefore it is working. The problem with Hetzner's API back then was that you couldn't specify the subnet of the network where the server should be attach to. I checked the API again and it seems like that this is possible now (see [changelog](https://docs.hetzner.cloud/changelog#2025-09-11-attach-to-network-support-ip-range)).

With this API change I think the PR could be updated and improved by providing the `ip_range` of the configured subnet and then it should work. Also the validation if the first subnet matches the configured one can be changed by just checking the configured subnet is contained in the network only.

Unfortunately, I have no time right now to give it a try.

PR #1635, opened 2025-07-18
🌱 build(tools): make helm and hcloud downloads platform-agnostic

tdabasinskas:
Replace hardcoded `linux-amd64` architecture with dynamic `$(go env GOOS)-$(go env GOARCH)` to support multiple platforms and architectures during tool installation.


guettli:
@tdabasinskas Your PR looks good, I will merge it. On the other hand: Development by Syself is currently done exclusively on linux and amd64. I guess you will have issues, if you try to use a different platform. For the created container images arm64 works fine, too. It is used in production by Syself. 

Which platform do you want to use?

PR #1637, opened 2025-07-30
:seedling: Add SkipCreatingHetznerSecretInWorkloadCluster to HetznerCluster

### LINKED ISSUE #1636: Disable workload-cluster secret
/kind feature

Currently it looks like the `hetzner` secret always gets created within the user cluster. To avoid leaking this information to the wrong people I would prefer to deploy the Hetzner Cloud specific services outside of the user cluster. In that case the secret should not be created.

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- cluster-api-provider-hetzner version: v1.0.3
- Kubernetes version: 1.33.2
- OS (e.g. from `/etc/os-release`): Ubuntu 24.04

guettli:
SkipCreatingHetznerSecretInWorkloadCluster indicates whether the Hetzner secret should be created in the workload cluster. By default the secret gets created, so that the ccm (running in the wl-cluster) can use that secret. If you prefer to not reveal the secret in the workload cluster, you can set this to value to false, so that the secret is not created. Be sure to run the ccm outside of the workload cluster in that case, e.g. in the management cluster.

Closes #1636

---

Additionally, `make generate-mocks` was changed, so that in runs in the build-container.

Otherwise, you might get this error, if you run a newer version of Go (e.g. 1.24.4):

> 2025/07/30 10:27:57 internal error: package "net/http" without types was imported from "github.com/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/robot"


api/v1beta1/hetznercluster_types.go
```diff
@@ -64,6 +64,10 @@ type HetznerClusterSpec struct {
 	// HetznerSecretRef is a reference to a token to be used when reconciling this cluster.
 	// This is generated in the security section under API TOKENS. Read & write is necessary.
 	HetznerSecret HetznerSecretRef `json:"hetznerSecretRef"`
+
+	// SkipCreatingHetznerSecretInWorkloadCluster indicates whether the Hetzner secret should be created in the workload cluster. By default the secret gets created, so that the ccm (running in the wl-cluster) can use that secret. If you prefer to not reveal the secret in the wl-cluster, you can set this to value to false, so that the secret is not created. Be sure to run the ccm outside of the wl-cluster in that case, e.g. in the management cluster.
```
janiskemper: doesn't it make sense to put this text on multiple lines?

guettli: @janiskemper my IDE setting wraps automatically, so I do not need it. But you are right,  this line is too long.

I wrapped it.

janiskemper (approved):

janiskemper (approved):

PR #1639, opened 2025-07-30
:seedling: When Robot API reaches a timeout, retry later.

guettli:
# When Robot API reaches a timeout, retry later

The Robot API of Hetzner returned this:

> failed to get bare metal server: Get "https://robot-ws.your-server.de/server/1234": net/http: TLS handshake timeout

We do want to retry N seconds later. Up to now the provisioning process was stopped and started from scratch.


janiskemper (approved):

guettli:
e2e tests failed because something is wrong with tps-token. I merge nevertheless. We will do intensive testing before releasing.

PR #1650, opened 2025-08-28
:seedling: Refactor Hcloud Provisioning (before introducing --hcloud-image-url-command)

guettli:


pkg/services/hcloud/server/server.go
```diff
@@ -32,16 +32,22 @@ import (
...
 	"github.com/syself/cluster-api-provider-hetzner/pkg/scope"
+	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
 	hcloudutil "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/util"
 	"github.com/syself/cluster-api-provider-hetzner/pkg/utils"
 )
 
 const (
 	serverOffTimeout = 10 * time.Minute
+
+	// requeueAfterCreateServer: TODO get a good value. It should work for most cases, so
```
janiskemper: do we want to merge this PR if it is a TODO to find a good value?

pkg/services/hcloud/server/server.go
```diff
@@ -92,25 +98,56 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 
-	// try to find an existing server
-	server, err := s.findServer(ctx)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get server: %w", err)
+	var server *hcloud.Server
+	if s.scope.HCloudMachine.Spec.ProviderID != nil {
+		server, err = s.findServer(ctx)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("handleBootToRealOS: %w", err)
```
janiskemper: error message is unrelated

pkg/services/hcloud/server/server.go
```diff
@@ -92,25 +98,56 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+		}
+
+		if server == nil {
+			// This should not happen in production, but happens in tests.
```
janiskemper: shouldn't we then fix the tests?

pkg/services/hcloud/server/server.go
```diff
@@ -92,25 +98,56 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 
-	// if no server is found we have to create one
-	if server == nil {
-		server, err = s.createServer(ctx)
-		if err != nil {
-			if errors.Is(err, errServerCreateNotPossible) {
-				return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
-			}
-			return reconcile.Result{}, fmt.Errorf("failed to create server: %w", err)
+	switch s.scope.HCloudMachine.Status.BootState {
```
janiskemper, line 126: the boot state "running" currently seems to lead to an error being returned..

pkg/services/hcloud/server/server.go
```diff
@@ -92,25 +98,56 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 
+	server, err := s.createServerFromImageName(ctx)
+	if err != nil {
+		if errors.Is(err, errServerCreateNotPossible) {
+			return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to create server: %w", err)
+	}
 	s.scope.SetProviderID(server.ID)
+	return reconcile.Result{RequeueAfter: requeueAfterCreateServer}, nil
```
janiskemper: until now, we didn't requeue and did multiple other actions before requeuing. What's the reason to stop doing this? This changes the behavior, because for min 10 seconds (the requeue time) we don't have the information populated.

janiskemper: it seems to me as if you don't set the next state, even though you should if we reach this point..

pkg/services/hcloud/server/server.go
janiskemper, line 126: since the handling of "boot state running" doesn't exist, we never perform the respective actions again that we would have otherwise taken. 

However, I think we should. At least we should not change this behavior drastically (never doing any "drift correction" which we had until now) without a good reason and documenting it

pkg/services/hcloud/server/server.go
```diff
@@ -92,25 +98,56 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	conditions.MarkTrue(s.scope.HCloudMachine, infrav1.BootstrapReadyCondition)
 
-	// try to find an existing server
-	server, err := s.findServer(ctx)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get server: %w", err)
+	var server *hcloud.Server
+	if s.scope.HCloudMachine.Spec.ProviderID != nil {
```
janiskemper, line 96: why do you have this code at all? Why not making it part of the respective switch statement? we can make this whole section part of "findServer" and just call that function in the respective handleState functions

pkg/services/hcloud/server/server.go
```diff
@@ -405,53 +448,11 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
...
-
-		// If the SSH key name doesn't exist, append it
-		if !keyExists {
-			sshKeySpecs = append(sshKeySpecs, infrav1.SSHKey{Name: string(sshKeyName)})
-		}
-	}
-
-	// get all ssh keys that are stored in HCloud API
-	sshKeysAPI, err := s.scope.HCloudClient.ListSSHKeys(ctx, hcloud.SSHKeyListOpts{})
+	caphSSHKeys, hcloudSSHKeys, err := s.getSSHKeys(ctx)
```
janiskemper, line 500: this change is not documented in the PR description

pkg/services/hcloud/server/server.go
```diff
@@ -798,19 +862,22 @@ func (s *Service) createLabels() map[string]string {
 	}
 }
 
-func filterHCloudSSHKeys(sshKeysAPI []*hcloud.SSHKey, sshKeysSpec []infrav1.SSHKey) ([]*hcloud.SSHKey, error) {
-	sshKeysAPIMap := make(map[string]*hcloud.SSHKey)
-	for i, sshKey := range sshKeysAPI {
-		sshKeysAPIMap[sshKey.Name] = sshKeysAPI[i]
+// convertCaphSSHKeysToHcloudSSHKeys converts the ssh keys from the spec to hcloud ssh keys.
```
janiskemper: It doesn't convert checks whether there is a match and if that is the case, returns th respective ssh key. Plz adjust the name

pkg/services/hcloud/server/server.go
guettli: I will set the value to 100 milllisecond and to measure to find a good value.

pkg/services/hcloud/server/server.go
guettli: fixed.

pkg/services/hcloud/server/server.go
guettli: I thought about that again. It can happen in production: If you delete the hcloud machine on the web UI, then this will happen: the resource has a providerID, but it does not exist. I looked at the old code, and there a new server got created. I will adapt the PR so that we do the same. It is an edge-case but caph should handle that.

guettli: new code:

```go
		if server == nil {
			// The server did disappear in HCloud? Maybe it was delete via web-UI.
			// We delete the stale ProviderID and start provisioning again
			msg := "hcloud server no longer available. Provisioning again"
			ctrl.LoggerFrom(ctx).Error(errors.New(msg), msg,
				"ProviderID", *s.scope.HCloudMachine.Spec.ProviderID,
				"BootState", s.scope.HCloudMachine.Status.BootState,
				"BootStateSince", s.scope.HCloudMachine.Status.BootStateSince,
			)
			s.scope.HCloudMachine.Spec.ProviderID = nil
			s.scope.HCloudMachine.SetBootState(infrav1.HCloudBootStateUnset)
			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
		}
	}
```

pkg/services/hcloud/server/server.go
guettli: fixed.

pkg/services/hcloud/server/server.go
guettli: Yes, Status.Addresses and Status.InstanceState get set when the server is available now (~10 seconds later).

This was just a gut feeling. If the server is not reachable yet, then I think it is better to not publish the IP yet.

Looking at the old code, I guess that often that code was reached:

```go
	case hcloud.ServerStatusStarting:
		// Requeue here so that server does not switch back and forth between off and starting.
		// If we don't return here, the condition ServerAvailable would get marked as true in this
		// case. However, if the server is stuck and does not power on, we should not mark the
		// condition ServerAvailable as true to be able to remediate the server after a timeout.
		conditions.MarkFalse(
			s.scope.HCloudMachine,
			infrav1.ServerAvailableCondition,
			infrav1.ServerStartingReason,
			"server is starting",
		)
		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
```

The delay of one minute is quite long. Creating a server from snaphot takes ~8..11 seconds (via hcloud cli).

A new machine is 5 seconds in "initializing" and then 6 seconds in "starting" (results of current timing). The Requeue for unknown (initializing) was 10 seconds. This means the one minute was hit. But this is just a guess.

Are there reasons for choosing one minute?

---

The extra step was added, so we can jump to the ImageURL flow more easily.

janiskemper: I have no opinion about the 1 minute. 
But then why not setting the condition accordingly to whatever you show there, as well as the instance state (which will be "starting")? I'm not sure about the IP.

guettli: yes, up to now this was handled via the next reconcile. So no test failed. I added SetBootState() to avoid the additional Reconcile.

pkg/services/hcloud/server/server.go
guettli: @janiskemper yes, the drift correction makes sense. I moved that code into the final state.

pkg/services/hcloud/server/server.go
guettli: Getting the server is needed in all states except StateUnset. I would like to avoid to repeat this block again and again:

```go
	var server *hcloud.Server
	if s.scope.HCloudMachine.Spec.ProviderID != nil {
		server, err = s.findServer(ctx)
		if err != nil {
			return reconcile.Result{}, fmt.Errorf("findServer: %w", err)
		}

		if server == nil {
			// The server did disappear in HCloud? Maybe it was delete via web-UI.
			// We delete the stale ProviderID and start provisioning again
			msg := "hcloud server no longer available. Provisioning again"
			ctrl.LoggerFrom(ctx).Error(errors.New(msg), msg,
				"ProviderID", *s.scope.HCloudMachine.Spec.ProviderID,
				"BootState", s.scope.HCloudMachine.Status.BootState,
				"BootStateSince", s.scope.HCloudMachine.Status.BootStateSince,
			)
			s.scope.HCloudMachine.Spec.ProviderID = nil
			s.scope.HCloudMachine.SetBootState(infrav1.HCloudBootStateUnset)
			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
		}
	}
```

We could add that to findServer(). But then we need to change the interface, because a Result gets returned.

I am fine with putting the current solution, or putting that into findServer and changing its result to (server, reconcile.Result).

What do you prefer?

pkg/services/hcloud/server/server.go
guettli: i added that to the PR description:

>  filterHCloudSSHKeys: The function does not filter, so it was renamed to getSSHKeys. Additionally, it was refactored to reduce the size of the createServer function.

pkg/services/hcloud/server/server.go
guettli: @janiskemper I updated the comment

// Returns error, if a caphSSHKey is not in allHcloudSSHKeys.

Please tell me your desired name.

pkg/services/hcloud/server/server.go
janiskemper: you cannot provision again.. If the server is gone then you should give a FailureMessage so that the server is killed!

EDIT: This is actually an improvement

pkg/services/hcloud/server/server.go
janiskemper: let's leave that open for a while and see what is the best solution for this case

pkg/services/hcloud/server/server.go
janiskemper: well the original name actually made more sense, didn't it? I'm happy for suggestions from your side, taking into account the fact that it is mainly about matching

pkg/services/hcloud/server/server.go
guettli: @janiskemper I set a failure now. b29766fa

pkg/services/hcloud/server/server.go
guettli: @janiskemper there is a major difference between provisioning a vanilla image (like ubuntu-YY-MM) and provisioning a snaphost. First takes ~11s, the second ~160s.

The boolean Image.RapidDeploy is available. This gets used to set RequeueAfter accordingly. Commit: [:seedling: Refactor Hcloud Provisioning, use Status.BootState by guettli · Pull Request #1650 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1650/commits/98b16793a25aa93ca1d40db3aba0e04a1c7c4e89)

pkg/services/hcloud/server/server.go
guettli: @janiskemper the Status gets updated now directly after the server was created, and during each Reconcile.

See: 23be79c3

pkg/services/hcloud/server/server.go
guettli: @janiskemper I am fine with the current implementation, but I could move the findServer() call into a helper functions, and call that helper function in each handleBootState...()

pkg/services/hcloud/server/server.go
guettli:
```go
// convertCaphSSHKeysToHcloudSSHKeys converts the ssh keys from the spec to hcloud ssh keys.
// allHcloudSSHKeys contains all keys, returned by hcloudClient.ListSSHKeys().
// Returns error, if a caphSSHKey is not in allHcloudSSHKeys.
func convertCaphSSHKeysToHcloudSSHKeys(allHcloudSSHKeys []*hcloud.SSHKey, caphSSHKeys []infrav1.SSHKey) ([]*hcloud.SSHKey, error) {
```

for me docstring, function name and implementation fit. But maybe I am missing something.

The old name (filterHCloudSSHKeys) was confusing (at least for me), because the result was not filtered. The length of the output (hcloud-ssh-keys) is always the length of the input (caph-ssh-keys).

janiskemper: Okay, I checked it again in the original code and was slightly wrong. It is not a filter operation (which I think it was in the past at one point), but an operation which ensures that all ssh keys that are specified in the spec also exist in the API. So it is kind of "ensure existance". 

I would therefore say we can maybe name it "ensureSSHKeys".

guettli: for me, a function called `ensureSSHKeys()` does not have a return value (except an error).

The above method takes caphSSHKeys and for each key in that list a corresponding hcloud-key gets returned.

For me this is a conversion. We could use `mapSpecSSHKeysToHCloud`, too. Or just general `convertSSHKeys`

api/v1beta1/hcloudmachine_types.go
```diff
@@ -128,6 +141,15 @@ func (r *HCloudMachine) SetConditions(conditions clusterv1.Conditions) {
 	r.Status.Conditions = conditions
 }
 
+// SetBootState sets Status.BootStates and updates Status.BootStateSince.
+func (r *HCloudMachine) SetBootState(bootState HCloudBootState) {
+	if r.Status.BootState == bootState {
```
janiskemper, line 146: this is not necessary

api/v1beta1/types.go
```diff
@@ -263,3 +263,17 @@ func (s *HCloudNetworkSpec) IsZero() bool {
 	}
 	return true
 }
+
+// HCloudBootState defines the boot state of an HCloud server during provisioning.
```
janiskemper: it's not only "during provisioning" but in general. I think we should remove "during provisioning".

controllers/hcloudmachine_controller.go
```diff
@@ -157,6 +159,36 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 			res = reconcile.Result{}
 			reterr = errors.Join(reterr, err)
 		}
+
+		readyReason := conditions.GetReason(machineScope.HCloudMachine, clusterv1.ReadyCondition)
+		readyMessage := conditions.GetMessage(machineScope.HCloudMachine, clusterv1.ReadyCondition)
+
+		duration := time.Since(startReconcile)
```
janiskemper, line 193: do we have a good reason to keep this in the code block that you added in hcloudmachine_controller? If not, I'd prefer to remove it. We can always add it for testing something specific

pkg/services/hcloud/server/server.go
```diff
@@ -92,42 +105,97 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			)
+
+			s.scope.SetError(msg, capierrors.CreateMachineError)
+			s.scope.HCloudMachine.SetBootState(infrav1.HCloudBootStateUnset)
+			record.Warn(s.scope.HCloudMachine, "NoHCloudServerFound", msg)
+
+			// no need to requeue.
+			return reconcile.Result{}, nil
+		}
+		updateHCloudMachineStatusFromServer(s.scope.HCloudMachine, server)
```
janiskemper: please leave an empty line before this

pkg/services/hcloud/server/server.go
```diff
@@ -92,42 +105,97 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			ctrl.LoggerFrom(ctx).Info("Updating old resource (pre BootState)",
+				"to", m.Status.BootState)
+			return reconcile.Result{RequeueAfter: requeueImmediately}, nil
 		}
+
+		m.SetBootState(infrav1.HCloudBootStateOperatingSystemRunning)
+		ctrl.LoggerFrom(ctx).Info("Updating old resource (pre BootState)",
+			"to", m.Status.BootState)
+
+		// Requeue once the new way. But in most cases nothing should have changed.
```
janiskemper: what do you mean by this? And you just set it on "running", what else is supposed to happen?

pkg/services/hcloud/server/server.go
```diff
@@ -342,23 +420,29 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, server *h
...
 			s.scope.HCloudMachine,
 			"FailedGetBootstrapData",
 			err.Error(),
 		)
-		return nil, fmt.Errorf("failed to get raw bootstrap data: %s", err)
+		return nil, nil, fmt.Errorf("failed to get raw bootstrap data: %s", err)
 	}
-
-	image, err := s.getServerImage(ctx)
+	image, err := getServerImage(ctx, s.scope.HCloudClient, s.scope.HCloudMachine, s.scope.HCloudMachine.Spec.Type, s.scope.HCloudMachine.Spec.ImageName)
```
janiskemper: why did you change this from a method to a function? We can skip all arguments if we use the scope.

pkg/services/hcloud/server/server.go
```diff
@@ -342,23 +420,29 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, server *h
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("createServerFromImageName: failed to get server image: %w", err)
```
janiskemper: we have another wrapper one level up, so you don't have to add the function name here IMO

pkg/services/hcloud/server/server.go
janiskemper: plz leave empty lines here in this function

pkg/services/hcloud/server/server.go
```diff
@@ -492,24 +534,91 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
...
-	s.scope.HCloudMachine.Status.SSHKeys = sshKeySpecs
+	s.scope.HCloudMachine.Status.SSHKeys = caphSSHKeys
 
 	conditions.MarkTrue(s.scope.HCloudMachine, infrav1.ServerCreateSucceededCondition)
 	record.Eventf(s.scope.HCloudMachine, "SuccessfulCreate", "Created new server %s with ID %d", server.Name, server.ID)
 	return server, nil
 }
 
-func (s *Service) getServerImage(ctx context.Context) (*hcloud.Image, error) {
+// getSSHKeys returns the sshKey slice for the status and the sshKey slice for the createServer API
```
janiskemper: please improve this explanation further. It takes the ssh keys referenced in the spec of either hcloudMachine or HetznerCluster, as well as the hetzner secret. 

It then lists all available ssh keys that have been uploaded in Hcloud. It ensures that all keys referenced in the spec or the secret exist in HCloud and returns an error if that is not the case. 

It then returns for each ssh key referenced in the spec or the secret two objects, one as input for the hcloud create server flow, another one for adding the ssh key to the status of the HcloudMachine object.

pkg/services/hcloud/server/server.go
janiskemper: why do you wanna update the status here already? Shouldn't we do that in the main  handle functions?

janiskemper (changes requested):

api/v1beta1/hcloudmachine_types.go
Dhairya-Arora01: won't we then always update the BootStateSince timestamp ?

janiskemper: ah that's the catch. Alright!

pkg/services/hcloud/server/server.go
janiskemper: @Dhairya-Arora01 do you think it makes sense to change the function findServer in a way that this additional code is integrated there? And then, that we just call findServer in the respective handle... functions, where we need it? Or do you prefer it at this place? 

I personally think that it might make more sense in the handle functions because then we have all relevant steps in there

pkg/services/hcloud/server/server.go
janiskemper: @Dhairya-Arora01

pkg/services/hcloud/server/server.go
Dhairya-Arora01: I don't think so... setting status in the main reconcile function make much sense.
And if we try to set them in the handle functions then we will be just repeating the code.

api/v1beta1/hcloudmachine_validation_test.go
```diff
@@ -126,14 +132,14 @@ func TestValidateHCloudMachineSpec(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := validateHCloudMachineSpec(tt.args.oldSpec, tt.args.newSpec)
+			got := validateHCloudMachineSpecUpdate(tt.args.oldSpec, tt.args.newSpec)
 
 			if len(got) == 0 {
-				assert.Empty(t, got)
+				assert.Empty(t, tt.want)
```
janiskemper: @Dhairya-Arora01 shouldn't this whole test be different? "if len(want) == 0 then ..."?

pkg/services/hcloud/server/server.go
```diff
@@ -92,42 +104,96 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			return reconcile.Result{}, fmt.Errorf("findServer: %w", err)
+		}
+
+		if server == nil {
+			// The server did disappear in HCloud? Maybe it was delete via web-UI.
+			// We set MachineError. CAPI will delete machine.
+			msg := fmt.Sprintf("hcloud server (%q) no longer available. Setting MachineError.",
+				*s.scope.HCloudMachine.Spec.ProviderID)
+
+			ctrl.LoggerFrom(ctx).Error(errors.New(msg), msg,
```
janiskemper: scope has a logger. Let's replace all "ctrl.LoggerFrom"

pkg/services/hcloud/server/server.go
```diff
@@ -92,42 +104,96 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+}
+
+func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, error) {
+	m := s.scope.HCloudMachine
+
+	if m.Spec.ProviderID != nil && *m.Spec.ProviderID != "" {
+		// This machine seems to be an existing machine which was created before introducing
+		// Status.BootState.
+
+		if !m.Status.Ready {
```
janiskemper: this code is duplicate in the two cases. Can you just put "SetBootState" in the if statement?

pkg/services/hcloud/server/server.go
```diff
@@ -92,42 +104,96 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	conditions.MarkTrue(s.scope.HCloudMachine, infrav1.BootstrapReadyCondition)
 
-	// try to find an existing server
-	server, err := s.findServer(ctx)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get server: %w", err)
+	var server *hcloud.Server
+	if s.scope.HCloudMachine.Spec.ProviderID != nil {
```
janiskemper, line 96: @Dhairya-Arora01 can you add a unit test for this case (of not finding a server but spec.providerID is set)? You can do it simply by mocking and calling "Reconcile" with a certain state

pkg/services/hcloud/server/server.go
```diff
@@ -702,6 +823,9 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 		}
 	}
 
+	logger := ctrl.LoggerFrom(ctx)
```
janiskemper: logger is part of scope

api/v1beta1/hcloudmachine_validation_test.go
```diff
@@ -130,21 +130,24 @@ func TestValidateHCloudMachineSpecUpdate(t *testing.T) {
...
 	}
-
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := validateHCloudMachineSpecUpdate(tt.args.oldSpec, tt.args.newSpec)
 
-			if tt.want == nil {
-				assert.Empty(t, got)
-				return
+			if len(got) == 0 {
```
janiskemper: @guettli I think this was not intended. You changed the previous code that fixed a bug here in the test

api/v1beta1/types.go
janiskemper: @guettli this was on purpose, please change it back

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 			conditions.MarkTrue(hcloudMachine, infrav1.HCloudTokenAvailableCondition)
 		}
 
+		if hcloudMachine.Status.BootState != infrav1.HCloudBootStateOperatingSystemRunning {
+			conditions.MarkFalse(hcloudMachine, infrav1.ServerAvailableCondition,
+				string(hcloudMachine.Status.BootState), clusterv1.ConditionSeverityWarning,
```
janiskemper: I think this should be "Info" not warning

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
...
+				"%s", hcloudMachine.Status.BootStateMessage)
+		}
+		if reterr != nil {
```
janiskemper: this will also lead to a change in the condition if the error is completely different. I don't think this is a good idea. This will lead to some unexpected situations. Let's always set the condition when we want to do it

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 			conditions.MarkTrue(hcloudMachine, infrav1.HCloudTokenAvailableCondition)
 		}
 
+		if hcloudMachine.Status.BootState != infrav1.HCloudBootStateOperatingSystemRunning {
+			conditions.MarkFalse(hcloudMachine, infrav1.ServerAvailableCondition,
```
janiskemper: this will overwrite situations where you explicitly set this because of an error. I would suggest that we do this in the main code exactly in the situations where we want to, instead of overwriting it here

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
...
+				string(hcloudMachine.Status.BootState), clusterv1.ConditionSeverityWarning,
+				"%s", reterr.Error())
+		}
+		// the Close() will use PatchHelper to store the changes.
 		if err := machineScope.Close(ctx); err != nil {
 			res = reconcile.Result{}
 			reterr = errors.Join(reterr, err)
 		}
 
+		if !cmp.Equal(initialHCloudMachine, hcloudMachine) {
```
janiskemper, line 166: Can you give a reference where a similar pattern is done? I have never seen it and would like to understand

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
...
+		}
+
 		readyReason := conditions.GetReason(machineScope.HCloudMachine, clusterv1.ReadyCondition)
 		readyMessage := conditions.GetMessage(machineScope.HCloudMachine, clusterv1.ReadyCondition)
 
-		if initialHcloudMachine.Status.BootState != machineScope.HCloudMachine.Status.BootState {
-			startBootState := initialHcloudMachine.Status.BootStateSince
+		duration := time.Since(startReconcile)
+
+		if duration > 5*time.Second {
```
janiskemper, line 195: Is there any possibility that this might happen? If not, let's remove it

controllers/hcloudmachine_controller.go
```diff
@@ -155,22 +155,72 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
...
+				"reterr", reterr,
+				"oldState", initialHCloudMachine.Status.BootState,
+				"newState", machineScope.HCloudMachine.Status.BootState,
+				"readyReason", readyReason,
+				"readyMessage", readyMessage,
+			)
+		}
+
+		if initialHCloudMachine.Status.BootState != machineScope.HCloudMachine.Status.BootState {
+			// BootState changed. Wait until the update arrived in the local cache to ensure
```
janiskemper: this comment - does it refer to the defer statement? This is quite misleading

pkg/services/hcloud/server/server.go
```diff
@@ -54,9 +54,28 @@ const (
 	// Continuous RequeueAfter in BootToRealOS.
 	requeueIntervalBootToRealOS = 10 * time.Second
 
+	// requeueAfterEnableRescueSystem: the hcloud API needs roughly 3 seconds
```
janiskemper: I would suggest that we move all of these different requeues to the respective places where they are used. Right now, this is quite hard to read, both here in the constants and below in the code. If they are not used in more than 2-3 places, I'd remove the constant and do it in the code

pkg/services/hcloud/server/server.go
```diff
@@ -145,27 +165,34 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
-		if !m.Status.Ready {
-			m.SetBootState(infrav1.HCloudBootStateBootToRealOS)
-		} else {
-			m.SetBootState(infrav1.HCloudBootStateOperatingSystemRunning)
+		if !hm.Status.Ready {
+			hm.SetBootState(infrav1.HCloudBootStateBootToRealOS)
+			msg := fmt.Sprintf("Updating old resource (pre BootState) to %s",
+				hm.Status.BootState)
+			ctrl.LoggerFrom(ctx).Info(msg)
+			hm.Status.BootStateMessage = msg
```
janiskemper: why do we need this new property? cannot we just use the conditions for it?

pkg/services/hcloud/server/server.go
```diff
@@ -145,27 +165,34 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
-	if m.Spec.ProviderID != nil && *m.Spec.ProviderID != "" {
+	if hm.Spec.ProviderID != nil && *hm.Spec.ProviderID != "" {
 		// This machine seems to be an existing machine which was created before introducing
 		// Status.BootState.
 
-		if !m.Status.Ready {
-			m.SetBootState(infrav1.HCloudBootStateBootToRealOS)
-		} else {
-			m.SetBootState(infrav1.HCloudBootStateOperatingSystemRunning)
+		if !hm.Status.Ready {
```
janiskemper, line 146: this if statement was improved in the existing code. Please use the original code

api/v1beta1/types.go
guettli: @janiskemper done.

api/v1beta1/hcloudmachine_validation_test.go
guettli: @janiskemper the code of Dhairya is now at this place again.

controllers/hcloudmachine_controller.go
guettli: @janiskemper ok, setting ConditionSeverityInfo now.

controllers/hcloudmachine_controller.go
guettli: ok, the code no longer shows the error in the Condition. I think this is very handy.

But for me, it is more important to get the PR merged.

controllers/hcloudmachine_controller.go
guettli: ok, I removed BootStateMessage.

controllers/hcloudmachine_controller.go
guettli: a reference: capi has func `waitForMachineCreation()`: https://github.com/kubernetes-sigs/cluster-api/blob/737edc88faed28b0bd81f6768ebe2ea276fdbe3a/exp/internal/controllers/machinepool_controller_phases.go#L526

controllers/hcloudmachine_controller.go
guettli: yes, it could be. I have seen it during manual testing. 

We can remove it, when we have a good monitoring of metrics.

Controller-runtime exports metrics like this.

controllers/hcloudmachine_controller.go
guettli: this comment was dead. Removed.

pkg/services/hcloud/server/server.go
guettli: ok, I inlined the requeue-constants, except for `requeueImmediately`.

pkg/services/hcloud/server/server.go
guettli: yes, removed BootStateMessage

pkg/services/hcloud/server/server.go
guettli: done.

controllers/hcloudmachine_controller.go
janiskemper: okay, and you are sure that the resourceVersion definitely changes right? 

Can you also add the reason as comment why you handle this not found error in that way?

controllers/hcloudmachine_controller.go
janiskemper: and what is the reason then? If you have seen it, we should fix the problem

pkg/services/hcloud/server/server.go
janiskemper: you still set twice the same message and introduced a new logger. Please check the code before you changed it now, we can use the same thing still.

controllers/hcloudmachine_controller.go
```diff
@@ -77,7 +79,7 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 
-	log = log.WithValues("HCloudMachine", klog.KObj(hcloudMachine))
+	log = log.WithValues("HCloudMachine", hcloudMachine.Name)
```
janiskemper: capi tells us we should do it with KObj in the guides, if they didn't change that.

pkg/services/hcloud/server/server.go
janiskemper: IMO this should be removed  @guettli, you added it back

pkg/services/hcloud/server/server.go
```diff
@@ -842,6 +858,9 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 		return nil, nil
 	}
 
+	logger := ctrl.LoggerFrom(ctx)
```
janiskemper: can you add the scope logger for this?

janiskemper (dismissed):

controllers/hcloudmachine_controller.go
guettli: > okay, and you are sure that the resourceVersion definitely changes right?

yes, resourceVersion changes when something changes (can be spec, status, metadata, ....)

(Generation increases on Spec changes, and does not change if you update the Status.)

guettli: > Can you also add the reason as comment why you handle this not found error in that way?

done.

controllers/hcloudmachine_controller.go
guettli: > and what is the reason then? If you have seen it, we should fix the problem

I will do so. But not now. Prio is lower, but it is on my radar.

pkg/services/hcloud/server/server.go
guettli: Yes, I am sorry. Something went wrong here. I removed the duplicate line and use the Logger from the Scope now.

controllers/hcloudmachine_controller.go
guettli: Yes, you are right:

> - Developers MUST use `klog.KObj` or `klog.KRef` functions when logging key value pairs for Kubernetes objects, thus

This is sooo verbose. For me, it makes no sense. This just duplicates values. The namespace is already in the key-value pairs which get logged.

I revert that, and need to find a way to reduce the key-value pairs with a tool.

pkg/services/hcloud/server/server.go
guettli: `updateHCloudMachineStatusFromServer(s.scope.HCloudMachine, server)`

> IMO this should be removed @guettli, you added it back

(somehow I can not reply to your message above that message)

I added that on purpose. I think it makes sense to keep the status updated. Is there a reason to not update the Status?

janiskemper: the status should always get updated in the handle functions already IMO. I think that is already the case. 

I understand fetching the server object once can reduce some lines of code, so that it is maybe better to do it outside of the handle functions. However, I would keep the logic minimal

guettli: @janiskemper I understand your concern. I added a comment, so that the intention is clear:

```go

		// We found a server, now update the Status. We do that for all BootStates,
		// so we can keep the handleBootState functions a bit shorter.
		updateHCloudMachineStatusFromServer(s.scope.HCloudMachine, server)
```

janiskemper: I would not like to keep the functions shorter though :sweat_smile:  They should contain the main logic! This is making them less transparent actually. That's why my point was to remove it and put it in the main one, where we also do the other status updates.

guettli: @janiskemper I disagree, but I implement your wish. `updateHCloudMachineStatusFromServer(hm, server)` gets called over and over again in each handleBootState func now.

janiskemper: thanks. I think you didn't push though.

guettli: @janiskemper is pusht now.

janiskemper (commented):
@Dhairya-Arora01 plz have a final look. From my side LGTM

PR #1679, opened 2025-09-19
:seedling: Provision baremetal via --baremetal-image-url-command

guettli:


api/v1beta1/hetznerbaremetalmachine_types.go
```diff
@@ -197,6 +201,9 @@ type Image struct {
 // GetDetails returns the path of the image and whether the image has to be downloaded.
 func (image Image) GetDetails() (imagePath string, needsDownload bool, errorMessage string) {
 	// If image is set, then the URL is also set and we have to download a remote file
+	if image.UseCustomImageURLCommand {
+		return "", false, "internal error: image.UseCustomImageURLCommand is active. Method GetDetails() should be used for the tranditional way (without image-url-command)."
```
janiskemper: typo

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+			"%s", msg)
+		return s.recordActionFailure(infrav1.FatalError, msg)
+	}
+
+	switch state {
+	case sshclient.ImageURLCommandStateRunning:
+		conditions.MarkFalse(host, infrav1.ServerAvailableCondition,
+			string(host.Spec.Status.ProvisioningState), clusterv1.ConditionSeverityInfo,
+			"image-url-command still running")
+		return actionContinue{delay: 5 * time.Second}
```
janiskemper: 5 seconds is very often. Shouldn't we make it 10? If it runs for multiple minutes?

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+
+	case sshclient.ImageURLCommandStateFinishedSuccessfully:
+		conditions.MarkFalse(host, infrav1.ServerAvailableCondition,
+			string(host.Spec.Status.ProvisioningState), clusterv1.ConditionSeverityInfo,
+			"baremetal-image-url-command finished successfully. Rebooting now")
+
+		record.Event(s.scope.HetznerBareMetalHost, "ImageURLCommandOutput", logFile)
+		s.scope.Logger.Info("ImageURLCommandOutput", "logFile", logFile)
```
janiskemper, line 1248: the event will usually be logged as well. I think this will make it be printed twice

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		out := sshClient.Reboot()
+		if err := handleSSHError(out); err != nil {
+			err = fmt.Errorf("failed to reboot server (after install-image): %w", err)
+			record.Warn(s.scope.HetznerBareMetalHost, "RebootFailed", err.Error())
+			return actionError{err: err}
+		}
+		msg := "machine image and cloud-init data got installed (via image-url-command)"
+		createSSHRebootEvent(ctx, s.scope.HetznerBareMetalHost, msg)
+
+		s.scope.Logger.Info(msg, "stdout", out.StdOut, "stderr", out.StdErr)
```
janiskemper: do you wanna remove this print statement? it looks like it is for testing only

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
```
janiskemper: this is part of the SaveHostAndReturn function

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+			"%s", msg)
+
+		// clear potential errors - all done
+		s.scope.HetznerBareMetalHost.ClearError()
+		return actionComplete{}
+
+	case sshclient.ImageURLCommandStateFailed:
+		record.Warn(s.scope.HetznerBareMetalHost, "InstallImageNotSuccessful", logFile)
+		msg := "image-url-command failed"
+		s.scope.Logger.Error(nil, msg, "logFile", logFile)
```
janiskemper, line 1274: this is unnecessary with the event I believe

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			string(host.Spec.Status.ProvisioningState),
+			clusterv1.ConditionSeverityInfo,
+			"%s", msg)
+
+		// clear potential errors - all done
+		s.scope.HetznerBareMetalHost.ClearError()
+		return actionComplete{}
+
+	case sshclient.ImageURLCommandStateFailed:
```
janiskemper, line 1271: you return actionError here.. Shouldn't you mark this as a fatalError?

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,122 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+
+	case sshclient.ImageURLCommandStateNotStarted:
+		data, err := s.scope.GetRawBootstrapData(ctx)
+		if err != nil {
+			return actionError{err: fmt.Errorf("baremetal GetRawBootstrapData failed: %w", err)}
+		}
+		exitStatus, stdoutStderr, err := sshClient.StartImageURLCommand(ctx, s.scope.ImageURLCommand, s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Image.URL, data, s.scope.Name())
+		if err != nil {
+			err := fmt.Errorf("StartImageURLCommand failed (retrying): %w", err)
+			// This could be a temporary network error. Retry.
```
janiskemper, line 1311: if we say this is necessary, then we should probably set a condition and/or event

api/v1beta1/hetznerbaremetalmachine_types.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: yes, removed.

pkg/services/baremetal/host/host.go
guettli: @janiskemper please elaborate. SaveHostAndReturn looks like this:

```go 
func SaveHostAndReturn(ctx context.Context, cl client.Client, host *infrav1.HetznerBareMetalHost) (res reconcile.Result, err error) {
	t := metav1.Now()
	host.Spec.Status.LastUpdated = &t

	if err := cl.Update(ctx, host); err != nil {
		if apierrors.IsConflict(err) {
			log := ctrl.LoggerFrom(ctx)
			log.Info("conflict error. Retrying", "err", err)
			return reconcile.Result{Requeue: true}, nil
		}
		return reconcile.Result{}, fmt.Errorf("failed to update host object: %w", err)
	}
	return res, nil
}
```

No Condition gets set.

pkg/services/baremetal/host/host.go
guettli: yes, I use that now:

```go 
		return s.recordActionFailure(infrav1.FatalError, msg)
```

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: @janiskemper I tested it. The line created by `s.scope.Logger.Info()` is visible. No second line. Events get only logged if you raise the loglevel.

pkg/services/baremetal/host/host.go
guettli: Events do not get logged automatically (see above).

pkg/services/baremetal/host/host.go
janiskemper: I meant markProvisioningPending and ReconcileState

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -87,6 +90,62 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 		return reconcile.Result{}, err
 	}
 
+	// ----------------------------------------------------------------
+	// Start: avoid conflict errors. Wait until local cache is up-to-date
+	// Won't be needed once this was implemented:
+	// https://github.com/kubernetes-sigs/controller-runtime/issues/3320
+	initialHost := bmHost.DeepCopy()
+	defer func() {
+		if cmp.Equal(initialHost, bmHost) {
```
janiskemper, line 105: IMO we should check this (and in the other PRs where you added it) in more detail. In which cases do we want to avoid this problem? When is it necessary? I'm sure it won't be necessary in all cases.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -87,6 +90,62 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+			latestFromLocalCache := &infrav1.HCloudMachine{}
+			getErr := r.Get(ctx, client.ObjectKeyFromObject(apiserverHost), latestFromLocalCache)
+			if apierrors.IsNotFound(getErr) {
+				// the object was deleted. All is fine.
+				return true, nil
+			}
+			if getErr != nil {
+				return false, getErr
+			}
+			if len(latestFromLocalCache.ResourceVersion) > len(apiserverRV) {
```
janiskemper: why don't we compare numbers?

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,142 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+
+	switch state {
+	case sshclient.ImageURLCommandStateRunning:
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			string(host.Spec.Status.ProvisioningState), clusterv1.ConditionSeverityInfo,
+			"image-url-command still running")
+		return actionContinue{delay: 10 * time.Second}
+
+	case sshclient.ImageURLCommandStateFinishedSuccessfully:
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
```
janiskemper: I believe that there is no point setting the conditions here. The only longer running action is the image command and I think if the condition instead just shows the state it is the same level of information. 
The events and logs would give the necessary information around the reboot and whether the image URl command finished successfully or not

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,142 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		if s.scope.ImageURLCommand == "" {
+			err = errors.New("internal error: --baremetal-image-url-command is not set?")
+			s.scope.Logger.Error(err, "")
+			conditions.MarkFalse(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition,
+				string(s.scope.HetznerBareMetalHost.Spec.Status.ProvisioningState),
+				clusterv1.ConditionSeverityWarning,
+				"%s", err.Error())
+			return actionError{err: err}
+		}
+		// get the information about storage devices again to have the latest names.
```
janiskemper, line 1298: this seems to be a new feature. Is that correct? Can you ideally make a separate PR out of this, to make it visible to someone else?

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,142 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		exitStatus, stdoutStderr, err := sshClient.StartImageURLCommand(ctx, s.scope.ImageURLCommand, s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.Image.URL, data, s.scope.Hostname(), deviceNames)
+		if err != nil {
+			err := fmt.Errorf("StartImageURLCommand failed (retrying): %w", err)
+			// This could be a temporary network error. Retry.
+			s.scope.Logger.Error(err, "",
+				"ImageURLCommand", s.scope.ImageURLCommand,
+				"exitStatus", exitStatus,
+				"stdoutStderr", stdoutStderr)
+			conditions.MarkFalse(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition,
+				string(s.scope.HetznerBareMetalHost.Spec.Status.ProvisioningState),
```
janiskemper: are you sure that this wouldn't get overwritten by ReconcileState and markProvisioningStatePending?

pkg/services/baremetal/host/host.go
guettli: @janiskemper I fail to understand your goal. Please tell me your desired state. This is the current code:

```go 
		if err := sshClient.Reboot().Err; err != nil {
			err = fmt.Errorf("failed to reboot server (after install-image): %w", err)
			record.Warn(s.scope.HetznerBareMetalHost, "RebootFailed", err.Error())
			return actionError{err: err}
		}

		msg := "machine image and cloud-init data got installed (via image-url-command)"
		createSSHRebootEvent(ctx, s.scope.HetznerBareMetalHost, msg)
		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
			string(host.Spec.Status.ProvisioningState),
			"%s", msg)

		// clear potential errors - all done
		return actionComplete{}

```

What do you want this part to look like?

controllers/hetznerbaremetalhost_controller.go
guettli: @janiskemper I see no benefit of reading stale data from the cache. Waiting ~5ms is fine. Error messages which appear again and again because there was a conflict are confusing. What is your goal?

controllers/hetznerbaremetalhost_controller.go
guettli: @janiskemper the ResourceVersion is a string.

I updated the comment to make it more clear:

```go 
				// RV has changed like from "999" to "1000"

```

pkg/services/baremetal/host/host.go
guettli: ok, removed.

pkg/services/baremetal/host/host.go
guettli: @janiskemper no, this is not a new feature. I just added a comment, so that the reader understands why we do it.

pkg/services/baremetal/host/host.go
guettli: I think it is not overwritten:

This should overwrite it:

```go
func (hsm *hostStateMachine) ReconcileState(ctx context.Context) (actionRes actionResult) {
	initialState := hsm.host.Spec.Status.ProvisioningState
	defer func() {
		if hsm.nextState != initialState {
			hsm.log.V(1).Info("changing provisioning state", "old", initialState, "new", hsm.nextState)
			hsm.host.Spec.Status.ProvisioningState = hsm.nextState

			cond := conditions.Get(hsm.host, infrav1.ProvisionSucceededCondition)
			if cond != nil && cond.Reason == infrav1.StillProvisioningReason {
				markProvisionPending(hsm.host, hsm.nextState)
			}
		}
	}()
```


And this happens before, so the new code would ovewrite it:

```go 
func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
	markProvisionPending(s.scope.HetznerBareMetalHost, infrav1.StateImageInstalling)

```

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,138 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+
+	duration := time.Since(host.Spec.Status.LastUpdated.Time)
+	// Please keep the number (7) in sync with the docstring of ImageURL.
+	if duration > 7*time.Minute {
+		// timeout. Something has failed.
+		msg := fmt.Sprintf("ImageURLCommand timed out after %s. Deleting machine",
+			duration.Round(time.Second).String())
+		s.scope.Logger.Error(nil, msg, "logFile", logFile)
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			string(host.Spec.Status.ProvisioningState), clusterv1.ConditionSeverityWarning,
```
janiskemper: you use the provisioning state as reason here and in other cases. I don't think that this is a good pattern. We currently have already "StillProvisioning" as reason. We should IMO either use that or add new reasons. Here, in this case, I believe a new reason makes sense. "ImageURLCommandTimedOutReason".

There are multiple similar cases to this and I explicitly write it down here and would like to also include the other cases.

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,138 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+			"%s", msg)
+		return s.recordActionFailure(infrav1.FatalError, msg)
+	}
+
+	switch state {
+	case sshclient.ImageURLCommandStateRunning:
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
```
janiskemper: I think this is unnecessary and the normal "still provisioning" condition would be enough here. Since we set this automatically, I believe you can skip this condition altogether

pkg/services/baremetal/host/host.go
janiskemper: I see no point in setting this condition. We should stick to the main condition that is set automatically and just indicates "still provisioning"

(done, somehow github shows no way to reply here, Thomas)

pkg/services/baremetal/host/host.go
janiskemper: From what I understand, you add the function obtainHardwareDetailsStorage again here, even though we didn't do that before, because something could have changed in the meantime. This is independent of the topic of this PR from my understanding, because this would be something we would profit from in general. 

If I misunderstood this code, please explain me.

guettli: @janiskemper obtainHardwareDetailsStorage gets called twice. It depends on the way the machine gets provisioned: with image-url-command or without it. This is all part of the topic of this PR.


first (existing) usage of obtainHardwareDetailsStorage (in createAutoSetupInput (without image-url-command)): https://github.com/syself/cluster-api-provider-hetzner/blob/4cdbf1ebdf3afe9b95ab7ca1aa11d4cf8b51dc5d/pkg/services/baremetal/host/host.go#L1600-L1610

second (new) usage of obtainHardwareDetailsStorage (in actionImageInstallingCustomImageURLCommand) : https://github.com/syself/cluster-api-provider-hetzner/blob/4cdbf1ebdf3afe9b95ab7ca1aa11d4cf8b51dc5d/pkg/services/baremetal/host/host.go#L1287-L1297

pkg/services/baremetal/host/host.go
guettli: done: 25314799

controllers/hetznerbaremetalhost_controller.go
janiskemper: I'm talking about cmp Equal... And to avoid unnecessary API calls if we don't have to care about stale data. I think you didn't realize I was talking about cmp.Equal.

pkg/services/baremetal/host/host.go
janiskemper: I understood it now. I forgot about this feature that we added some time ago. Thanks!

controllers/hetznerbaremetalhost_controller.go
guettli: Currently, the code looks like this:

```go 
	// ----------------------------------------------------------------
	// Start: avoid conflict errors. Wait until local cache is up-to-date
	// Won't be needed once this was implemented:
	// https://github.com/kubernetes-sigs/controller-runtime/issues/3320
	initialHost := bmHost.DeepCopy()
	defer func() {
		if cmp.Equal(initialHost, bmHost) {
			// Nothing has changed. No need to wait.
			return
		}
```

The `cmp.Equal()` makes sense. When the host has not changed during the provisioning (in the final state), then we do not need to wait.

> I'm talking about cmp Equal... And to avoid unnecessary API calls if we don't have to care about stale data.

What API calls to you want to avoid? To robot-API or to api-server?

When dealing with third party APIs, it can always lead to strange and random errors when you read stale data, because the controller will do things twice (good case). Bad case: even more strange things happen. For example only Spec could be stale, and State nor, or vice-versa...

pkg/services/baremetal/host/host.go
guettli: @janiskemper yes, the condition gets overwritten. I removed it.

pkg/services/baremetal/host/host.go
guettli: done, no condition gets set here.

controllers/hetznerbaremetalhost_controller.go
janiskemper: I'm talking about api-server calls that come because we ignore the cache and directly call the Kubernetes api in this function. We can reduce them by ignoring some updates to the host object, because the stale cache might not be relevant for all form of updates.

We don't have to optimize that, but we should write a comment in this PR stating that this can be improved later by not use cmp.Equal but filtering for only those updates where we really want to avoid stale cache.

controllers/hetznerbaremetalhost_controller.go
janiskemper: Just checking with len whether the numbers are greater or smaller doesn't seem to work.. I checked with this small go program:

```
func main() {
	// comparing lengths of string literals
	cond1 := len("900") > len("899")
	cond2 := len("100") > len("99")

	if cond1 && cond2 {
		fmt.Println("Both conditions are true")
	} else {
		fmt.Printf("cond1=%v, cond2=%v\n", cond1, cond2)
	}
}

``` 

Am I missing something? Are you sure about this?

janiskemper: but tbh the two checks you have one after another, one with "len" one without, I actually don't understand it...

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,131 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		return actionError{err: fmt.Errorf("StateOfImageURLCommand failed: %w", err)}
+	}
+
+	duration := time.Since(host.Spec.Status.LastUpdated.Time)
+	// Please keep the number (7) in sync with the docstring of ImageURL.
+	if duration > 7*time.Minute {
+		// timeout. Something has failed.
+		msg := fmt.Sprintf("ImageURLCommand timed out after %s. Deleting machine",
+			duration.Round(time.Second).String())
+		s.scope.Logger.Error(nil, msg, "logFile", logFile)
```
janiskemper, line 1235: the nil logger is already part of the other PR's review process.

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,131 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			"ImageURLCommandTimedout", clusterv1.ConditionSeverityWarning,
```
janiskemper: Timedout -> TimedOut search and replace plz

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,131 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+			"ImageURLCommandFailed", clusterv1.ConditionSeverityWarning,
+			"%s", msg)
+		return s.recordActionFailure(infrav1.FatalError, msg)
+
+	case sshclient.ImageURLCommandStateNotStarted:
+		data, err := s.scope.GetRawBootstrapData(ctx)
+		if err != nil {
+			return actionError{err: fmt.Errorf("baremetal GetRawBootstrapData failed: %w", err)}
+		}
+		if s.scope.ImageURLCommand == "" {
```
janiskemper, line 1286: is this the first time we check it for bare metal? If so, then we should not return an error but "nil" because we should stop requeuing. I'm not sure if we have another similar case already, because we have to find a return value that makes sense and that let's us stop reconciliation.

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,131 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+				"stdoutStderr", stdoutStderr)
+			conditions.MarkFalse(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition,
+				"StartImageURLCommandFailed",
+				clusterv1.ConditionSeverityWarning,
+				"%s", msg)
+			return s.recordActionFailure(infrav1.ProvisioningError, msg)
+		}
+
+		conditions.MarkFalse(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition,
+			"ImageURLCommandStared",
```
janiskemper: typo

pkg/services/baremetal/host/host.go
```diff
@@ -1202,6 +1205,131 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+			clusterv1.ConditionSeverityInfo,
+			"baremetal-image-url-command started")
+
+		record.Event(s.scope.HetznerBareMetalHost, "ExecuteImageURLCommand",
```
janiskemper: I believe that you should create an event maybe in the cases where this fails, and maybe when it finishes, but why creating one when you start the command?!

pkg/services/baremetal/host/host_test.go
```diff
@@ -97,6 +99,152 @@ var _ = Describe("SetErrorMessage", func() {
...
+		svc.scope.HetznerBareMetalHost.Spec.RootDeviceHints = &infrav1.RootDeviceHints{
+			WWN: "eui.0025388801b4dff2",
+		}
+		res := svc.actionImageInstalling(ctx)
+		Expect(res).To(BeAssignableToTypeOf(actionFailed{}))
+		c := conditions.Get(host, infrav1.ProvisionSucceededCondition)
+		Expect(c.Message).To(ContainSubstring("StartImageURLCommand failed with non-zero exit status. Deleting machine"))
+	})
+
+	It("timeouts after 7 minutes", func() {
```
janiskemper: timeouts -> times out

controllers/hetznerbaremetalhost_controller.go
guettli: @janiskemper sooner or later controller-runtime will support "wait for update to be in the local cache".

https://github.com/kubernetes-sigs/controller-runtime/issues/3320

Then we do not need our own solution. There is already a link to that issue in the code.

Please tell me what text you want to put there.

controllers/hetznerbaremetalhost_controller.go
guettli: I extracted the comparison to utils.IsLocalCacheUpToDate() and wrote a unit-test. This showed that there was something wrong when the length changed. Is fixed in 1ab7ff87

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: @janiskemper I added that:

```go 
			// this can only be changed by updating the controller. This will make the
			// controller reconcile all resources.
			return actionContinue{delay: time.Hour}
```

ok?

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: I guess we had something like this before. But it does not matter. I removed that line.

pkg/services/baremetal/host/host_test.go
guettli: fixed.

controllers/hetznerbaremetalhost_controller.go
janiskemper:
```
We can potentially optimize this further by ensuring that the cache is up to date only in the cases where an outdated cache would lead to problems. Currently, we ensure that the cache is up to date in all cases, i.e. for all possible changes to the HetznerBareMetalHost object.
```

pkg/services/baremetal/host/host_test.go
janiskemper: Sorry, I believe that you didn't use the correct one.. "times out" not "time outs"

controllers/hetznerbaremetalhost_controller.go
guettli: great, I addded the comment.

pkg/services/baremetal/host/host_test.go
guettli: fixed.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -88,6 +92,58 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+		// The object changed. Wait until the new version is in the local cache
+
+		// Get the latest version from the apiserver.
+		apiserverHost := &infrav1.HetznerBareMetalHost{}
+
+		// Use uncached APIReader
+		err := r.APIReader.Get(ctx, client.ObjectKeyFromObject(bmHost), apiserverHost)
+		if err != nil {
+			reterr = errors.Join(reterr,
+				fmt.Errorf("failed get apiserverHost via uncached APIReader: %w", err))
```
janiskemper: hetznerBareMetalHost not apiserverHost ;)

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -88,6 +92,58 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+			return
+		}
+
+		apiserverRV := apiserverHost.ResourceVersion
+
+		err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, 3*time.Second, true, func(ctx context.Context) (done bool, err error) {
+			// new resource, read from local cache
+			latestFromLocalCache := &infrav1.HCloudMachine{}
```
janiskemper: This has to be HetznerBareMetalHost not HCloudMachine

pkg/services/baremetal/host/host.go
```diff
@@ -1215,6 +1218,130 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+		s.scope.Logger.Info("ImageURLCommandOutput", "logFile", logFile)
+
+		// Update name in robot API
+		if _, err := s.scope.RobotClient.SetBMServerName(s.scope.HetznerBareMetalHost.Spec.ServerID, s.scope.Hostname()); err != nil {
+			record.Warn(s.scope.HetznerBareMetalHost, "SetBMServerNameFailed", err.Error())
+			s.handleRobotRateLimitExceeded(err, "SetBMServerName")
+			return actionError{err: fmt.Errorf("failed to update name of host in robot API: %w", err)}
+		}
+
+		if err := sshClient.Reboot().Err; err != nil {
```
janiskemper, line 1258: can you add a comment that you reboot here? It's quite hard to read because it is all in one line.

pkg/services/baremetal/host/host.go
```diff
@@ -1215,6 +1218,130 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
+	case sshclient.ImageURLCommandStateFailed:
+		record.Warn(s.scope.HetznerBareMetalHost, "InstallImageNotSuccessful", logFile)
+		msg := "image-url-command failed"
+		s.scope.Logger.Error(nil, msg, "logFile", logFile)
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			"ImageURLCommandFailed", clusterv1.ConditionSeverityWarning,
+			"%s", msg)
+		return s.recordActionFailure(infrav1.FatalError, msg)
+
+	case sshclient.ImageURLCommandStateNotStarted:
```
janiskemper, line 1280: more empty lines here in this code block plz

controllers/hetznerbaremetalhost_controller.go
guettli: fixed.

controllers/hetznerbaremetalhost_controller.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: done.

PR #1686, opened 2025-09-25
:seedling: Add `--baremetal-ssh-after-install-image=false` to controller

guettli:
Currently, it defaults to `true`, so that the old provisioning happens.

If you do not want the caph controller to ssh to the baremetal machines after `install-image`, then set this argument to `false`.

The baremetal ProvisioningState "ensure-provisioned" will be skipped if set to `false`. Skipping that is perfectly fine, because MachineHealthChecks are responsible for checking the health of a Node. If it is not healthy Remediation will be triggered. The step "ensure-provisioned" provides no real benefit.

BTW, remediation uses `addRebootAnnotation(host.Annotations)` not ssh or robot-api to reboot a machine.

## WorkloadClusterClientFactory

For making the access to the wl-cluster testable the interface `WorkloadClusterClientFactory` was introduced.

## Manual Testing

I used a machine which uses ssh-port 100 after install-image. I temporarily added a panic, so that
I would see, if ssh would get used:

```go
func (f *sshFactory) NewClient(in Input) Client {
	if in.Port == 100 {
		panic("todo todo todo todo todo todo just testing. should not happen.")
	}
	return &sshClient{
		privateSSHKey: in.PrivateKey,
		ip:            in.IP,
		port:          in.Port,
	}
}
```

Provisioning with `--baremetal-ssh-after-install-image=false` worked.


api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -310,6 +311,22 @@ type ControllerGeneratedStatus struct {
 	// Conditions define the current service state of the HetznerBareMetalHost.
 	// +optional
 	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
+
+	// ExternalIDs contains values of external systems.
```
janiskemper: of -> from here and below

main.go
```diff
@@ -107,6 +108,8 @@ func main() {
...
 	fs.StringVar(&preProvisionCommand, "pre-provision-command", "", "Command to run (in rescue-system) before installing the image on bare metal servers. You can use that to check if the machine is healthy before installing the image. If the exit value is non-zero, the machine is considered unhealthy. This command must be accessible by the controller pod. You can use an initContainer to copy the command to a shared emptyDir.")
 	fs.BoolVar(&skipWebhooks, "skip-webhooks", false, "Skip setting up of webhooks. Together with --leader-elect=false, you can use `go run main.go` to run CAPH in a cluster connected via KUBECONFIG. You should scale down the caph deployment to 0 before doing that. This is only for testing!")
+	fs.BoolVar(&sshAfterInstallImage, "baremetal-ssh-after-install-image", true, "Connect to the baremetal machine after install-image and ensure it is provisioned. Current default is true, but we might change that to false. Background: If company A maintains the management-cluster for company B, then B might not want that a controller of A is able to connect to the servers.")
```
janiskemper: we can formulate that in a more neutral way. "Background: Users might not want the controller to be able to ssh onto the servers"

pkg/scope/workloadclusterclient.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, fmt.Errorf("failed to acquire secret: %w", err)
+	}
+	kubeconfigBytes, ok := kubeconfigSecret.Data[secret.KubeconfigDataName]
+	if !ok {
+		return nil, fmt.Errorf("missing key %q in secret data (WorkloadClientConfigFromKubeconfigSecret)", secret.KubeconfigDataName)
+	}
+	return clientcmd.NewClientConfigFromBytes(kubeconfigBytes)
+}
+
+// WorkloadClusterClientFactory is interface to get a new controller-runtime Client to access a
```
janiskemper: *an interface

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+	if host.Spec.Status.ExternalIDs.RebootAnnotationSince.IsZero() {
+		host.Spec.Status.ExternalIDs.RebootAnnotationSince = metav1.Now()
+	}
 
-			wantHostName := s.scope.Hostname()
+	wlClient, err := s.scope.WorkloadClusterClientFactory.NewWorkloadClient(ctx)
+	if err != nil {
+		err = fmt.Errorf("actionProvisioned (Reboot via Annotation), failed to get wlClient: %w", err)
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
+			infrav1.StillProvisioningReason,
```
janiskemper: this is the wrong reason. We have provisioned the server successfully. Right now, we have a problem because the server is supposed to be rebooted, which means usually that the MHC triggered.

We should probably even think about another condition. What would you propose?

This applies here and for all conditions below

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
 
-			if trimLineBreak(out.StdOut) == wantHostName {
-				// Reboot has been successful
-				s.scope.HetznerBareMetalHost.Spec.Status.Rebooted = false
-				s.scope.HetznerBareMetalHost.ClearRebootAnnotations()
+	machine, err := util.GetOwnerMachine(ctx, s.scope.Client, s.scope.HetznerBareMetalMachine.ObjectMeta)
+	if err != nil {
+		err = fmt.Errorf("actionProvisioned (Reboot via Annotation), GetOwnerMachine failed: %w",
+			err)
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
```
janiskemper: We can skip this condition. We also don't set a condition if another object is not found (e.g. HetznerBareMetalMachine)

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
-				if errors.Is(err, errUnexpectedHostName) {
-					// One possible reason: The machine gets used by a second wl-cluster
-					record.Warnf(s.scope.HetznerBareMetalHost, "UnexpectedHostName",
-						"Provisioned: wanted %q. %s", wantHostName, err.Error())
-				}
-				return actionError{err: fmt.Errorf("failed to handle incomplete boot - actionProvisioned: %w", err)}
+	if machine.Status.NodeRef == nil {
+		// Very unlikely, but we want to avoid a panic.
+		err = errors.New("machine.Status.NodeRef is nil?")
+		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
```
janiskemper: let's not worry about this condition if this won't happen. I would remove it

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+			clusterv1.ConditionSeverityWarning, "%s",
+			err.Error())
+		return actionError{err: err}
+	}
+
+	isRebooted := host.Spec.Status.Rebooted
+	if !isRebooted {
+		// Reboot now
+
+		// Set current BootID, so we can detect a succesfull reboot
```
janiskemper: typo -> successful

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+		host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID = currentBootID
+
+		if s.scope.SSHAfterInstallImage {
```
janiskemper, line 1910: add some empty lines plz

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+			infrav1.StillProvisioningReason,
+			clusterv1.ConditionSeverityInfo, "%s",
+			msg)
+		host.Spec.Status.Rebooted = true
+		return actionContinue{delay: 10 * time.Second}
+	}
+
+	// Reboot has been done already. Check whether it has been successful
+
+	if host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID == "" {
```
janiskemper: I believe that you would create a loop with this check and returning an error.

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+		host.Spec.Status.ExternalIDs.RebootAnnotationSince.Time = time.Time{}
+
+		host.ClearRebootAnnotations()
+
+		host.ClearError()
+		conditions.MarkTrue(host, infrav1.ProvisionSucceededCondition)
+		return actionComplete{}
+	}
+
+	s.scope.Logger.Info("foooo", "long", host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID,
```
janiskemper: remove this

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+			clusterv1.ConditionSeverityWarning, "%s",
+			err.Error())
+		return actionError{err: err}
+	}
 
-			if trimLineBreak(out.StdOut) == wantHostName {
-				// Reboot has been successful
-				s.scope.HetznerBareMetalHost.Spec.Status.Rebooted = false
-				s.scope.HetznerBareMetalHost.ClearRebootAnnotations()
+	machine, err := util.GetOwnerMachine(ctx, s.scope.Client, s.scope.HetznerBareMetalMachine.ObjectMeta)
```
janiskemper, line 1865: plz explain here what you want to do here. That's not clear right now. Write down in a comment the whole plan of using the boot ID of the node to check if they changed, which indicates a reboot.

pkg/services/baremetal/host/host.go
```diff
@@ -1789,65 +1804,222 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+		conditions.MarkTrue(host, infrav1.ProvisionSucceededCondition)
+		return actionComplete{}
+	}
+
+	s.scope.Logger.Info("foooo", "long", host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID,
+		"cur", currentBootID)
+
+	if !s.scope.SSHAfterInstallImage {
+		// s.scope.SSHAfterInstallImage is false: No ssh allowed.
+		// We can only wait for the BootID in the wl-cluster to change.
```
janiskemper, line 1976: We should include a timeout

pkg/services/baremetal/host/host_test.go
```diff
@@ -1652,3 +1656,60 @@ var _ = Describe("actionProvisioned", func() {
 		}),
 	)
 })
+
+var _ = Describe("actionProvisioned SSHAfterInstallImage=false", func() {
+	It("test reboot annotation for SSHAfterInstallImage=false", func() {
```
janiskemper: this formulation is not how it is intended. What I can suggest, is to split this up to a BeforeEach and smaller "It" functions that each have a specific purpose, e.g. "It reboots the server via hardware reboot".

janiskemper (changes requested):

api/v1beta1/hetznerbaremetalhost_types.go
guettli: fixed.

pkg/scope/workloadclusterclient.go
guettli: fixed.

main.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: I added `HostHasNoRebootAnnotation`. It is not nice to express this, because the term must match to ...=false.

If you have a better idea, please tell me.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: please tell me your desired implementation. I am unsure what you want.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: I added that timeout at the top of the func:

```go 
	rebootDuration := time.Since(host.Spec.Status.ExternalIDs.RebootAnnotationSince.Time)
	if rebootDuration > 5*time.Minute {
		msg := fmt.Sprintf("Rebooting timed out after: %s", rebootDuration.Round(time.Second))
		s.scope.Logger.Info(msg)
		conditions.MarkFalse(
			s.scope.HetznerBareMetalHost,
			infrav1.HostHasNoRebootAnnotationCondition,
			"TimedOut",
			"%s",
			msg,
		)
		return s.recordActionFailure(infrav1.FatalError, msg)
	}
```

pkg/services/baremetal/host/host.go
janiskemper: I would say we shouldn't go into a loop. If you don't know how to solve that, I can check. 

One note: I believe that we don't need this check in the first place and could remove it

guettli: I removed the extra check.

pkg/services/baremetal/host/host_test.go
guettli: @janiskemper I created three `It()` functions. BeforeEach was not needed. Each has a slightly different setup.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: done.

pkg/services/baremetal/host/host.go
guettli: I would like to avoid a panic, even if it is unlikely that this field is nil. A panic will bring the whole controller down.

pkg/services/baremetal/host/host.go
```diff
@@ -1894,21 +1909,24 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
 			}
+
 			sshClient := s.scope.SSHClientFactory.NewClient(in)
+
 			out := sshClient.Reboot()
 			if err := handleSSHError(out); err != nil {
-				conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
-					infrav1.StillProvisioningReason,
+				conditions.MarkFalse(host, infrav1.HostHasNoRebootAnnotationCondition,
+					"SSHAfterInstallImageFailed",
```
janiskemper: This reason seems to not fit

pkg/services/baremetal/host/host.go
```diff
@@ -1917,35 +1935,28 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
-	if host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID == "" {
-		err := errors.New("internal error host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID not set")
-		s.scope.Logger.Error(err, "")
-		conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
-			infrav1.StillProvisioningReason,
-			clusterv1.ConditionSeverityWarning, "%s",
-			err.Error())
-		return actionError{err: err}
-	}
+	// Reboot has been done already. Check whether it has been successful We connet to the
```
janiskemper: typo and grammar

pkg/services/baremetal/host/host.go
```diff
@@ -2008,30 +2016,32 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
 			clusterv1.ConditionSeverityWarning, "%s",
 			err.Error())
 		return actionError{err: err}
 	}
-	conditions.MarkFalse(host, infrav1.ProvisionSucceededCondition,
-		infrav1.StillProvisioningReason,
+
+	conditions.MarkFalse(host, infrav1.HostHasNoRebootAnnotationCondition,
+		"TriggeredRebootViaSSH",
 		clusterv1.ConditionSeverityWarning,
```
janiskemper: I think you can make this an "info".

pkg/services/baremetal/host/host.go
```diff
@@ -1957,18 +1968,15 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 		host.ClearRebootAnnotations()
 
 		host.ClearError()
-		conditions.MarkTrue(host, infrav1.ProvisionSucceededCondition)
+		conditions.MarkTrue(host, infrav1.HostHasNoRebootAnnotationCondition)
```
janiskemper: You have to mark this true also in the case where you check the hostname via ssh

pkg/services/baremetal/host/host.go
```diff
@@ -1805,65 +1820,232 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+		return actionComplete{}
+	}
+
+	if !s.scope.SSHAfterInstallImage {
+		// s.scope.SSHAfterInstallImage is false: No ssh allowed.
+		// We can only wait for the BootID in the wl-cluster to change.
+		conditions.MarkFalse(host, infrav1.HostHasNoRebootAnnotationCondition,
+			"WaitingForNodeToBeRebooted",
+			clusterv1.ConditionSeverityWarning,
```
janiskemper: this should be "info" IMO

pkg/services/baremetal/host/host.go
janiskemper: I propose "RebootSucceeded" as condition @guettli and we should set this only after a successful reboot was performed - not before. With that I mean that if the reboot annotation was never set, then this condition should also not be set.

pkg/services/baremetal/host/host.go
janiskemper: I'm only talking about the condition, not the error.

pkg/services/baremetal/host/host.go
guettli: use RebootViaSSHFailed now

pkg/services/baremetal/host/host.go
guettli: new:
```go 
	// Reboot has already been performed. Now, verify its success by connecting to the wl-cluster
	// and checking the BootID. If the BootID has changed, the reboot was successful.
```

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: ok, it is named RebootSucceededCondition now.

pkg/services/baremetal/host/host.go
guettli: ok, I removed the condition.

api/v1beta1/conditions_const.go
```diff
@@ -233,3 +233,8 @@ const (
 	// DeprecatedRateLimitExceededCondition reports whether the rate limit has been reached.
 	DeprecatedRateLimitExceededCondition clusterv1.ConditionType = "RateLimitExceeded"
 )
+
+const (
+	// RebootSucceededCondition indicates that the machine gets rebooted.
```
janiskemper: "... that the machine got rebooted successfully"

pkg/services/baremetal/host/host.go
```diff
@@ -1805,65 +1820,230 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+				return actionError{err: err}
 			}
-			return actionContinue{delay: 10 * time.Second}
-		}
-		// Reboot now
-		out := sshClient.Reboot()
-		if err := handleSSHError(out); err != nil {
-			return actionError{err: err}
 		}
+		msg := fmt.Sprintf("Rebooting because annotation was set. Old BootID: %s", currentBootID)
```
janiskemper, line 1944: some empty lines would be good in this section

pkg/services/baremetal/host/host.go
```diff
@@ -1805,65 +1820,230 @@ func analyzeSSHOutputProvisioned(out sshclient.Output) (isTimeout, isConnectionR
...
+		err = fmt.Errorf(errMsgFailedHandlingIncompleteBoot, err)
+		conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
+			"RebootFailed",
+			clusterv1.ConditionSeverityWarning, "%s",
+			err.Error())
+		return actionError{err: err}
+	}
+
+	conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
+		"TriggeredRebootViaSSH",
```
janiskemper: IMO this should be the same condition as above:
```
conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
			"WaitingForNodeToBeRebooted",
			"Waiting for BootID of Node (in wl-cluster) to change (%s)",
			time.Since(host.Spec.Status.ExternalIDs.RebootAnnotationSince.Time).Round(time.Second))
```

api/v1beta1/conditions_const.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: fixed.

pkg/services/baremetal/host/host.go
guettli: done.

api/v1beta1/conditions_const.go
```diff
@@ -235,6 +235,6 @@ const (
 )
 
 const (
-	// RebootSucceededCondition indicates that the machine gets rebooted.
+	// RebootSucceededCondition indicatesthat the machine got rebooted successfully.
```
janiskemper: typo

guettli: fixed.

PR #1693, opened 2025-10-16
:seedling: HCloudMachineType: allow all values.

guettli:
The list of valid machine types gets changed by
Hetzner from time to time. CAPH no longer
validates this string. It is up to you to use a
valid type. Not all types are available in all
locations.

Recent change: https://docs.hetzner.cloud/changelog#2025-10-16-server-types-deprecated

---

Additionally 1:

Show underlaying error in Condition, when a server can't be created.

Example:

The condition only showed:

> "failed to create HCloud server tcs-guettli-x39-1-33-v2-md-arm-987vb-p69g9-grgqc"

now the error is included:

> failed to create HCloud server tcs-guettli-imi-1-33-v2-md-arm-fmzb4-fv42w-7dbs4: error during placement (resource_unavailable, 3e96a3ff14675f0144febd731d8ad72a)'

---

Additionally 2:

When creating a machine in hcloud failed (for example a typo in the machine type, or resource_unavailable), then remediation failed up to now. This was fixed.

--- 

Additionally 3:

The fix in the Remedation revealed that some tests were broken. These got fixed.

### DIFF
--- a/pkg/services/hcloud/remediation/remediation.go
+++ b/pkg/services/hcloud/remediation/remediation.go
@@ -56,7 +56,8 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		}
 	}
 
-	// stop remediation if server does not exist or ProviderID is nil.
+	// stop remediation if server does not exist or ProviderID is nil (in this case the server
+	// cannot exist).
 	if server == nil {
 		s.scope.HCloudRemediation.Status.Phase = infrav1.PhaseDeleting
 


pkg/services/hcloud/remediation/remediation.go
```diff
@@ -49,6 +50,9 @@ func NewService(scope *scope.HCloudRemediationScope) *Service {
 // Reconcile implements reconcilement of HCloudRemediation.
 func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err error) {
 	server, err := s.findServer(ctx)
+	if errors.Is(err, hcloudutil.ErrNilProviderID) {
```
janiskemper: shouldn't we stop the remediation in this case rather? Like we do below with "phase deleting"?

guettli: fixed, we check for nil ProviderID before calling findServer now.

PR #1703, opened 2025-11-03
:sparkles: Step 2: support upstream hcloud-ccm secret/key conventions

guettli:
# New PR:

[:seedling: Step 2: support upstream hcloud-ccm secret/key conventions (II) by guettli · Pull Request #1920 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1920)

# Old:

Previous PR: [:seedling: Support upstream providerId format (Step1) #1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850)

Next PR: [:seedling: Set v1.1.x to main branch, and release. #1781](https://github.com/syself/cluster-api-provider-hetzner/pull/1781)

This PR is step 2 of the work to support the upstream [hcloud CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager/) for bare-metal clusters.

The core ProviderID migration work was moved to [#1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850). The refactor-only helper rename now lives in [#1914](https://github.com/syself/cluster-api-provider-hetzner/pull/1914).

This PR contains the follow-up changes that make the upstream CCM easier to use in practice.

# Changes

- mirror the HCloud token in the workload-cluster secret under both `token` and `hcloud`, in addition to the configured key, so both CCM variants can read it
- if `spec.hetznerSecret.name` is not `hcloud`, create an additional workload-cluster secret named `hcloud` for upstream compatibility
- update API comments and generated CRD docs to describe the upstream-compatible secret naming/key behavior
- keep recommending upstream defaults such as secret name `hcloud`, token key `token`, and Robot keys `robot-user` / `robot-password`
- add a `note` entry to the workload-cluster secret to make it obvious that CAPH reconciles it


controllers/hetznercluster_controller.go
```diff
@@ -500,73 +505,101 @@ func reconcileTargetSecret(ctx context.Context, clusterScope *scope.ClusterScope
...
+		if wlSecret.Data == nil {
+			wlSecret.Data = make(map[string][]byte)
 		}
 
-		secret.Data[clusterScope.HetznerCluster.Spec.HetznerSecret.Key.HCloudToken] = hetznerToken
+		wlSecret.Data[clusterScope.HetznerCluster.Spec.HetznerSecret.Key.HCloudToken] = hcloudToken
+
+		// upstream hcloud-ccm uses the secret key "token", while the old Syself ccm used "hcloud".
+		// For compatibilty, we create always the other key, too.
+		for _, key := range []string{"token", "hcloud"} {
```
janiskemper, line 595: right now we create a duplicate right? Wouldn't it be enough to create the right key (e.g. "token" for the upstream ccm)?

controllers/hetznercluster_controller.go
```diff
@@ -500,73 +505,101 @@ func reconcileTargetSecret(ctx context.Context, clusterScope *scope.ClusterScope
...
-		if secret.Data == nil {
-			secret.Data = make(map[string][]byte)
+		if wlSecret.Data == nil {
+			wlSecret.Data = make(map[string][]byte)
 		}
 
-		secret.Data[clusterScope.HetznerCluster.Spec.HetznerSecret.Key.HCloudToken] = hetznerToken
+		wlSecret.Data[clusterScope.HetznerCluster.Spec.HetznerSecret.Key.HCloudToken] = hcloudToken
+
+		// upstream hcloud-ccm uses the secret key "token", while the old Syself ccm used "hcloud".
```
janiskemper: can you specify what "used" means? This is configured in the Helm charts right?

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -55,8 +55,7 @@ import (
 // TODO: Implement logic for removal of unpaid servers.
 
 const (
-	// providerIDPrefix is a prefix for ProviderID.
-	providerIDPrefix = "hcloud://"
+	providerIDPrefix = "hrobot://"
```
janiskemper: you cannot just replace it here, can you? All existing servers would use the old prefix...

controllers/hetznercluster_controller.go
guettli: The result should work for both CCMs.

By creating both keys we are safe. Users can try to upstream hcloud-ccm. If it fails, they can easily switch back to the old syself-ccm.

Why not create both keys?

controllers/hetznercluster_controller.go
guettli: The current Syself ccm uses the fixed key "hcloud". If you use automatic key-reloading, then you even can't configure that to be "token". This PR improves that, so that "token" gets read, too. But this is not merged and not deployed yet.

Please tell me, if you think this could be phrased differently.

[:seedling: Support hrobot:// ProviderID of upstream ccm. by guettli · Pull Request #72 · syself/hetzner-cloud-controller-manager](https://github.com/syself/hetzner-cloud-controller-manager/pull/72)

pkg/services/baremetal/baremetal/baremetal.go
guettli: Yes, this is safe. This constant is only used at one place. Look at `setProviderID`.

Wait, I have an idea how to make this more obvious. I remove that constant. It is only used at one place. No need for a constant.

Done: b3759505

Now it is obvious. This is only used at one place, and only if the machine has no ProviderID yet.

controllers/hetznercluster_controller.go
janiskemper: you create both keys in both secrets. But you actually need only one key per secret right?

controllers/hetznercluster_controller.go
janiskemper: what do you suggest to write so that this is clear? And that people don't have to think about whether this means that it is hardcoded, or that it is the default configuration?

pkg/services/baremetal/baremetal/baremetal.go
janiskemper: okay...

controllers/hetznercluster_controller.go
guettli: yes

controllers/hetznercluster_controller.go
guettli: I think that the current comment is fine (unchanged).

```go
		// upstream hcloud-ccm uses the secret key "token", while the old Syself ccm used "hcloud".
		// For compatibilty, we create always the other key, too.
```

Please tell me what text you want, if you want a different text.

controllers/hetznercluster_controller.go
janiskemper: IMO it is less confusing for users if we have only the correct key in the secret. That's why I would implement it in that way

controllers/hetznercluster_controller.go
janiskemper: If I understand correctly, then "token" is the default for a configurable variable for the hcloud-ccm. If that is correct, then I would specify that here. Right now it sounds as if it is hardcoded

guettli: done.

New text:

```go 
		// upstream hcloud-ccm uses by default the secret key "token", while the old Syself ccm used
		// "hcloud". For compatibility, we create always the other key, too.
```

controllers/hetznercluster_controller.go
guettli: I have concerns. If someone just changes the name of the secret, but does not change the key-mapping, then things won't work.

But I changed something. The code and docs are now explicit to suggest the defaults of upstream hcloud-ccm.

If the settings are different, the created secret contains a corresponding note.

ok like this?

janiskemper: if it is configurable though - don't we have to make it configurable in CAPH as well?

guettli: @janiskemper both values are configurable in caph: the name of the secret in the wl-cluster, and the name of the key in the secret.

Or do you mean: Make it configurable to create the second secret/key? I would like to avoid that. But if it is important, I can add it.

janiskemper: I'm actually just referring to your concern that you wrote about before and try to understand it. Can you explain, if I misunderstood?

guettli: > I have concerns. If someone just changes the name of the secret, but does not change the key-mapping, then things won't work.

If someone used the Syself ccm in the past, and then reads the [migration docs](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/main/docs/guides/robot/migrating-from-syself.md), then he might forget to update the `HetznerCluster.Spec.HetznerSecret.Key.HCloudToken` in his cluster.

I think the current implementation is a good compromise.

janiskemper: So what you currently do is storing the same hcloud token three times in the same secret, right? 
1. using the configurable key from HetznerCluster.Spec.HetznerSecret.Key.HCloudToken
2. hardcoded key "token"
3. hardcoded key "hcloud".

Right? 

I think this will confuse people. We should check what is really necessary and then make the documentation clear. Otherwise we will have to start documenting why we have the same thing thrice in the secret.

guettli: Suggestion: We deprecate `HetznerCluster.Spec.HetznerSecret.Key.HCloudToken`. In the docstring we explain that the controller will currently create "token", "hcloud" and the specified value.

In the long run (maybe caph version released at the end of 2027) we will remove that field and will only create "token".

I think it is perfectly fine if that value can't be configured.

What do you think about that?

janiskemper: I suppose so. Can you discuss this with @batistein?

guettli: @janiskemper I talked to Sven. He is fine with creating three keys in the secret. But he does not want to deprecate this setting.

janiskemper: okay can you then update the comments in the function to make that clear that these three keys are created for the hcloud token?
EDIT: And this should be also documented I suppose

guettli: @janiskemper I added a comment to the function and to the related field in the struct:

623f05aa

ok now?

api/v1beta1/types.go
```diff
@@ -122,8 +122,11 @@ type HetznerSecretRef struct {
 // HetznerSecretKeyRef defines the key name of the HetznerSecret.
 // Need to specify either HCloudToken or both HetznerRobotUser and HetznerRobotPassword.
 type HetznerSecretKeyRef struct {
-	// HCloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
-	// We recommend to use "token", because this is the default of upstream hcloud-ccm.
+	// hcloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
```
janiskemper: capital "HCloudToken"

api/v1beta1/types.go
```diff
@@ -122,8 +122,11 @@ type HetznerSecretRef struct {
...
+	// We recommend to use "token", because this is the default of upstream hcloud-ccm, while the
+	// legacy Syself ccm uses "hcloud". For maximal compatibility up to three keys get created in the
```
janiskemper: let's write "Syself fork" instead of "Syself ccm"

api/v1beta1/types.go
guettli: done, although we should switch to the recommendation: Use the string which is used in the json-tag. Because the end-user does not know about the Go code. The end users see the OpenAPI schema.

api/v1beta1/types.go
guettli: @janiskemper code uses "Syself ccm fork" now. ok?

api/v1beta1/types.go
```diff
@@ -101,9 +101,19 @@ type HCloudPlacementGroupStatus struct {
 
 // HetznerSecretRef defines all the names of the secret and the relevant keys needed to access Hetzner API.
 type HetznerSecretRef struct {
-	// Name defines the name of the secret.
-	// +kubebuilder:default=hetzner
+	// Name defines the name of the secret. The name gets used for reading the credential in the
+	// mgt-cluster, and it gets used for creating a secret in the wl-cluster. About the secret in
+	// the wl-cluster: Attention, the upstream hcloud-ccm helm chart expects the name to be
```
janiskemper, line 106: note for myself: maybe we can get a PR in the upstream CCM to support "hetzner"

api/v1beta1/types.go
```diff
@@ -113,18 +123,32 @@ type HetznerSecretRef struct {
 // Need to specify either HCloudToken or both HetznerRobotUser and HetznerRobotPassword.
 type HetznerSecretKeyRef struct {
 	// HCloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
+	// We recommend to use "token", because this is the default of upstream hcloud-ccm, while the
```
janiskemper: I'm not sure how this fits together with the current default "hcloud-token". 

I expect that most people use the standard configuration, so that I'm not sure about changing the defaults here.

api/v1beta1/types.go
```diff
@@ -113,18 +123,32 @@ type HetznerSecretRef struct {
...
+	// secret for HCLOUD_TOKEN: "hcloud", "token" and the value of hcloudToken. This way we ensure
+	// that the ccm in the wl-cluster finds the secret.
+	//
 	// +optional
-	// +kubebuilder:default=hcloud-token
+	// +kubebuilder:default=token
 	HCloudToken string `json:"hcloudToken"`
-	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is stored.
+
+	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is
```
janiskemper, line 137: Also here I'm not sure about changing the default. 

We probably need to change a lot of our own templates to make this work

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -903,3 +897,41 @@ func analyzePatchError(err error, ignoreNotFound bool) error {
...
+	// prefixRobotLegacy is the prefix used by the Syself ccm.
+	prefixRobotLegacy = "hcloud://bm-"
+
+	// prefixRobotNew is the prefix used by the HCloud ccm.
+	prefixRobotNew = "hrobot://"
+
+	// UseHrobotProviderIdForBaremetalAnnotation on a HetznerCluster defines the ProviderID format
+	// for baremetal nodes. Only "hcloud://bm-" and "hrobot://" are allowed values. If not set, then
+	// the old format ("hcloud://bm-") gets used.
+	UseHrobotProviderIdForBaremetalAnnotation = "capi.syself.com/use-hrobot-provider-id-for-baremetal"
```
janiskemper: annotations are usually defined in the api folder AFAIK

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -903,3 +897,41 @@ func analyzePatchError(err error, ignoreNotFound bool) error {
...
+
+// GenerateProviderID returns the providerID for the given machine. If the ProviderID already
+// exists, then this will be returned. If the ProviderID is empty, it uses the old format by default
+// (hcloud://bm-NNNN), except the Annotation `capi.syself.com/use-hrobot-provider-id-for-baremetal`
+// on the hetznerCluster is set to `hrobot://`.
+func GenerateProviderID(machine *clusterv1.Machine, hetznerCluster *infrav1.HetznerCluster, serverNumber int) (string, error) {
+	if machine.Spec.ProviderID != nil && *machine.Spec.ProviderID != "" {
+		return *machine.Spec.ProviderID, nil
+	}
+	prefix, ok := hetznerCluster.Annotations[UseHrobotProviderIdForBaremetalAnnotation]
```
janiskemper: why do we check for a prefix? I assume that, with this name, the annotation should be set on "true" or "false" probably. So I would check:
if set and value is "true" then use hrobot, otherwise use old version.

pkg/services/baremetal/baremetal/baremetal.go
janiskemper: this annotation has to be documented. Currently, we have zero documentation on how to use the old and new CCM etc.

api/v1beta1/types.go
guettli: @janiskemper the code creates both. This should be fine.

api/v1beta1/types.go
guettli: @janiskemper same here. Both get currently created. For now this should not matter.

pkg/services/baremetal/baremetal/baremetal.go
guettli: @janiskemper yes, this makes sense. Done.

pkg/services/baremetal/baremetal/baremetal.go
guettli: ok, changed to a boolean.

pkg/services/baremetal/baremetal/baremetal.go
guettli: done in: docs/caph/03-reference/08-annotations.md

docs/caph/02-topics/05-baremetal/03-creating-workload-cluster.md
```diff
@@ -71,8 +71,34 @@ This requires a secret containing access credentials to both Hetzner Robot and H
 
 {% /callout %}
 
-If you have configured your secret correctly in the previous step then you already have the secret in your cluster.
-Let's deploy the hetzner CCM helm chart.
+First you need to decide if you want to use the Syself CCM or the upstream HCloud CCM.
```
janiskemper: I think this needs to go partially to the beginning. If people decide to use the upstream CCM, then they already created their cluster without the annotation, which means that they cannot apply the new CCM without the cluster failing. 

We need to make people aware, then add the information that the annotation is needed in this case to the step where the HetznerCluster is applied, and then here refer to the two CCMs again

test/e2e/Makefile
```diff
@@ -75,6 +75,7 @@ e2e-ccm-templates:
 	--set env.HCLOUD_TOKEN.valueFrom.secretKeyRef.name=hetzner \
 	--set env.HCLOUD_TOKEN.valueFrom.secretKeyRef.key=hcloud \
 	--set networking.enabled=false > $(REPO_ROOT)/test/e2e/data/ccm/hcloud-ccm.yaml
+	sed -i '1s/^/# Created by e2e-ccm-templates. DO NOT EDIT.\n/' $(REPO_ROOT)/test/e2e/data/ccm/hcloud-ccm.yaml
```
janiskemper: this change seems unrelated. can you open a separate PR for it or add it to another PR?

docs/caph/02-topics/05-baremetal/03-creating-workload-cluster.md
guettli: @janiskemper I moved the text up, but I realized that things are easier, when we first release our new ccm, and then write the docs.

After [:seedling: Support hrobot:// ProviderID of upstream ccm. by guettli · Pull Request #72 · syself/hetzner-cloud-controller-manager](https://github.com/syself/hetzner-cloud-controller-manager/pull/72) is merged:

> Update this paragraph, use new format (hrobot://) by setting the annotation.

test/e2e/Makefile
guettli: removed it.

docs/caph/02-topics/05-baremetal/03-creating-workload-cluster.md
```diff
@@ -13,6 +13,41 @@ Secrets as of now are hardcoded given we are using a flavor which is essentially
...
+  machine (in mgt-cluster) is which Node (in wl-cluster).
+- Creates LoadBalancers
+
+If you are unsure, use the HCloud CCM. In the long run we (Syself) want to switch from our fork to
+the upstream CCM.
+
+The CAPH controller creates the required secrets in the workload cluster for you. You only need to
+install the CCM. To allow switching the ccm, the controller creates the secret "hetzner" and
+"hcloud" in the workload-cluster. These secrets contain the HCLOUD_TOKEN, so that the ccm can
+connect to the HCloud API.
```
janiskemper: I would write the following:

The CCM calls the Hetzner APIs. To authenticate, it reads the credentials from a secret. This secret has to be in the workload cluster, when the CCM runs in the workload cluster. CAPH creates the secret and syncs the credentials specified in the management cluster to the workload cluster. The Syself fork and the HCloud CCM use a different naming pattern. That's why CAPH generates two secrets named "hetzner" and "hcloud".

guettli: @janiskemper yes, that's better. I updated it.

api/v1beta1/types.go
```diff
@@ -101,9 +101,19 @@ type HCloudPlacementGroupStatus struct {
...
+	// mgt-cluster, and it gets used for creating a secret in the wl-cluster. About the secret in
+	// the wl-cluster: Attention, the upstream hcloud-ccm helm chart expects the name to be
+	// "hcloud". The Syself ccm fork defaults to "hetzner". For compatibility with upstream hcloud-ccm
+	// the controller creates two secrets, if the name is different from "hcloud" (one with name
+	// "hcloud", one with name being the value of this setting). The secret will be created in the
+	// namespace "kube-system" of the workload-cluster. We recommend to use "hcloud", because this is
+	// the default of upstream hcloud-ccm. Set `spec.skipCreatingHetznerSecretInWorkloadCluster`, if
+	// you don't want that secret in the wl-cluster to be created.
+	//
+	// +kubebuilder:default=hcloud
```
janiskemper: should we change this default? This is potentially breaking if the current default is used by someone already.

api/v1beta1/types.go
```diff
@@ -101,9 +101,19 @@ type HCloudPlacementGroupStatus struct {
 
 // HetznerSecretRef defines all the names of the secret and the relevant keys needed to access Hetzner API.
 type HetznerSecretRef struct {
-	// Name defines the name of the secret.
-	// +kubebuilder:default=hetzner
+	// Name defines the name of the secret. The name gets used for reading the credential in the
+	// mgt-cluster, and it gets used for creating a secret in the wl-cluster. About the secret in
+	// the wl-cluster: Attention, the upstream hcloud-ccm helm chart expects the name to be
+	// "hcloud". The Syself ccm fork defaults to "hetzner". For compatibility with upstream hcloud-ccm
+	// the controller creates two secrets, if the name is different from "hcloud" (one with name
```
janiskemper: I don't understand the text in the brackets

api/v1beta1/types.go
janiskemper: I'm not sure if this means that we should change the default.. 
Especially given the fact that we hardcode "token" anyway and create is anyway, I don't see a clear strategy here yet. We say that we recommend "token", but then we say that we create it anyway, and we also change the default to it, even though we create it anyway.

api/v1beta1/types.go
janiskemper: if both are hardcoded in the code, why changing the default here?

api/v1beta1/types.go
guettli: some day I would like to change that. But we can do that later. Default was changed back to "hetzner": 438e674b

api/v1beta1/types.go
guettli: I updated the wording. Better now? 0a87ea0a

api/v1beta1/types.go
guettli: I am unsure, too.

api/v1beta1/types.go
guettli: For me it makes sense, when the default is equal to the upstream hcloud ccm. This looks aligned.

docs/caph/01-getting-started/02-quickstart/03-creating-a-workload-cluster.md
```diff
@@ -73,6 +73,11 @@ clusterctl get kubeconfig my-cluster > $CAPH_WORKER_CLUSTER_KUBECONFIG
 
 Cilium is used as a CNI solution in this guide. The following command deploys it to your cluster:
 
+The file `templates/cilium/cilium.yaml` is a repo-provided Helm values file in this repository.
```
janiskemper: these changes seem unrelated. Do you wanna open another PR?

guettli: @janiskemper moved to: [:book: Improve docs "Creating a Workload Cluster" by guettli · Pull Request #1899 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1899)

api/v1beta1/types.go
janiskemper: @guettli I wouldn't change the default here.

api/v1beta1/types.go
```diff
@@ -113,18 +123,32 @@ type HetznerSecretRef struct {
 // Need to specify either HCloudToken or both HetznerRobotUser and HetznerRobotPassword.
 type HetznerSecretKeyRef struct {
 	// HCloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
+	// We recommend to use "token", because this is the default of upstream hcloud-ccm, while the
+	// legacy Syself ccm fork uses "hcloud". For maximal compatibility up to three keys get created in the
```
janiskemper: the current Syself CCM does NOT use the default that we set here?! I would expect this to be the case. If it is, I would leave the default and only create the one the upstream CCM expects.

api/v1beta1/types.go
janiskemper: This is our API and if people use the default, then this would be unnecessarily breaking. I would leave the default

janiskemper:
@guettli this PR is huge. I think it takes a long time to even find the comments. Do you just wanna close this and create a new one? Maybe with fewer commits? Please incorporate my comments though.

api/v1beta1/types.go
guettli: @janiskemper ok, I kept the default, and added some more docs. It is important to know that this value gets used twice: In the mgt-cluster and in the wl-cluster. See 9b47d8d1

api/v1beta1/types.go
guettli: @janiskemper yes, strange but true. The ccm itself (Go code) does not read K8s secret. It is up to the deployment to either set env-vars or a file. They key our ccm uses is "hcloud".

guettli:
> @guettli this PR is huge. I think it takes a long time to even find the comments. Do you just wanna close this and create a new one? Maybe with fewer commits? Please incorporate my comments though.

OK, I close this PR, and we continue on the new II #1920

PR #1713, opened 2025-11-10
:seedling: update-operator-dev-deployment.sh: Fail if background job fails.

guettli:
update-operator-dev-deployment.sh: Fail if background job fails.

Additionally: if there is no git branch active, try to get the git tag for the container-image name. Useful, if you used something like `git switch -d v1.0.7`.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -316,6 +316,8 @@ func rebootTypesFromStringList(rebootTypeStringList []string) ([]infrav1.RebootT
 	return rebootTypes, nil
 }
 
+// ensureSSHKey ensures that the given ssh key is known to the Robot-API.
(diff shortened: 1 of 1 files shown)


hack/update-operator-dev-deployment.sh
```diff
@@ -66,6 +66,10 @@ if ! echo "$current_context" | grep -P '.*-admin@.*-mgt-cluster'; then
 fi
 
 branch=$(git branch --show-current)
+if [ "$branch" == "" ]; then
```
Dhairya-Arora01, line 71: can you add a comment here too

Dhairya-Arora01 (approved):

hack/update-operator-dev-deployment.sh
guettli: @Dhairya-Arora01 I added comments to the script. Please check again.

Issue #1506, opened 2024-11-08
Flaky unit-test: Hetzner ClusterReconciler cluster tests load balancer [It] should take over an existing load balancer with correct name


guettli:
/kind bug

Test unit-test is flaky:

```
Summarizing 1 Failure:
  [FAIL] Hetzner ClusterReconciler cluster tests load balancer [It] should take over an existing load balancer with correct name
  /home/runner/work/cluster-api-provider-hetzner/cluster-api-provider-hetzner/controllers/hetznercluster_controller_test.go:562
```

Here it failed:

https://github.com/syself/cluster-api-provider-hetzner/actions/runs/11740839469/job/32708240520#step:7:1920

When retrying it, the test did not fail again.

guettli:
It is still flaky:

[:seedling: Restart OCI download if too slow. Timeout after ten minute… · syself/cluster-api-provider-hetzner@7c5a729](https://github.com/syself/cluster-api-provider-hetzner/actions/runs/15607137837/job/43959194939#step:7:5116)
```
Summarizing 1 Failure:
  [FAIL] Hetzner ClusterReconciler cluster tests load balancer [It] should take over an existing load balancer with correct name
  /home/runner/work/cluster-api-provider-hetzner/cluster-api-provider-hetzner/controllers/hetznercluster_controller_test.go:585

Ran 114 of 114 Specs in 47.198 seconds
FAIL! -- 113 Passed | 1 Failed | 0 Pending | 0 Skipped
```

Issue #1435, opened 2024-08-13
Flaky Test: HCloudMachineReconciler Basic test correct server [It] creates the HCloud machine in Hetzner 1 


guettli:
/kind bug

This unit-test is flaky:

>  [FAIL] HCloudMachineReconciler Basic test correct server [It] creates the HCloud machine in Hetzner 1 (flaky)
>  /home/runner/work/cluster-api-provider-hetzner/cluster-api-provider-hetzner/controllers/hcloudmachine_controller_test.go:380

Example: https://github.com/syself/cluster-api-provider-hetzner/actions/runs/10366318563/job/28695292042?pr=1433#step:7:1952

Running the test again fixes the issue.

But this is not nice.

Why does it sometimes fail?

PR #1723, opened 2025-11-20
:seedling: Fix flaky unit-tests 🎉🎉🎉

### LINKED ISSUE #1685: Flaky Test: checks if PhaseWaiting is set when retryLimit is reached
/kind bug

This test is flaky:

>   [FAIL] HCloudRemediationReconciler Basic hcloudremediation test [It] checks if PhaseWaiting is set when retryLimit is reached

Fail: https://github.com/syself/cluster-api-provider-hetzner/actions/runs/18000294399/job/51215860077

ok after re-run: https://github.com/syself/cluster-api-provider-hetzner/actions/runs/18000294399

guettli:
Before this PR roughly 3 of 10 `make test-unit` calls failed because of a flaky test. Many tests already contained the string "(flaky)" in the name.

The root-cause is that mocks and reconcilers are shared between tests.

All tests run sequentially (no concurrency). But changes done in one test could influence the second test.

Running tests in isolation (for example via `FIt()`) always succeeded.

This PR updates CreateNamespace() to ResetAndCreateNamespace().

Mocks get re-created via a Reset() method.

On my local device the tests were successful 70 times in a row (still running).

---

Introduces a Resetter/ResetAndCreateNamespace pattern for tests, to:

- re-create mocks cleanly per test,
- set a test namespace on reconcilers.

---

Resetting alone did not remove the flakiness. This PR ensures that mocks which have a state (like HCloudClient) do not get used, when a test has finished. 

The background is explained in the [kubebuilder docs: envTest, Namespace usage Limitations](https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation): Namespaces can't be deleted in envTests. This means objects from test-1 might still exist and getting reconciled while test-2 runs. When test-1 uses the stateful hcloudClient, although this test has already passed, it modifies the environment of test-2. Resetting does not resolve this.

This PR uses the pattern suggested by the kubebuilder docs: FooReconciler.Namespace. At the top of the Reconcile method the namespace gets checked. If set and different, reconcile is stopped.

---

The test for hbmm deletion was updated. It checks that the host gets deprovisioned. And the same test gets done in phase "ensure-provisioned". 

---

HetznerBareMetalHostReconciler in "avoid conflict errors": Ignore NotFound error.

---

Closes #1685 #1506 #1435

### DIFF
--- a/test/helpers/envtest.go
+++ b/test/helpers/envtest.go
@@ -57,8 +57,6 @@ import (
 	robotclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/robot"
 	sshclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/ssh"
 	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
-	fakeclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client/fake"
-	"github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/mockedsshclient"
 	"github.com/syself/cluster-api-provider-hetzner/pkg/utils"
 )
 


controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -390,16 +391,83 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
...
+				Expect(host.Spec.Status.ProvisioningState).To(Equal(infrav1.StateProvisioned))
+
+				By("Setting State to 'ensure-provisioned'")
+				host.Spec.Status.ProvisioningState = infrav1.StateEnsureProvisioned
+				err = testEnv.Update(ctx, host)
+				Expect(err).To(BeNil())
+
+				Eventually(func() error {
+					err := testEnv.Get(ctx, hostKey, host)
+					apierrors.IsNotFound(err)
```
Dhairya-Arora01: This should be removed, I guess.

Dhairya-Arora01 (approved):

janiskemper (changes requested):
the test doesn't test deletion anymore, but deprovisioning of the host when a hetznerbaremetalmachine is deleted. Can you explain why you change the test? Also, if you wanna do that change, you should update the test description

controllers/hetznerbaremetalmachine_controller_test.go
guettli: thank you, fixed.

test/helpers/envtest.go
```diff
@@ -106,6 +104,19 @@ func init() {
 	}
 }
 
+// Resetter provides an interface, so that test-suites using TestEnvironment can define their custom
+// resetting. Mocks get used in two places: In the real code (like FooReconciler) and during the
```
janiskemper: mocks get used in real code? What do you mean by that

test/helpers/envtest.go
```diff
@@ -106,6 +104,19 @@ func init() {
...
+// tests (for FooMock.On("....")). The Resetter contains the pointer to the places where mocks need
+// to be updated on Reset().
+type Resetter interface {
+	// Reset resets the objects shared between tests. This avoids that changes done by the first
+	// test will modify the environment of the second test. Testify Mocks should be initialized with
```
janiskemper: I don't know this Test function. Where does it come from? What does it do? Can you maybe give a more specific example of this panic situation?

test/helpers/envtest.go
```diff
@@ -191,29 +204,15 @@ func NewTestEnvironment() *TestEnvironment {
 		klog.Fatalf("failed to set up webhook with manager for HCloudRemediationTemplate: %s", err)
 	}
 	// Create a fake HCloudClientFactory
-	hcloudClientFactory := fakeclient.NewHCloudClientFactory()
-
-	rescueSSHClient := &sshmock.Client{}
```
janiskemper, line 198: you remove all of this from the init function. Why? If we do that, we should specify a reason why we don't initialize - probably in the function description

controllers/controllers_suite_test.go
```diff
@@ -63,6 +68,74 @@ func TestControllers(t *testing.T) {
...
+	testEnv.HCloudSSHClientFactory = mockedsshclient.NewSSHFactory(hcloudSSHClient)
+	testEnv.RescueSSHClient = rescueSSHClient
+	testEnv.OSSSHClientAfterInstallImage = osSSHClientAfterInstallImage
+	testEnv.OSSSHClientAfterCloudInit = osSSHClientAfterCloudInit
+	testEnv.RobotClientFactory = robotClientFactory
+	testEnv.RobotClient = robotClient
+
+	// Reset clients used by Reconcile() and the namespace
+	r.HetznerClusterReconciler.HCloudClientFactory = hcloudClientFactory
+	r.HetznerClusterReconciler.Namespace = namespace
```
janiskemper, line 139: this operation doesn't sound like "reset" to me. It is actually an initialization with a new namespace! What do you suggest we should do in order to make this more transparent?

controllers/controllers_suite_test.go
```diff
@@ -71,51 +144,62 @@ var _ = BeforeSuite(func() {
...
-		Client:              testEnv.Manager.GetClient(),
-		APIReader:           testEnv.Manager.GetAPIReader(),
-		RateLimitWaitTime:   5 * time.Minute,
-		HCloudClientFactory: testEnv.HCloudClientFactory,
-	}).SetupWithManager(ctx, testEnv.Manager, controller.Options{})).To(Succeed())
-
-	Expect((&HetznerBareMetalRemediationReconciler{
+	}
+	Expect(hetznerClusterReconciler.SetupWithManager(ctx, testEnv.Manager, controller.Options{})).To(Succeed())
+	resetter.HetznerClusterReconciler = hetznerClusterReconciler
```
janiskemper: I suggest having an initialization function of the resetter struct, instead of defining this separately. I think that's the best practice in Go.

janiskemper (changes requested):

test/helpers/envtest.go
guettli: Good question. I thought about a better text, but finally decided to keep it short, and removed the sentence.

```go
// Resetter provides an interface, so that test-suites using TestEnvironment can define their custom
// resetting.
type Resetter interface {
```

janiskemper: Okay I have no idea what this sentence with the mock meant. If it is not relevant, then I'm fine like this.

test/helpers/envtest.go
guettli: Docs of testify: [Test()](https://pkg.go.dev/github.com/stretchr/testify@v1.11.1/mock#Mock.Test)

I have seen that several times before: a mock panics because it gets called in an expected way. This panic happens in Reconcile(), and the test does not fail immediately. Maybe the test fails (timeout) or it even sometimes passes.

If you use `Test(t)` then the test fails directly when the mock fails (no panic). This is very handy. I learned that two days ago, so this is the first time we use that.

janiskemper: > Test should not be called on an object that is going to be used in a goroutine other than the one running the test function.

This is from the docs of the function. Since we use an event-based system with Reconcile functions not being called by us, it seems to me as if we shouldn't use this function.

test/helpers/envtest.go
guettli: I updated the docstring:

```go
// NewTestEnvironment creates a new environment spinning up a local api-server. Factories for
// clients (like HCloudClientFactory) are not created. This gets done in ResetAndCreateNamespace(),
// which should be use at the beginning of each test.
func NewTestEnvironment() *TestEnvironment {
```

ok like this?

controllers/controllers_suite_test.go
guettli: We could rename Reset() to ResetTestEnvironment(). 

Let's look at the usage:

```
testEnv.Reset()
```

vs

```
testEnv.ResetTestEnvironment()
```

I think this does not make a big difference.

For me the name is fine. Please suggest an alternative, if you want to change the function name.

guettli: I updated the docstring of the interface:

```go
	// Reset resets the objects shared between tests. This avoids that changes done by the first
	// test will modify the environment of the second test.
	//
	// namespace: is the name of the new namespace which the test will use. Related: https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
	//
	// testEnv: the TestEnvironment which should be resetted.
	//
	// t: g.GinkgoT()
	Reset(namespace string, testEnv *TestEnvironment, t g.FullGinkgoTInterface)
```

Maybe it is more clear now?

controllers/controllers_suite_test.go
guettli: done in 11015478

test/helpers/envtest.go
guettli: I am unsure if that comment in the docstring of Test() does apply for us. It works fine.

Without Test():

```
------------------------------
• [FAILED] [7.035 seconds]
HetznerBareMetalHostReconciler Tests with bm machine Test root device life cycle [It] gives an error if no root device hints are set
/home/guettli/syself/cluster-api-provider-hetzner/controllers/hetznerbaremetalhost_controller_test.go:318

  [FAILED] Timed out after 5.000s.
  Expected
      <bool>: false
  to be true
  In [It] at: /home/guettli/syself/cluster-api-provider-hetzner/controllers/hetznerbaremetalhost_controller_test.go:324 @ 11/24/25 12:39:06.194
------------------------------
```

Somewhere in the logs you see::

```
E1124 12:39:03.791820 2131287 controller.go:347] "Reconciler error" err=<
    panic: 
    assert: mock: I don't know what to return because the method call was unexpected.
        Either do Mock.On("GetBMServer").Return(...) first, or remove the GetBMServer() call.
        This method was unexpected:
            GetBMServer(int)
            0: 1
```

If you run several tests, then you do not know which test that panic belongs to.

---
---

With someMock.Test(t)

```
------------------------------
• [FAILED] [7.053 seconds]
HetznerBareMetalHostReconciler Tests with bm machine Test root device life cycle [It] gives an error if no root device hints are set
/home/guettli/syself/cluster-api-provider-hetzner/controllers/hetznerbaremetalhost_controller_test.go:318

  [FAILED] 
  assert: mock: I don't know what to return because the method call was unexpected.
    Either do Mock.On("GetBMServer").Return(...) first, or remove the GetBMServer() call.
    This method was unexpected:
        GetBMServer(int)
        0: 1
    at: [/home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/mocks/robot/Client.go:85 /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/host/host.go:177 /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/host/state_machine.go:255 /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/host/state_machine.go:106 /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/host/host.go:140 /home/guettli/syself/cluster-api-provider-hetzner/controllers/hetznerbaremetalhost_controller.go:295 /home/guettli/syself/cluster-api-provider-hetzner/controllers/hetznerbaremetalhost_controller.go:288 /home/guettli/syself/cluster-api-provider-hetzner/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:119 /home/guettli/syself/cluster-api-provider-hetzner/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:334 /home/guettli/syself/cluster-api-provider-hetzner/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:294 /home/guettli/syself/cluster-api-provider-hetzner/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:255 /home/linuxbrew/.linuxbrew/Cellar/go/1.25.1/libexec/src/runtime/asm_amd64.s:1693]
  In [It] at: /home/guettli/syself/cluster-api-provider-hetzner/vendor/github.com/stretchr/testify/mock/mock.go:351 @ 11/24/25 12:37:58.435
```

This is much better, because you see where the error belongs to.

janiskemper: I don't know it either. If we decide to use it, I would ask you to update the description. In the description you write that it panics otherwise. I'm not sure if it is just imprecise or incorrect. Because the actual behavior is what you showed in the example: that even though the code exits, nothing happens and we silently fail because of timeouts

test/helpers/envtest.go
janiskemper: maybe I misunderstood - can you explain me what you do with the namespace that you give as input? In the code I see that you just take the namespace that you freshly created as input and initialize the Reconciler with it. That very much looks like an initialization and not a reset. 

You named the other function "ResetAndCreateNamespace"

test/helpers/envtest.go
guettli: BTW, I noticed that the generated code has a `NewClient(t)` function which seems to be more common. That gets used now: a0e24a6d

guettli: > I would ask you to update the description. In the description you write that it panics otherwise.

I removed that sentence from the comment. It is an implementation detail. We should be aware of that, and use NewClient(t) in the future. 

The best solution would be a linter which forces us to use NewClient(t). But I found none.

test/helpers/envtest.go
guettli: The namespace is documented like this now:

The background gets explained here: https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation

That's why I added the link to the docs.

```go 
// Resetter provides an interface, so that test-suites using TestEnvironment can define their custom
// resetting.
type Resetter interface {
	// Reset resets the objects shared between tests. This avoids that changes done by the first
	// test will modify the environment of the second test.
	//
	// namespace: is the name of the new namespace which the test will use. Related: https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
	//
	// testEnv: the TestEnvironment which should be resetted.
	//
	// t: g.GinkgoT()
	Reset(namespace string, testEnv *TestEnvironment, t g.FullGinkgoTInterface)
}
```

What do you want to add or change?

test/helpers/envtest.go
guettli: NewClient(t) did not work, because this requires that all On(...) calls get actually made.

This would be great, but is too much work for now. We can clean that up later.

It makes sense that every On(...) gets actually made, and if not, the test fails. But that is a legacy issue which is not part of that PR.

test/helpers/envtest.go
janiskemper: you have a Reset function which initializes a new namespace with the reconciler. You call this Reset function in another function ResetAndCreateNamespace. In theory the function should be ResetAndInitNamespace and not Reset.

janiskemper: btw I realized that I commented here but it is rather about the "	r.HetznerClusterReconciler.Namespace = namespace" conversation

controllers/controllers_suite_test.go
janiskemper: > maybe I misunderstood - can you explain me what you do with the namespace that you give as input? In the code I see that you just take the namespace that you freshly created as input and initialize the Reconciler with it. That very much looks like an initialization and not a reset.
You named the other function "ResetAndCreateNamespace"


> you have a Reset function which initializes a new namespace with the reconciler. You call this Reset function in another function ResetAndCreateNamespace. In theory the function should be ResetAndInitNamespace and not Reset.


these were the comments that I wrote in the other chat but should have written here

guettli: Great, renamed it to `ResetAndInitNamespace`.

test/helpers/envtest.go
guettli: I renamed Reset() to ResetAndInitNamespace().

test/helpers/envtest.go
```diff
@@ -115,15 +127,19 @@ type TestEnvironment struct {
...
 	OSSSHClientAfterCloudInit    *sshmock.Client
 	RobotClient                  *robotmock.Client
 	cancel                       context.CancelFunc
 	RateLimitWaitTime            time.Duration
+	Resetter                     Resetter
 }
 
-// NewTestEnvironment creates a new environment spinning up a local api-server.
+// NewTestEnvironment creates a new environment spinning up a local api-server. Factories for
+// clients (like HCloudClientFactory) are not created. This gets done in ResetAndCreateNamespace(),
```
janiskemper: ResetAndInitNamespace?

controllers/controllers_suite_test.go
```diff
@@ -63,59 +66,152 @@ func TestControllers(t *testing.T) {
...
+		HCloudRemediationReconciler:           hcloudRemediationReconciler,
+		HetznerBareMetalRemediationReconciler: hetznerBareMetalRemediationReconciler,
+	}
+}
+
+var _ helpers.Resetter = &ControllerResetter{}
+
+func (r *ControllerResetter) ResetAndInitNamespace(namespace string, testEnv *helpers.TestEnvironment, t FullGinkgoTInterface) {
+	rescueSSHClient := &sshmock.Client{}
+	rescueSSHClient.Test(t)
```
janiskemper, line 106: can you maybe comment here what it does?

guettli: I added this comment:

```go
// ResetAndInitNamespace implements Resetter.ResetAndInitNamespace(). Documentation is on the
// interface.
```

or do you think custom documentation is needed?

test/helpers/envtest.go
guettli: new comment:

```go
// NewTestEnvironment creates a new environment spinning up a local api-server. Factories for
// clients (like HCloudClientFactory) are not created. This gets done in
// Resetter.ResetAndInitNamespace(), which should be use at the beginning of each test.

```

controllers/controllers_suite_test.go
janiskemper: I actually was referring to the Test() method! I didn't know it before and I believe I'll not be the only one, so it would be nice to have a quick comment describing what it is used for.

guettli: ok, done.

Dhairya-Arora01 (approved):

PR #1732, opened 2025-11-26
:seedling: Less noise for unit-test

guettli:
use hack/filter-caph-controller-manager-logs.py to filter noise.


janiskemper:
from my side it's fine to merge - it only is about a new feature for unit tests.

PR #1735, opened 2025-11-27
:seedling: Refactoring PR for: Remove usage of FailureReason and FailureMessage (baremetal)

guettli:
Some changes of this PR got extracted, so that the PR gets smaller and easier to review:

[:seedling: Remove usage of FailureReason and FailureMessage (baremetal) by guettli · Pull Request #1716 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1716)

- Add/extended comments to some parts.
- move rate-limit handling code up, to return early if rate-limit is active.
- avoid usage of capierrors package.
- test code: Align capi-machine name and infra-machine name. Remediation code depends on that capi feature.
- `tee tmp/test-unit-unfiltered.log` in test-unis.sh: Write unmodfied output to tmp directory. This is handy if you want to see all logs, without running tests again.
- export func GetAssociatedHost(), so that it can be re-used.
- refactored getAssociatedHost() to getAssociatedHostAndPatchHelper(), so that the code can be re-used.
- add log when hbmm is deleting.
- avoid duplicate implemenation of splitHostKey()
- refactored setOwnerRemediatedConditionNew() to setOwnerRemediatedConditionToFailed(): Avoid duplicated code by adding functionality to setOwnerRemediatedConditionToFailed(): Use the same message for the created Event and the condition, and set Phase to PhaseDeleting.
- removed named return parameters: `return res, nil --> return reconcile.Result{}, nil`
- hbmm have an annotation pointing to the hbmh. The reference is a string in the format "namespace/hbmh-name". The namespace will be ignored from now on, because cross-namespace references are not allowed. This is not a breaking change, because we expect that both namespace strings are always equal.

### DIFF
--- a/pkg/services/baremetal/remediation/remediation.go
+++ b/pkg/services/baremetal/remediation/remediation.go
@@ -53,11 +53,6 @@ func NewService(scope *scope.BareMetalRemediationScope) *Service {
 func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	host, err := host.GetAssociatedHost(ctx, s.scope.Client, s.scope.BareMetalMachine)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return reconcile.Result{}, s.setOwnerRemediatedConditionToFailed(ctx,
-				"exit remediation because host not found")
-		}
-
 		// retry
 		err := fmt.Errorf("failed to find the unhealthy host (will retry): %w", err)
 		record.Warn(s.scope.BareMetalRemediation, "FailedToFindHost", err.Error())
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -40,7 +40,9 @@ const (
 	DeprecatedBareMetalHostFinalizer = "hetznerbaremetalhost.infrastructure.cluster.x-k8s.io"
 
 	// HostAnnotation is the key for an annotation that should go on a HetznerBareMetalMachine to
-	// reference what HetznerBareMetalHost it corresponds to.
+	// reference what HetznerBareMetalHost it corresponds to. The annotation is a string in the
+	// format "namespace/hbmh-name". Note: We should remove the namespace, as cross-namespace
+	// references are not allowed.
 	HostAnnotation = "infrastructure.cluster.x-k8s.io/HetznerBareMetalHost"
 
 	// WipeDiskAnnotation indicates which Disks (WWNs) to erase before provisioning
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -132,7 +132,7 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 func (s *Service) Delete(ctx context.Context) (reconcile.Result, error) {
 	// get host - ignore if not found
 	host, helper, err := s.getAssociatedHostAndPatchHelper(ctx)
-	if err != nil && !apierrors.IsNotFound(err) {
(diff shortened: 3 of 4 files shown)


pkg/services/baremetal/remediation/remediation.go
```diff
@@ -51,56 +50,41 @@ func NewService(scope *scope.BareMetalRemediationScope) *Service {
...
-			return res, err
-		}
-		record.Eventf(
-			s.scope.BareMetalRemediation,
-			"ExitRemediation",
-			"exit remediation because host is not provisioned or in maintenance mode. Provisioning state: %s. Maintenance mode: %v",
-			host.Spec.Status.ProvisioningState, host.Spec.MaintenanceMode != nil && *host.Spec.MaintenanceMode,
-		)
-		return res, nil
+	if host == nil {
```
janiskemper, line 62: can this happen? We handle the not found error... at least it wouldn't make sense

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -217,48 +191,50 @@ func (s *Service) timeUntilNextRemediation(now time.Time) time.Duration {
 	return nextRemediation
 }
 
-// setOwnerRemediatedConditionNew sets MachineOwnerRemediatedCondition on CAPI machine object
+// setOwnerRemediatedConditionToFailed sets MachineOwnerRemediatedCondition on CAPI machine object
 // that have failed a healthcheck.
-func (s *Service) setOwnerRemediatedConditionNew(ctx context.Context) error {
+func (s *Service) setOwnerRemediatedConditionToFailed(ctx context.Context, msg string) (reconcile.Result, error) {
```
janiskemper: you return reconcile result but don't ever have a non-empty value.. That doesn't make sense

pkg/services/baremetal/remediation/remediation.go
guettli: @janiskemper, yes, it can be nil:

```go
func GetAssociatedHost(ctx context.Context, crClient client.Client, hbmm *infrav1.HetznerBareMetalMachine) (*infrav1.HetznerBareMetalHost, error) {
	annotations := hbmm.GetAnnotations()
	// if no annotations exist on machine, no host can be associated
	if annotations == nil {
		return nil, nil
	}

```

janiskemper: okay, but that's weird for that GetAssociatedHost function no? Then we should probably directly handle the not found error and convert it to nil. 

Or is there any reason that the function returns in two ways the same thing?

pkg/services/baremetal/remediation/remediation.go
guettli: yes, this makes sense. 

I updated it: The function only returns an error, no more a reconcile.Result.

pkg/services/baremetal/remediation/remediation.go
guettli: @janiskemper I just refactored the method, so that it could be re-used. It was like this before. For now, I would like to keep it like this. But I could change it, if you think so.

janiskemper: I would like to change it. Before, it was not a library function, so it was clear from the inline code. Right now, with having this in a separate package, that looks strange.

guettli: @janiskemper: GetAssociatedHost() now returns nil, when the api-server returns not-found for that hbmh.

ok now?

janiskemper:
okay understood it now. Can you plz add this information to the PR description?

guettli:
> okay understood it now. Can you plz add this information to the PR description?

done. Anything else which is blocking? I would like to merge the PR.

PR #1744, opened 2025-12-05
:book: Announce v1.1.x branch and our current way of handling PRs.

guettli:
Announce v1.1.x branch and our current way of handling PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,14 +35,17 @@
 ## You are seeing CAPH v1.0.x
 
 On December 5, 2025, we reset the main branch to the git tag
-[v1.0.7](https://github.com/syself/cluster-api-provider-hetzner/tree/v1.0.7) so paying customers
-see the documentation for the service we currently offer.
+[v1.0.7](https://github.com/syself/cluster-api-provider-hetzner/tree/v1.0.7). The first v1.1.x
+release is still in progress, and documentation published from `main` is often assumed to describe
+the latest stable release. To avoid confusion, we continue development and documentation work for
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,18 @@
 - [Contributing](#-getting-involved-and-contributing)
 - [Contact](#-contact)
 
+## You are seeing CAPH v1.0.x
+
+On December 5, 2025, we reset the main branch to the git tag
+[v1.0.7](https://github.com/syself/cluster-api-provider-hetzner/tree/v1.0.7) so paying customers
```
batistein: It's not about paying customers. The v1.1.x version hasn't even been released yet, and we don't want to publish documentation for something that's still in progress. The problem with documentation in the main branch is that people often assume it's the latest stable version. To avoid confusion, we have decided to continue development and documentation in a different branch. Once the first version of v1.1.x is released and ready to use, we can use the main branch to update the documentation on the hosted page.

README.md
```diff
@@ -32,6 +32,18 @@
...
+see the documentation for the service we currently offer.
+
+Development for CAPH v1.1.x happens on the
+[v1.1.x](https://github.com/syself/cluster-api-provider-hetzner/tree/v1.1.x) branch. Please base
```
batistein, line 44: shouldn't it be raise instead of base?

guettli: I (and AI) thinks "base" is the correct verb here.

README.md
guettli: @batistein README was updated. Please check again.

PR #1758, opened 2025-12-12
:seedling: Better error message, when hcloud create server fails (to main)

guettli:
Example (old)

> server name is already used (uniqueness_error, 3353a2d4b216ba6a74ea0c60f51ad7f1)

> server location disabled (resource_unavailable, 6b34ff946f6f3e38a789d43b55e8560c)

The new message includes: name, server-type and location.

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -474,21 +474,22 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
 			serverType = opts.ServerType.Name
 		}
 
-		msg := fmt.Sprintf("failed to create HCloud server %q in %q (type %q)",
-			hm.Name, opts.Location.Name, serverType)
+		msg := fmt.Sprintf("failed to create HCloud server in %q (type %q)",
+			opts.Location.Name, serverType)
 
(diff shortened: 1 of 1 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -468,27 +468,27 @@ func (s *Service) createServer(ctx context.Context) (*hcloud.Server, error) {
...
+		hm := s.scope.HCloudMachine
+		serverType := "nil"
+		if opts.ServerType != nil {
+			serverType = opts.ServerType.Name
+		}
+
+		msg := fmt.Sprintf("failed to create HCloud server %q in %q (type %q)",
+			hm.Name, opts.Location.Name, serverType)
+
+		if hcloudutil.HandleRateLimitExceeded(hm, err, "CreateServer") {
```
janiskemper, line 480: not related to this PR: Why do we have two "handleRateLimitExceeded" here? Something seems to be wrong

guettli: @janiskemper you are right. This duplicate rate-limit handling made no sense. I update the PR. Additionally, the machine name is no longer a part of the message. This is part of the Event and the Condition. No need to duplicate that.

PR #1770, opened 2025-12-16
:seedling: Remove usage of FailureReason and FailureMessage (hcloud, PR2)

guettli:
HCloud PR

> Deprecated: This field is deprecated and is
> going to be removed when support for v1beta1
> will be dropped. Please see
> https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240916-improve-status-in-CAPI-resources.md
> for more details.

Previous PR in "train": Baremetal PR [1716](https://github.com/syself/cluster-api-provider-hetzner/pull/1716)

Somehow the first PR (#1717) was automatically closed by Github. This is the second PR.

---

This PR handles the case where the controller set a FailureReason in the past. Now, that capi no longer handles FailureReason, our controller does that. The controller does not have the permission to delete a capi machine. So the controller sets an Annotation on the capi machine, so that capi starts remediation. Our remediation controller detects that the machine is not in State OperatingSystemRunning, and stops remediation (because rebooting makes no sense it that case).

Changes:

* Extracted function SetErrorAndRemediate(), so that it can be used without a MachineScope (for example in a test).
* Reconcile of hcloud remediation: Do not try to reboot, when HCloudBootStateProvisioningFailed is set. Use `setOwnerRemediatedConditionToFailed()` and stop reconciling.
* Do not reconcile hcloudmachine, when HCloudBootStateProvisioningFailed is set. Deletion is handled before that new if-statement.
* In test: do not test PlacementGroup. Otherwise, machine would not get provisioned (in test).
* added test: `It("should delete machine if SetErrorAndRemediate() was called" ...`
* tests: removed machinePatchHelper, because it was not used.
* Reconcile of hcloud remediation: Better err msg to differentiate between ProviderId is not set and server-not-found.
* Avoid using named returns, when not needed: return res, nil --> return reconcile.Result{}, nil
* Better err msg when a PlacementGroup does not exist.
* create server: Align message in condition to message in error.
* avoid using capierrors  (will be removed in future versions of capi).
* tests: align capi-machine name to hcloudmachine name.

### DIFF
--- a/pkg/services/hcloud/server/server_test.go
+++ b/pkg/services/hcloud/server/server_test.go
@@ -835,7 +835,7 @@ var _ = Describe("Reconcile", func() {
 		By("calling reconcile")
 		_, err := service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateBootingToRealOS))
 
 		By("ensuring the bootstate has transitioned to BootStateBootingToRealOS")
 
@@ -850,7 +850,7 @@ var _ = Describe("Reconcile", func() {
 
 		_, err = service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateOperatingSystemRunning))
 
 		By("ensuring the bootstate has transitioned to BootStateOperatingSystemRunning once the server's status changes to running")
 		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateOperatingSystemRunning))
@@ -912,7 +912,7 @@ var _ = Describe("Reconcile", func() {
 		By("calling reconcile")
 		_, err := service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateInitializing))
 
 		By("ensuring the bootstate has transitioned to Initializing")
 
@@ -943,7 +943,7 @@ var _ = Describe("Reconcile", func() {
 			}, nil).Once()
 		_, err = service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateEnablingRescue))
 
 		By("ensuring the bootstate has transitioned to EnablingRescue")
 		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateEnablingRescue))
@@ -970,7 +970,7 @@ var _ = Describe("Reconcile", func() {
 		)
 		_, err = service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateEnablingRescue))
 
 		By("ensuring the bootstate has transitioned to EnablingRescue")
 		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateEnablingRescue))
@@ -990,7 +990,7 @@ var _ = Describe("Reconcile", func() {
 		})
 		_, err = service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateBootingToRescue))
 
 		By("ensuring the bootstate has transitioned to BootingToRescue")
 		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateBootingToRescue))
@@ -1009,7 +1009,7 @@ var _ = Describe("Reconcile", func() {
 		startImageURLCommandMock := testEnv.HCloudSSHClient.On("StartImageURLCommand", mock.Anything, mock.Anything, mock.Anything, mock.Anything, "my-machine", []string{"sda"}).Return(0, "", nil)
 		_, err = service.Reconcile(ctx)
 		Expect(err).To(BeNil())
-		Expect(noErrorOccured(service.scope)).To(BeNil())
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateRunningImageCommand))
 
 		By("ensuring the bootstate has transitioned to RunningImageCommand")
 		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateRunningImageCommand))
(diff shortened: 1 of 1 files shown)


api/v1beta1/types.go
```diff
@@ -292,4 +292,8 @@ const (
 
 	// HCloudBootStateOperatingSystemRunning indicates that the server is successfully running.
 	HCloudBootStateOperatingSystemRunning HCloudBootState = "OperatingSystemRunning"
+
+	// HCloudBootStateProvisioningFailed indicates that provisioning failed. The capi machine, and
+	// the hcloud machine will get delelted.
```
janiskemper: typo

controllers/hcloudremediation_controller_test.go
```diff
@@ -318,5 +325,94 @@ var _ = Describe("HCloudRemediationReconciler", func() {
 					isPresentAndFalseWithReason(capiMachineKey, capiMachine, clusterv1.MachineOwnerRemediatedCondition, clusterv1.WaitingForRemediationReason)
 			}, timeout).Should(BeTrue())
 		})
+		It("should delete machine if SetErrorAndRemediate() was called", func() {
+			By("Checking Environment")
```
janiskemper: is this necessary? I don't really understand this check

controllers/hcloudremediation_controller_test.go
```diff
@@ -318,5 +325,94 @@ var _ = Describe("HCloudRemediationReconciler", func() {
...
+					return err
+				}
+				err = testEnv.Status().Update(ctx, hcloudMachine)
+				if err != nil {
+					return err
+				}
+				return nil
+			}).Should(Succeed())
+
+			By("Wait until hcloud has condition set.")
```
janiskemper: you mean boot state is set to failed

controllers/hcloudremediation_controller_test.go
```diff
@@ -318,5 +325,94 @@ var _ = Describe("HCloudRemediationReconciler", func() {
...
+			Eventually(func() error {
+				err := testEnv.Get(ctx, client.ObjectKeyFromObject(hcloudMachine), hcloudMachine)
+				if err != nil {
+					return err
+				}
+				if hcloudMachine.Status.BootState != infrav1.HCloudBootStateProvisioningFailed {
+					return fmt.Errorf("not set: HCloudBootStateProvisioningFailed")
```
janiskemper: "not set" doesn't fit IMO

controllers/hcloudremediation_controller_test.go
```diff
@@ -318,5 +325,94 @@ var _ = Describe("HCloudRemediationReconciler", func() {
...
+				},
+			}
+
+			err = controllerutil.SetOwnerReference(capiMachine, rem, testEnv.GetScheme())
+			Expect(err).Should(Succeed())
+
+			err = testEnv.Create(ctx, rem)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			By("Wait until remediation is done")
```
janiskemper: you check here only things that happen on CAPI side. 

I actually wonder right now how anything happens there.. Do we have CAPI controllers in our test setup?

Anyway: We should focus on testing our controller. Currently, I would say that we can do this by checking that certain actions were performed, or that the object has a certain status. I think that would make more sense than analyzing the CAPI objects.

pkg/scope/machine.go
```diff
@@ -126,13 +127,44 @@ func (m *MachineScope) PatchObject(ctx context.Context) error {
...
+// created, and the DeleteMachineSucceededCondition will be set to False on the hcloud-machine. It
+// gets used, when a not-recoverable error happens. Example: hcloud server was deleted by hand in
+// the hcloud UI.
+func (m *MachineScope) SetErrorAndRemediate(ctx context.Context, message string) error {
+	return SetRemediateMachineAnnotationToDeleteMachine(ctx, m.Client, m.Machine, m.HCloudMachine, message)
+}
+
+// SetRemediateMachineAnnotationToDeleteMachine sets "cluster.x-k8s.io/remediate-machine" annotation
+// on the corresponding CAPI machine. This will trigger CAPI to start remediation. Our remediation
+// contoller will use HasFatalError(öööö) to differentiate between a remediate (with reboot) and delete
```
janiskemper: ööö?

pkg/scope/machine.go
```diff
@@ -126,13 +127,44 @@ func (m *MachineScope) PatchObject(ctx context.Context) error {
...
+// (no reboot gets tried). Finally the capi machine and the infra machine will be deleted.
+//
+// Background: the hcloudmachine controller has no permission to delete a capi machine. That's why
+// this extra step (via remediate-machine annotation) is needed.
+func SetRemediateMachineAnnotationToDeleteMachine(ctx context.Context, crClient client.Client, capiMachine *clusterv1.Machine, hcloudMachine *infrav1.HCloudMachine, message string) error {
+	// Create a patch base
+	patch := client.MergeFrom(capiMachine.DeepCopy())
```
janiskemper, line 148: why don't we use the patch helper from CAPI?

pkg/services/hcloud/remediation/remediation.go
```diff
@@ -145,27 +159,23 @@ func (s *Service) handlePhaseRunning(ctx context.Context, server *hcloud.Server)
...
-func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result, err error) {
+func (s *Service) handlePhaseWaiting(ctx context.Context) (reconcile.Result, error) {
 	nextCheck := s.timeUntilNextRemediation(time.Now())
 
 	if nextCheck > 0 {
 		// Not yet time to stop remediation, requeue
 		return reconcile.Result{RequeueAfter: nextCheck}, nil
 	}
 
-	// When machine is still unhealthy after remediation, setting of OwnerRemediatedCondition
```
janiskemper, line 156: can you add this explanation to the comment of the setOwnerRemediatedConditionToFailed function?

pkg/services/hcloud/server/server_test.go
```diff
@@ -1032,3 +1058,10 @@ func isPresentAndFalseWithReason(getter conditions.Getter, condition clusterv1.C
 	return objectCondition.Status == corev1.ConditionFalse &&
 		objectCondition.Reason == reason
 }
+
+func noErrorOccured(s *scope.MachineScope) error {
```
janiskemper: I don't like this function. If you wanna use it, I propose to add it as method defined on BootState "IsFailed"

api/v1beta1/types.go
guettli: fixed.

controllers/hcloudremediation_controller_test.go
guettli: It was needed to find the reason why this did not work at the beginning. After fixing how the hcloudMachine gets created, this is no longer needed. Removed.

controllers/hcloudremediation_controller_test.go
guettli: fixed.

controllers/hcloudremediation_controller_test.go
guettli: fixed.

controllers/hcloudremediation_controller_test.go
guettli: @janiskemper I updated the By, maybe it is more clear now:

> 			By("Wait until our remediation controller has set condition on capi machine")


(no, capi is not running in the envTest).

pkg/scope/machine.go
guettli: Docstring was updated.

pkg/scope/machine.go
guettli: The patchHelper of capi does a lot of magic which is not needed here.

pkg/services/hcloud/remediation/remediation.go
guettli: @janiskemper I  updated the docstring:

```go 
// setOwnerRemediatedConditionToFailed sets MachineOwnerRemediatedCondition on CAPI machine object
// that have failed a healthcheck. This will make capi delete the capi and hcloud machine.
func (s *Service) setOwnerRemediatedConditionToFailed(ctx context.Context, msg string) error {
```

The old text about "preflight" was a bit confusing. Is that ok for you now?

pkg/services/hcloud/server/server_test.go
guettli: I don't like it, too. I inlined the function and it checks the bootState now.

PR #1794, opened 2026-02-24
✨Adding support for ClusterAPI IPAM Provider

Archisman-Mridha:

Adds support for obtaining `private IP` for a `Hetzner Bare Metal server` using `Cluster API IPAM Provider`.

**TODOs**:

- [ ] bind the private IP to the server's network interface, using `cloud-init`


Archisman-Mridha:
Oops, sorry for the mess. I found out that I need to create the PR against `v1.1.x` while creating the PR (and not making the changes).

PR #1857, opened 2026-03-11
:seedling: Handle rate-limits carefully

Dhairya-Arora01:
In some of the cases, function `handleRateLimit()` returns both server and error is nil. When server as nil is used at other places (which expects it to be non-nil) it causes errors.

We need to handle the cases where variable server is used and is nil carefully.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -116,6 +116,14 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	// update the machine
 	if err := s.update(ctx); err != nil {
+		if apierrors.IsNotFound(err) {
+			err := fmt.Errorf("host not found for machine %q. Setting Failure, remediation will start: %w", s.scope.Machine.Name, err)
+			s.scope.BareMetalMachine.SetFailure(capierrors.UpdateMachineError, err.Error())
+
+			// returning error as nil here because otherwise it will lead continuous reconcile loop.
+			return reconcile.Result{}, nil
+		}
+
 		return checkForRequeueError(err, "failed to update machine")
 	}
 
@@ -207,11 +215,6 @@ func (s *Service) Delete(ctx context.Context) (res reconcile.Result, err error) 
 func (s *Service) update(ctx context.Context) error {
 	host, helper, err := s.getAssociatedHost(ctx)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			err := fmt.Errorf("host not found for machine %q. Setting Failure, remediation will start: %w", s.scope.Machine.Name, err)
-			s.scope.BareMetalMachine.SetFailure(capierrors.UpdateMachineError, err.Error())
-			return err
-		}
 		return fmt.Errorf("failed to get host: %w", err)
 	}
 	if host == nil {
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/hetznercloud/hcloud-go/v2/hcloud"
 	corev1 "k8s.io/api/core/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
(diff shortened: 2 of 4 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -90,27 +91,69 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			conditions.MarkFalse(
+				s.scope.HCloudMachine,
+				infrav1.ServerAvailableCondition,
+				infrav1.ServerNotFoundReason,
+				clusterv1.ConditionSeverityError,
+				"failed to find server: %s",
+				err.Error(),
+			)
+
+			return reconcile.Result{}, nil
```
guettli: @Dhairya-Arora01 Why no requeue here?

pkg/services/hcloud/server/server.go
```diff
@@ -90,27 +91,73 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	conditions.MarkTrue(s.scope.HCloudMachine, infrav1.BootstrapReadyCondition)
 
-	// try to find an existing server
+	// try to find an existing server.
 	server, err := s.findServer(ctx)
 	if err != nil {
+		// if it is not a rate limit exceeded error, set ServerNotFound condition on HCloudMachine.
+		if !hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
```
janiskemper, line 98: I think we should log an error here or just return it. Why exactly do we requeue instead of returning the error?

pkg/services/hcloud/server/server.go
```diff
@@ -90,27 +91,73 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+				infrav1.ServerNotFoundReason,
+				clusterv1.ConditionSeverityError,
+				"failed to find server: %s",
+				err.Error(),
+			)
+
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+		}
+
+		// Don't set rate limit exceeded condition on HCloudMachine, if providerID is set.
```
janiskemper: we can improve this comment

"We have a rate limit exceeded error. However, if the providerID is already set, indicating that the server has been created previously, then we don't want to set the rate limit exceeded error in the status, because it is misleading"


Side note: Why do we use ProviderID? Is this the best option? Because it doesn't necessarily indicate that the reconcilement is ready. For example, if we still need to add the server to the control plane load balancer as target.

pkg/services/hcloud/server/server.go
```diff
@@ -90,27 +91,73 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+		if s.scope.HCloudMachine.Spec.ProviderID == nil {
+			hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+		}
+
 		return reconcile.Result{}, fmt.Errorf("failed to get server: %w", err)
```
guettli: @Dhairya-Arora01 

Why differentiate between ProviderID being set or not?

Why not do this for both cases (and remove the `if`)?

```go
hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
```

Dhairya-Arora01: If the provider id is set - that would mean server exists already and finding the server again lead to rate limit which is bit misleading that;s why we want to continue reconcile in that case.

guettli: ok, this looks good now:

```go 
	// try to find an existing server.
	server, err := s.findServer(ctx)
	if err != nil {
		// if it is not a rate limit exceeded error, set ServerNotFound condition on HCloudMachine.
		if !hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
			conditions.MarkFalse(
				s.scope.HCloudMachine,
				infrav1.ServerAvailableCondition,
				infrav1.ServerNotFoundReason,
				"failed to find server: %s",
				err.Error(),
			)

			return reconcile.Result{}, fmt.Errorf("failed to find server: %w", err)
		}

		// We have a rate limit exceeded error. However, if the providerID is already set,
		// indicating that the server has been created previously, then we don't want to set the rate limit exceeded error
		// in the status, because it is misleading.
		if s.scope.HCloudMachine.Spec.ProviderID == nil {
			hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
		}

	}
```

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -207,11 +215,6 @@ func (s *Service) Delete(ctx context.Context) (res reconcile.Result, err error)
 func (s *Service) update(ctx context.Context) error {
 	host, helper, err := s.getAssociatedHost(ctx)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			err := fmt.Errorf("host not found for machine %q. Setting Failure, remediation will start: %w", s.scope.Machine.Name, err)
-			s.scope.BareMetalMachine.SetFailure(capierrors.UpdateMachineError, err.Error())
-			return err
-		}
```
guettli, line 214: This makes tests fail:

```console 
❯ go test ./pkg/services/baremetal/baremetal -count=1               
Running Suite: Baremetal Suite - /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/baremetal
===================================================================================================================
Random Seed: 1773232057

Will run 41 of 41 specs
••••
------------------------------
• [FAILED] [0.000 seconds]
Service.update [It] marks BareMetalMachine as failed when the associated host no longer exists
/home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/baremetal/baremetal_test.go:365

  [FAILED] Expected
      <*errors.MachineStatusError | 0x0>: nil
  not to be nil
  In [It] at: /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/baremetal/baremetal_test.go:402 @ 03/11/26 13:27:37.812
------------------------------
••••••••••••••••••••••••••••••••••••

Summarizing 1 Failure:
  [FAIL] Service.update [It] marks BareMetalMachine as failed when the associated host no longer exists
  /home/guettli/syself/cluster-api-provider-hetzner/pkg/services/baremetal/baremetal/baremetal_test.go:402

Ran 41 of 41 Specs in 0.002 seconds
FAIL! -- 40 Passed | 1 Failed | 0 Pending | 0 Skipped
```

pkg/services/hcloud/server/server.go
```diff
@@ -213,6 +264,10 @@ func handleRateLimit(hm *infrav1.HCloudMachine, err error, functionName string,
 func (s *Service) Delete(ctx context.Context) (res reconcile.Result, err error) {
 	server, err := s.findServer(ctx)
 	if err != nil {
+		if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
```
janiskemper: shouldn't we do "handleRateLimit" here?

PR #1865, opened 2026-03-11
:sparkles: Use namespace annotation instead of --ignored-namespaces

### LINKED ISSUE #1864: Use Annotation on NS instead of command line arguments
At the beginning of every reconcile the controller should look at the namespace of the resource.

If the NS has Annotation capi.syself.com/skip-namespace with value true set, then this resource should not be reconciled.

Remove the command line argument `--ignored-namespaces` we will use the annotation only.

guettli:
## Summary
- remove the `--ignored-namespaces` command-line argument and all reconciler wiring for it
- add namespace-level skip logic based on annotation `capi.syself.com/skip-namespace: "true"`
- apply the namespace annotation check at the start of every CAPH reconciler loop that handles namespaced CAPH resources
- add RBAC permission to read namespaces
- document the new annotation and add unit tests for the namespace skip helper

## Testing
- `go test ./controllers -run Test_shouldSkipReconciliationForNamespace -count=1`
- `go test ./... -run '^$'`
- `go test ./...` (fails in this environment due to missing envtest binaries and missing e2e config)

Closes #1864

### DIFF
--- a/controllers/namespace_skip.go
+++ b/controllers/namespace_skip.go
@@ -23,9 +23,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-)
 
-const skipNamespaceAnnotation = "capi.syself.com/skip-namespace"
+	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
+)
 
 //+kubebuilder:rbac:groups="",resources=namespaces,verbs=get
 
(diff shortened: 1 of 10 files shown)


controllers/namespace_skip.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const skipNamespaceAnnotation = "capi.syself.com/skip-namespace"
```
janiskemper: let's define this in the api folder with the other annotations

guettli: Done, I moved the constant to `api/v1beta1` (`SkipNamespaceAnnotation`) and switched controllers/tests to use the shared API constant.

PR #1868, opened 2026-03-12
:sparkles: make existing subnet logic more robust

skuethe:

I saw that https://github.com/syself/cluster-api-provider-hetzner/pull/1633 got merged although there where still some open concerns in the comments about it's implementation.

- the list of subnet returned from HCloud API is unstable, so validating against the first [subnets IPRange](https://github.com/syself/cluster-api-provider-hetzner/pull/1633/changes#diff-05e1ae5363dcc702c1b1f13e95187ee1c642b973f2c440a6f28f1e1fcaaa9b35R177) is unstable
- if we have multiple subnet of type `cloud`, there is no guarantee that created resources are being applied to the correct subnet by HCloud

This PR is my take in making that logic more robust. It:

- allows only **one** existing subnet of type `cloud`
- allows any additional number of subnets with type `vswitch` (although HCloud is limited to one vswitch subnet per network)
- validates the subnet CIDR of an existing subnet of type `cloud` against the configured subnet CIDR

So, in detail it explicitly fails if:

- there are two or more subnet of type `cloud`
- there is one subnet of type `vswitch` but no subnet of type `cloud`
- there is one subnet of type `cloud` but it does not match the configured CIDR

- I explicitly changed the logic from `len(networks[0].Subnets) > 1` to `len(networks[0].Subnets) >= 1` (line 175) so that we can always validate the configured subnets CIDR
- I decided to only "ignore" subnets of type `NetworkSubnetTypeVSwitch` and not `NetworkSubnetTypeServer`, as those are deprecated and AFAIK unavailable anyways.
- This subnet logic can be improved once upgrading to `hcloud-go` version `>= 2.30.0`, as it then allows to specify the correct `IPRange` via `ServerAttachToNetworkOpts` to target the correct subnet even if there are multiple subnet of type `cloud`
- I modified and extended the network tests. Test output:
```shell
hack/tools/bin/ginkgo run  ./pkg/services/hcloud/network | ./hack/filter-caph-controller-manager-logs.py -

2026/03/12 09:58:24 maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined
Running Suite: Network Suite - /home/sk/git/github/skuethe/cluster-api-provider-hetzner/pkg/services/hcloud/network
===================================================================================================================

Will run 10 of 10 specs
•••••••createErr (0x0,0x0)
•••

Ran 10 of 10 Specs in 0.000 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 966.672615ms
Test Suite Passed
```

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [x] add unit tests

### DIFF
--- a/pkg/services/hcloud/network/network.go
+++ b/pkg/services/hcloud/network/network.go
@@ -172,14 +172,35 @@ func (s *Service) findNetwork(ctx context.Context) (*hcloud.Network, error) {
 		return nil, nil
 	}
 
-	if len(networks[0].Subnets) > 1 {
-		configuredSubnet := s.scope.HetznerCluster.Spec.HCloudNetwork.SubnetCIDRBlock
-		firstSubnet := networks[0].Subnets[0]
+	if len(networks[0].Subnets) >= 1 {
+		// Allow existing subnets, but validate them
 
-		// Allow multiple subnets only if the first subnet matches the configured one. On attaching a server to a
-		// network the first subnet is used.
-		if firstSubnet.IPRange.String() != configuredSubnet {
-			return nil, fmt.Errorf("multiple subnets found and first subnet %s doesn't match the configured %s", firstSubnet.IPRange.String(), configuredSubnet)
+		// When multiple subnets of type "cloud" are present,
+		// there currently is no guarantee that resources will be created in the correct subnet
+		// The current solution is to allow only one subnet of type "cloud" and any subnet of type "vswitch" (technically limited to 1 though)
+		// Official support for multiple "cloud" subnets can be added when using hcloud-go version >= 2.30.0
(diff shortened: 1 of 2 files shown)


skuethe:
So, coming back to this: @fgrutsch has more or less already confirmed that their use-case matches, in their comment:
https://github.com/syself/cluster-api-provider-hetzner/pull/1633#issuecomment-3573995216

Also, pinging @ashish1099 again.

ashish1099:
@skuethe perfect, the above patch looks fine and we got it working with the last patch.

so all good and yes nice to have to run an air gap k8s cluster

guettli:
Lychee link checker failed. Flake, I merge.

PR #1881, opened 2026-03-17
:seedling: Handle missing providerID during hcloud remediation

### LINKED ISSUE #1880: HCloudRemediation returns an error when providerID is not found
/kind bug

In case the HCloudMachine has no providerID yet and is remediated, the HCloudRemediation returns an error:

> failed to find the server of unhealthy machine: failed to get serverID from providerID: nil providerID

This is due to the findServer logic. However, having no providerID is not a problem because it means that there was an issue with creating the server.

 The MachineHealthCheck should therefore just remove the HCloudMachine and not return an error.

guettli:
## Summary
- treat a missing hcloud providerID as an absent server during remediation
- add regression coverage for the service and reconciler paths

Fixes #1880


pkg/services/hcloud/remediation/remediation.go
```diff
@@ -160,6 +161,9 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 	serverID, err := s.scope.ServerIDFromProviderID()
 	if err != nil {
+		if errors.Is(err, hcloudutil.ErrNilProviderID) {
```
janiskemper, line 164: can you add a comment here? IMO we can say "a nil providerID means that the server hasn't been created yet"

guettli: @janiskemper done in d886aace6

PR #1883, opened 2026-03-17
:seedling: Harden envtest bootstrap in unit tests

guettli:
## Summary
This PR hardens the unit test harness around envtest startup and shutdown.

The main failure mode was that `hack/test-unit.sh` could pick up a pre-existing `KUBEBUILDER_ASSETS` from the caller environment. If that variable pointed to envtest binaries from a different checkout, the controller suite tried to start `etcd` from the wrong path and failed before the test suite even began.

Once envtest startup failed, the cleanup path could trigger follow-up panics:
- `NewTestEnvironment()` attempted to call `env.Stop()` while startup had already failed.
- `TestEnvironment.Stop()` assumed a non-nil cancel function.
- `AfterSuite` unconditionally touched the wait group even if `BeforeSuite` had never finished setting things up.

## Changes
- In `hack/test-unit.sh`, fail fast if `KUBEBUILDER_ASSETS` is already set, then resolve envtest assets explicitly for the current checkout.
- In `test/helpers/envtest.go`, avoid the extra `env.Stop()` call on startup failure and make `Stop()` tolerate partially initialized state.
- In `controllers/controllers_suite_test.go`, only stop the test environment and wait on the target-manager wait group when the environment was actually created.

## Result
With these changes, unit tests no longer accidentally reuse stale envtest paths, and startup failures produce a single clear error instead of cascading panics during shutdown.

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -164,15 +164,6 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 	startReconcile := time.Now()
 	// Always close the scope when exiting this function so we can persist any HCloudMachine changes.
 	defer func() {
-		if reterr != nil && errors.Is(reterr, hcloudclient.ErrUnauthorized) {
-			conditions.MarkFalse(hcloudMachine, infrav1.HCloudTokenAvailableCondition, infrav1.HCloudCredentialsInvalidReason, clusterv1.ConditionSeverityError, "wrong hcloud token")
-			// Don't return an error, no point retrying with invalid credentials.
-			reterr = nil
-			res = reconcile.Result{}
-		} else {
-			conditions.MarkTrue(hcloudMachine, infrav1.HCloudTokenAvailableCondition)
-		}
-
 		// the Close() will use PatchHelper to store the changes.
 		if err := machineScope.Close(ctx); err != nil {
 			res = reconcile.Result{}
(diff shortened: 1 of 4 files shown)


janiskemper (commented):
I'm not the expert here. @Dhairya-Arora01 maybe? @batistein ?

hack/test-unit.sh
```diff
@@ -18,17 +18,25 @@
 trap 'echo "Warning: A command has failed. Exiting the script. Line was ($0:$LINENO): $(sed -n "${LINENO}p" "$0")"; exit 3' ERR
 set -Eeuo pipefail
 
+if [[ -n ${KUBEBUILDER_ASSETS:-} ]]; then
+    echo "KUBEBUILDER_ASSETS=$KUBEBUILDER_ASSETS is set. Please do not do that."
```
Dhairya-Arora01: Please do not do that -> This variable is managed by this script; unset it and re-run.

guettli: yes, this makes sense. I updated the err msg.

hack/test-unit.sh
```diff
@@ -18,13 +18,20 @@
 trap 'echo "Warning: A command has failed. Exiting the script. Line was ($0:$LINENO): $(sed -n "${LINENO}p" "$0")"; exit 3' ERR
 set -Eeuo pipefail
 
+if [[ -n ${KUBEBUILDER_ASSETS:-} ]]; then
+    echo "KUBEBUILDER_ASSETS=$KUBEBUILDER_ASSETS is set. Please do not do that -> This variable is managed by this script; unset it and re-run."
```
Dhairya-Arora01, line 22: I meant replacing "Please do not do that" with "This variable is managed by this script; unset it and re-run"

PR #1887, opened 2026-03-18
:seedling: Do not reconcile on irrecoverable errors

### LINKED ISSUE #1885: Stop reconciling HCloudMachines with certain errors on creation
/kind feature

There are certain errors that we get from HCloud API on creation of a server that make it impossible / very unlikely that this will ever change:
- invalid_input (e.g. unsupported location for server type)
- resource_unavailable (e.g. "server location disabled")

Currently, we retry on these errors rather frequently - leading to many useless API calls. 

I propose to stop reconciling HCloudMachines that have a certain error type. They are only allowed to get deleted - otherwise we won't try to bring them into the cluster anymore. 

This has two aspects:
1. We can check based on the status if such an error is there and if so, return an empty result without error - instead of trying to create the server again.
2. We shouldn't allow the MachineHealthCheck to remediate the server, because this will lead to the same problem again.

Regarding the MachineHealthChecks in HCloudRemediation: They are meant to fix a machine by either rebooting it, or replacing it. The errors that we talk about here cannot be solved by either of the two measures. The only way is to replace the HCloudMachine with another one that is configured based on a different template. 

We should therefore also ignore in HCloudRemediation those HCloudMachines with these error types and pretend to the MHC that the machine is fine. That's of course not the case, but the MHC cannot do anything anyway.

There might be a rare case where a remediation might help, because e.g. the resource unavailability issue has been solved in the meantime on Hetzner side. For now, I think we can ignore this and therefore require manual interaction in that case. 

We could at one point think about allowing remediation e.g. for objects that are older than 24 hours. This would lead to some more API calls, but might solve the case where resource unavailability issues have been solved in the meantime.

Dhairya-Arora01:
Do not reconcile or remidiate on irrecoverable errors like `invalid_input` and `resource_unavailable` as reconciling again won't change the result for these errors and it will rather stuck in a loop.

Fixes #1885

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -48,6 +48,8 @@ const (
 	ServerTypeNotFoundReason = "ServerTypeNotFound"
 	// ServerCreateFailedReason indicates that server could not get created.
 	ServerCreateFailedReason = "ServerCreateFailedReason"
+	// ServerCreateFailedIrrecoverableErrorReason indicates that server creation failed with an irrecoverable error.
+	ServerCreateFailedIrrecoverableErrorReason = "ServerCreateFailedIrrecoverableError"
 )
 
 const (
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -142,6 +142,21 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		// otherwise create server.
 		server, err = s.createServer(ctx)
 		if err != nil {
+			// Terminal errors like invalid_input (e.g. unsupported location for server type)
+			// or resource_unavailable (e.g. server location disabled) will never succeed on retry.
+			// Mark the machine as irrecoverably failed and stop reconciling.
+			if hcloud.IsError(err, hcloud.ErrorCodeInvalidInput) || hcloud.IsError(err, hcloud.ErrorCodeResourceUnavailable) {
+				conditions.MarkFalse(
+					s.scope.HCloudMachine,
+					infrav1.ServerCreateSucceededCondition,
+					infrav1.ServerCreateFailedIrrecoverableErrorReason,
+					clusterv1.ConditionSeverityError,
+					"%s",
+					err.Error(),
+				)
+				return reconcile.Result{}, nil
+			}
+
 			if errors.Is(err, errServerCreateNotPossible) {
 				return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 			}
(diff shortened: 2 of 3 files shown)


api/v1beta1/conditions_const.go
```diff
@@ -48,6 +48,8 @@ const (
 	ServerTypeNotFoundReason = "ServerTypeNotFound"
 	// ServerCreateFailedReason indicates that server could not get created.
 	ServerCreateFailedReason = "ServerCreateFailedReason"
+	// ServerCreateFailedIrrecoverableErrorsReason indicates that server creation failed with an irrecoverable error.
```
janiskemper: I would name it IrrevocableError not Errors

pkg/services/hcloud/server/server.go
```diff
@@ -142,6 +142,25 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		// otherwise create server.
 		server, err = s.createServer(ctx)
 		if err != nil {
+			// Terminal errors like invalid_input (e.g. unsupported location for server type)
+			// or resource_unavailable (e.g. server location disabled) will never succeed on retry.
+			// Mark the machine as irrecoverably failed and stop reconciling
+			if hcloud.IsError(err, hcloud.ErrorCodeInvalidInput) || hcloud.IsError(err, hcloud.ErrorCodeResourceUnavailable) {
+				s.scope.SetError(
```
janiskemper: will this trigger the MHC or lead to removing the machine? I'm not sure

pkg/services/hcloud/server/server.go
```diff
@@ -142,6 +142,25 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+					fmt.Sprintf("server creation failed with irrecoverable error: %s", err.Error()),
+					capierrors.CreateMachineError,
+				)
+				conditions.MarkFalse(
+					s.scope.HCloudMachine,
+					infrav1.ServerCreateSucceededCondition,
+					infrav1.ServerCreateFailedIrrecoverableErrorsReason,
+					clusterv1.ConditionSeverityError,
+					"irrecoverable error creating server: %s",
```
janiskemper: i would remove the "irrecoverable error creating server". that would just make it impossible to read the actual message when listing resources.

pkg/services/hcloud/server/server.go
Dhairya-Arora01: From CAPI book

```quote
In case an infrastructure provider reports that a InfraMachine resource is in a state that cannot be recovered (terminal failure) by setting status.failureReason and status.failureMessage as defined by the deprecated v1beta1 contract, the “core” Machine controller will surface those info in the corresponding fields in the Machine’s status.deprecated.v1beta1 struct.

However, those info won’t have any impact on the Machine lifecycle as before (the Machine controller won’t consider the presence of status.failureReason and status.failureMessage info as “terminal failures”; similarly, the MachineHealthCheck controller won’t consider the presence of status.failureReason and status.failureMessage to determine when a Machine needs remediation).
```

PR #1888, opened 2026-03-18
:seedling: Stop reconciling on unauthorized error

### LINKED ISSUE #1886: Stop reconciling HCloudMachines if they show unauthorized
/kind feature

If an HCloudMachine shows unauthorized, which we store in a condition, then we can stop reconciling. The MachineHealthChecks will kill the machine eventually (at least if https://github.com/syself/cluster-api-provider-hetzner/issues/1880 is solved). Until it is killed and recreated, there is no point in retrying API calls. 

We can therefore check for the condition and if it exists, just stop requeuing. 

This is similar to https://github.com/syself/cluster-api-provider-hetzner/issues/1885, but here I wouldn't block MachineHealthChecks, so that the respective Machines would be remediated by the MHC and newly created again.

Dhairya-Arora01:
If the hcloudmachine as HCloudCredentialsInvalid condition, stop reconciling it. The MHC should kill the machine, otherwise reconciling this would not change the result and would result in Hetzner API rate-limit.

Fixes #1886

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -100,6 +100,15 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 
 	log = log.WithValues("Machine", klog.KObj(machine))
 
+	// If the HCloudMachine has an unauthorized / HCloudCredentialsInvalid error, stop reconciling.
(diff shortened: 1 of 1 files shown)


janiskemper (commented):
LGTM. Can you add tests though?

PR #1889, opened 2026-03-19
:seedling: Stop reconciling on irrecoverable and unauthorized errors

Dhairya-Arora01:
HCloudMachine should not be reconciled further if it encounters an irrecoverable error such as `invalid_input`, `resource_unavailable` OR
`unauthorized` error which occurs when a wrong HCloudToken is used to call the Hetzner API.

Additionally, the remediation should be skipped as well for irrecoverable errors.

This is inline with the changes made to the main branch via #1887 and #1888

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -92,15 +92,6 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 
 	log = log.WithValues("HCloudMachine", klog.KObj(hcloudMachine))
 
-	// If the HCloudMachine has an unauthorized / HCloudCredentialsInvalid error, stop reconciling.
-	// There is no point in retrying API calls with an invalid token.
-	// The MachineHealthCheck will eventually remediate the machine by deleting and recreating it.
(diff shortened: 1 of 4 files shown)


controllers/hcloudmachine_controller.go
```diff
@@ -92,6 +92,15 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 
 	log = log.WithValues("HCloudMachine", klog.KObj(hcloudMachine))
 
+	// If the HCloudMachine has an unauthorized / HCloudCredentialsInvalid error, stop reconciling.
+	// There is no point in retrying API calls with an invalid token.
+	// The MachineHealthCheck will eventually remediate the machine by deleting and recreating it.
+	if conditions.IsFalse(hcloudMachine, infrav1.HCloudTokenAvailableCondition) &&
```
janiskemper: I think that we concluded today to not do that, but to instead just not return an error and stop reconciling. But not checking whether the condition is set

janiskemper (approved):

PR #1891, opened 2026-03-20
:bug: Handle unauthorized errors before rate limits

### LINKED ISSUE #1890: Condition indicating that API token is wrong gets removed when rate limit exceeded
/kind bug

There are certain cases where our main reconcile function of HCloudMachine doesn't return an error, even if the Hetzner API gave us an error, e.g. when we are hitting rate limits. The reason being that we don't want to end up in a loop where we hit the API again and again. However, we have currently a defer statement that checks for unauthorized errors. In case we don't return an error anymore, e.g. when rate limit exceeded, then in the defer statement we conclude that the unauthorized error is gone - which is not correct.

We have two options:
a) handle this specific case in some way
b) move the handling of the unauthorized error to the respective function.

I vote for using b). It is currently very hard to understand and the only advantage is that whenever we add API calls in the reconcile loop, we don't need to worry about the unauthorized error handling. It would be a lot cleaner if we just handled the unauthorized error once in the first place where we do an API call and that's it. 

We can check all relevant controllers for similar cases.

Dhairya-Arora01:
Handle unauthorized errors before rate limits otherwise as per the current setup if the unauthorized error is set on an HCloudMachine and we encounter a rate limit exceeded error then in the defer statement we remove the unauthorized error which is wrong.

This commit is a bug fix.

Here, we check if the unauthorized error is check in the following methods:
- `getServerType()`: This is called before creating the server i.e. when provider id is not set. 
- `findServer()`: This is called when provider id is set.

In both the methods we set the condition `HCloudTokenAvailable` condition to false. And at the call site (not the immediate call sites, but the call site for their parent function in the reconcile function) of these functions we check if it is an `unauthorized` error. If so, do not reque or return an error as there is no point in retrying with wrong HCloud token.

Fixes #1890

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -40,6 +40,7 @@ import (
 	"github.com/syself/cluster-api-provider-hetzner/pkg/scope"
 	secretutil "github.com/syself/cluster-api-provider-hetzner/pkg/secrets"
 	sshclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/baremetal/client/ssh"
+	hcloudclient "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/client"
 	hcloudutil "github.com/syself/cluster-api-provider-hetzner/pkg/services/hcloud/util"
 	"github.com/syself/cluster-api-provider-hetzner/pkg/utils"
 )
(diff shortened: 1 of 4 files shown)


janiskemper (commented):
is it possible to add tests with using a mock of HCloud API to ensure that the right condition is set in the two cases?

janiskemper (commented):
thanks! Can you also add a test that shows that we keep this condition when we return a rate limit?

pkg/services/hcloud/server/server.go
```diff
@@ -1204,8 +1204,11 @@ func (s *Service) getServerImage(ctx context.Context, imageName string) (*hcloud
 			return nil, err
 		}
 
-		// Otherwise, mark HCloudTokenAvailableCondition as true.
-		conditions.MarkTrue(s.scope.HCloudMachine, infrav1.HCloudTokenAvailableCondition)
+		// Only mark HCloudTokenAvailableCondition as true if the token wasn't previously marked as invalid.
```
janiskemper: does this make sense at all? We are not totally sure if it is true or not - that would depend on the error. With certain errors we can be sure that we are not sure (and should keep the previous state). 

Instead of adding this code, I would actually argue it is better to not do it.

janiskemper (approved):

PR #1893, opened 2026-03-23
:seedling: Replace Update with Patch for HBMH

### LINKED ISSUE #1587: Replace HetznerBareMetalHost "update" call with "patch" to avoid conflict errors
/kind bug

We observed, that having a conflict error while deprovisioning a HetznerBareMetalHost is very common, as the HetznerBareMetalMachine controller also updates it. This lead to situations, usually under heavy load, where the object completely stopped reconciling. 

It is weird that this happened and not really the fault of our "requeue" mechanism, but still, we could avoid this altogether, if we manage to get rid of "update". 

We can avoid errors that lead to the HetznerBareMetalHost reconciling multiple times in the same state, or even stopping to reconcile at all, if we use the patch logic.

I don't remember why we used "update" instead of "patch". There was a reason. That reason might be irrelevant by now, as we took this decision years ago. We should still ensure that we don't run into problems.

Dhairya-Arora01:
Replace `Update()` with `Patch()` in HBMH controller to prevent conflicts.

We expect the reason for the conflict as follows:
1. HBMM (uses patchhelper) and HBMH (uses update) controllers Get the Host resource, let's say with resourceVersion: 1000

2. HBMM removes fields from `host.spec` ```go host.spec.status.installImage = nil host.spec.status.userData = nil host.spec.status.sshStatus = nil ``` Upon patching the spec resourceVersion is incremented to 1001.

3. HBMH, still has the stale copy (with resourceVersion: 1000), updates the status.conditions using client.Updat - which is an UPDATE request, API server sees the request with
resourceVersion: 1000 and returns an error because the current resourceVersion is 1001. Also since the UPDATE tries to **replace** the main object in storage completely, the conflict should happen on the changed spec.status thing.

Fixes #1587 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


controllers/hetznerbaremetalhost_controller.go
```diff
@@ -108,71 +106,29 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 		return reconcile.Result{}, err
 	}
 
-	// ----------------------------------------------------------------
-	// Start: avoid conflict errors. Wait until local cache is up-to-date
```
guettli, line 112: @Dhairya-Arora01 Why was `Wait until local cache is up-to-date` removed? This is not related to the current issue: Use Patch instead of Delete.

guettli (changes requested):
There are still several Update() calls:

```console 
controllers/hetznerbaremetalhost_controller.go:201:11:      err := r.Update(ctx, bmHost)
controllers/hetznerbaremetalhost_controller.go:213:11:      err := r.Update(ctx, bmHost)
controllers/hetznerbaremetalhost_controller.go:344:12:          err := r.Update(ctx, bmHost)
controllers/hetznerbaremetalhost_controller.go:357:15:          if err := r.Update(ctx, bmHost); err != nil {
```

---

Additionally, this gets lost:

```go 
host.Spec.Status.LastUpdated = &t
```

guettli (approved):
lgtm.

janiskemper (approved):

PR #1898, opened 2026-03-24
:seedling: Stop reconciling if associated host is not found for hbmm

### LINKED ISSUE #1897: Set a condition, stop reconciling and mark `HetznerBaremetalMachine` for remediation if the corresponding host is not found
While reconciling a `HetznerBaremetalMachine` object, if the corresponding `HetznerBaremetalHost` no longer exists. Then we should set a condition that the corresponding host is not found and stop reconciling the object.

Additionally, we should also mark the machine for remediation by setting the annotation `cluster.x-k8s.io/remediate-machine`. 

> [!TIP]
> This can be done using the function [`SetErrorAndRemediate()`](https://github.com/syself/cluster-api-provider-hetzner/blob/7b9f7e4595787e3cdf9f7dcc4129fc2d0c485c32/pkg/scope/machine.go#L130-L137)

---

Changes should be made on top of release branch `v1.1.x`

Dhairya-Arora01:
If associated `HetznerBaremetalHost` object no longer exists for a `HetznerBaremetalMachine`, hbmm controller does the following:
- Set condition `HostReady` to true with reason `HostNotFound`.
- Set annotation `cluster.x-k8s.io/remediate-machine` on the CAPI machine to mark it for remediation.
- Do not reque or return an error to avoid further reconciliation.

Fixes #1897

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hetznerbaremetalmachine_controller_test.go
+++ b/controllers/hetznerbaremetalmachine_controller_test.go
@@ -763,7 +763,7 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 		})
 	})
 
-	Context("Tests with hosts that got deleted later", func() {
+	Context("Tests with hosts which get deleted later", func() {
 		var (
 			host        *infrav1.HetznerBareMetalHost
 			hostKey     client.ObjectKey
--- a/controllers/hetznerbaremetalmachine_controller_test.go
+++ b/controllers/hetznerbaremetalmachine_controller_test.go
@@ -763,6 +763,106 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 		})
 	})
 
+	Context("Tests with hosts that got deleted later", func() {
+		var (
+			host        *infrav1.HetznerBareMetalHost
+			hostKey     client.ObjectKey
+			capiMachine *clusterv1.Machine
+			bmmKey      client.ObjectKey
+		)
+
+		BeforeEach(func() {
+			host = helpers.BareMetalHost(
+				hostName,
+				testNs.Name,
+				helpers.WithRootDeviceHintWWN(),
+				helpers.WithHetznerClusterRef(hetznerClusterName),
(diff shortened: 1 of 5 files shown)


pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -126,7 +139,21 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 	}
 
 	// set providerID if necessary
-	if err := s.setProviderID(ctx); err != nil {
+	err = s.setProviderID(ctx)
+	if apierrors.IsNotFound(err) {
```
janiskemper: is this necessary? To me it seems not, because we would always catch this and return in the "update" function above.

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -52,6 +52,10 @@ func NewService(scope *scope.BareMetalRemediationScope) *Service {
 // Reconcile implements reconcilement of HetznerBareMetalRemediations.
 func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	host, err := host.GetAssociatedHost(ctx, s.scope.Client, s.scope.BareMetalMachine)
+	if apierrors.IsNotFound(err) {
```
janiskemper: I think it's slightly more efficient to do this within "err != nil" right?

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -763,6 +763,106 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 		})
 	})
 
+	Context("Tests with hosts that got deleted later", func() {
```
janiskemper: *get

janiskemper (approved):

PR #1901, opened 2026-03-26
Fix: Ensure conntrack package installed

gcaracuel:

Recently, Hetzner's Ubuntu 24.04 images have been observed to lack the conntrack utility by default. This absence causes kubeadm to fail during various critical steps, as conntrack is a fundamental dependency for network connection tracking required by Kubernetes. This leads to provisioning failures for clusters utilizing this specific OS image.

This PR ensures `conntrack` package is installed as part of the pre-tasks

**TODOs**:

- [x] squash commits
- [x] include documentation
- [x] add unit tests


guettli:
@gcaracuel please take a look at an issue create by my colleague Dhairya [Kubelet in k8s v1.31.14 no longer depends on `conntrack` but still requires it at runtime · Issue #3259 · kubernetes/kubeadm](https://github.com/kubernetes/kubeadm/issues/3259)

Afaik conntrack is no longer needed.

gcaracuel:
> Kubelet in k8s v1.31.14 no longer depends on `conntrack` but still requires it at runtime · Issue #3259 · kubernetes/kubeadm

Still needed on older versions that's how I faced this issue and those version are still in the CAPH docs, in fact the quick start still uses 1.31.6.

Faced this issue testing and upgrade to 1.35.x so anybody with a CAPH cluster prior to 1.32 may face this issues and need to change that postInstallScript. If we are ok with this then I will close the PR.

PR #1781, opened 2026-01-22
:seedling: Set v1.1.x to main branch, and release (Overview)

guettli:
Previous PR: [Step 2: support upstream hcloud-ccm secret/key conventions (II)  #1920](https://github.com/syself/cluster-api-provider-hetzner/pull/1920)
Next PR: #1775 (first beta2 PR)

>[!NOTE]
> This is the last PR without beta2.
> We should merge that to main (or reset main), and the release that to our commmercial customers.
> After that we can work on beta2.

At the end of 2025 the main was reset to reflect the last release.

After this PR is merged to v1.1.x we plan to reset main again, and release caph v1.1.x

# Tasks

- Update README.md and Pull-Request-Template to reflect the changes.
- Fix e2e tests.
- Maybe update CAPI to a newer version (which still supports beta1)
- Reset main
- Release
- Deploy
- Work on beta2

# Commits behind main

These commits of main are not in v1.1.x branch.

Status explains whether the change is already covered in `v1.1.x`, superseded there, or tracked by a follow-up backport PR.

"✔" in first column means checked by Thomas.

"👀" in first column means in review.

| Status | Status explanation | main | v1.1.x | Description |
| --- | --- | --- | --- | --- |
| ✔  | merged | [`e08b791e / #1926`](https://github.com/syself/cluster-api-provider-hetzner/pull/1926) | [`#1931`](https://github.com/syself/cluster-api-provider-hetzner/pull/1931) | Ignores the Robot `server.Rescue` capability flag during bare-metal preparation because the API may omit it even when rescue activation still works. |
| ✔  | merged | [`24f77fcc / #1930`](https://github.com/syself/cluster-api-provider-hetzner/pull/1930) | [`#1932`](https://github.com/syself/cluster-api-provider-hetzner/pull/1932) | `v1.1.x` does not have the early `HCloudCredentialsInvalid` reconcile short-circuit from `main`, so that part is not needed here. The delete-path nil-`ProviderID` guard is still needed on `v1.1.x` |
| ✔  | merged | [`a271b443 / #1909`](https://github.com/syself/cluster-api-provider-hetzner/pull/1909) | [`#1908`](https://github.com/syself/cluster-api-provider-hetzner/pull/1908) | Serializes bare-metal e2e jobs globally so different PRs cannot use the shared `ci-pool` at the same time. Equivalent change already landed in `v1.1.x`. |
[...]

PR #1903, opened 2026-03-27
:seedling: Backport hcloud rate-limit handling to v1.1.x

guettli:
## Summary
- stop treating hcloud rate-limit failures during `findServer()` as if the server disappeared
- let reconcile requeue while a not-yet-ready machine is rate-limited
- keep deletion paths using the existing rate-limit helper
- add regression tests for wrapped `findServer()` rate-limit errors and deletion handling

## Backport Source
- Backport of #1857

## Why
Without this backport, `v1.1.x` can mis-handle a rate-limited `findServer()` call and fall into the missing-server path.

## Overview PR

#1781


pkg/services/hcloud/server/server.go
```diff
@@ -111,6 +111,13 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 			if errors.Is(err, hcloudclient.ErrUnauthorized) {
 				return reconcile.Result{}, nil
 			}
+			if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
+				if !s.scope.HCloudMachine.Status.Ready {
+					hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
```
Dhairya-Arora01, line 116: In `findServer` method we already handle the rate limit exceeded error. Maybe we can reque with delay here.

guettli: @Dhairya-Arora01 can you please elaborate. What is your desired solution?

Dhairya-Arora01: Ah you can ignore that comment - I was wrong.

Dhairya-Arora01 (approved):

PR #1906, opened 2026-03-27
:seedling: Use IgnoreNotFound when fetching related objects in controllers

### LINKED ISSUE #1894: Missing "IgnoreNotFound" in "get object" functionality
/kind bug

We have multiple places where we get related objects, e.g. the Machine and HetznerCluster objects for a HCloudMachine object. We should always ignore not found errors to not repeatedly reconcile in cases of dangling objects. These other objects should always be there, and if they are not, some user action probably caused the inconsistency. The controller cannot do anything.

Go through all controllers and ensure that we do this: https://github.com/syself/cluster-api-provider-hetzner/blob/774ec79de9da772e4db47c87b2b709ce4c4410c9/controllers/hcloudmachine_controller.go#L86

Let's do this for the v1.1.x branch.

abdullah599:
Uses `client.IgnoreNotFound` when fetching related objects (Machine, Cluster, HetznerCluster, HetznerBareMetalMachine) across all controllers. This prevents repeated reconciliation loops when related objects are missing due to dangling references (the controller cannot fix these inconsistencies, so it should stop reconciling instead of retrying indefinitely).

Fixes #1894

Base branch is `v1.1.x`.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachinetemplate_controller.go
+++ b/controllers/hcloudmachinetemplate_controller.go
@@ -102,7 +102,7 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
 	if cluster == nil {
 		log.Info(fmt.Sprintf("%s is missing ownerRef to cluster %s/%s",
 			machineTemplate.Kind, machineTemplate.Namespace, machineTemplate.Name))
-		return reconcile.Result{Requeue: true}, nil
+		return reconcile.Result{}, nil
 	}
 	machineTemplate.Status.OwnerType = cluster.Kind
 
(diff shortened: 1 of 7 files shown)


controllers/hcloudmachinetemplate_controller.go
```diff
@@ -96,8 +96,11 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
 
 	var cluster *clusterv1.Cluster
 	cluster, err = util.GetOwnerCluster(ctx, r, machineTemplate.ObjectMeta)
-	if err != nil || cluster == nil {
-		log.Info(fmt.Sprintf("%s is missing ownerRef to cluster or cluster does not exist %s/%s",
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	if cluster == nil {
```
Dhairya-Arora01, line 102: I think if cluster == nil i.e. the owner ref is not set yet, even then we can return without reque because once owner ref is set (i.e. there is an update) , the object will be reconciled again.

Similar at other places
- GetOwnerMachine

Dhairya-Arora01 (approved):
LGTM, @janiskemper plz take a look

janiskemper (approved):
Lgtm. I assume we didn't forget anything

PR #1907, opened 2026-03-27
:seedling: Stop reconciling HBMH on `unauthorized` error

### LINKED ISSUE #1895: Stop reconciling HetznerBaremetalHost on `unauthorized` error
Currently while provisioning a HetznerBaremetalHost, if wrong robot credentials are configured by the user then the HBMH's provisioning gets stuck without the reason being surfaced to the user through conditions. 

**Solution**:
When we recieve the error `unauthorized` on making the first call to Robot API in an HBMH's phase - we should set a condition indicating that the provided robot credentials are invalid and should stop reconciling it further.

---

Changes should be made on top of release branch `v1.1.x`

Dhairya-Arora01:
If we recieve the error `unauthorized` on making the first call to Robot API in an HBMH's phase then we should mark condition `CredentialsAvailable` to false with reason `RobotCredentialsInvalid` and stop reconciling.

Fixes #1895

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hetznerbaremetalhost_controller_test.go
+++ b/controllers/hetznerbaremetalhost_controller_test.go
@@ -835,6 +835,95 @@ var _ = Describe("HetznerBareMetalHostReconciler - missing secrets", func() {
 			}, timeout).Should(BeTrue())
 		})
 	})
+
+	Context("Wrong Robot Credentials - unauthorized", func() {
+		BeforeEach(func() {
+			host = helpers.BareMetalHost(
+				hostName,
+				testNs.Name,
+				helpers.WithHetznerClusterRef(hetznerClusterName),
+				helpers.WithRootDeviceHintWWN(),
+			)
+			Expect(testEnv.Create(ctx, host)).To(Succeed())
+
+			key = client.ObjectKey{Namespace: testNs.Name, Name: host.Name}
+
+			hetznerSecret = getDefaultHetznerSecret(testNs.Name)
+			Expect(testEnv.Create(ctx, hetznerSecret)).To(Succeed())
+
+			rescueSSHSecret = helpers.GetDefaultSSHSecret("rescue-ssh-secret", testNs.Name)
+			Expect(testEnv.Create(ctx, rescueSSHSecret)).To(Succeed())
+
+			osSSHSecret = &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "os-ssh-secret",
+					Namespace: testNs.Name,
+				},
+				Data: map[string][]byte{
+					"private-key": []byte("os-ssh-secret-private-key"),
+					"sshkey-name": []byte("ssh-key-name"),
+					"public-key":  []byte("my-public-key"),
+				},
+			}
+			Expect(testEnv.Create(ctx, osSSHSecret)).To(Succeed())
+
+			robotClient.ExpectedCalls = nil
+			robotClient.On("GetBMServer", mock.Anything).Return(nil, models.Error{
+				Code:    models.ErrorCodeUnauthorized,
+				Message: "You are not authorized - wrong RobotCredentials",
+			}).Once()
+		})
+
+		It("should set CredentialsAvailable condition to false if Robot API returned unauthorized", func() {
+			By("making the Robot client return an unauthorized error")
+			Eventually(func() bool {
+				return isPresentAndFalseWithReason(key, host, infrav1.RobotCredentialsAvailableCondition, infrav1.RobotCredentialsInvalidReason)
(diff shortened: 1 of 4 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -331,12 +350,31 @@ func (s *Service) ensureSSHKey(sshSecretRef infrav1.SSHSecretRef, sshSecret *cor
 	}
 	hetznerSSHKeys, err := s.scope.RobotClient.ListSSHKeys()
 	if err != nil {
+		// If Robot API returned "unauthorized" error - mark condition CredentialsAvailable as false
```
janiskemper: do we need it here as well? And below? Or just in the first occasion where we call the API?

janiskemper: my proposal would be that we add it once at the beginning and one while deprovisioning and otherwise remove it. I had a very hard time checking which pattern you used, e.g. if you always added it on every API call. I think that's not necessary. The code would look cleaner if we add it only on these two places. WDYT?

Dhairya-Arora01: @janiskemper , the pattern that I followed is - check at the first API call of every provisioning state if we get an unauthorized error. 
Because what if the robot credentials are changed in between the provisioning states ? We have the hbmh reconciliation logic as state machine where we first try to deduce the provisioning state and then execute the logic of that provisioning state. That is why I think we need to check for unauthorized error on first API call in every provisioning state.

janiskemper: problem I see is that it is also in utility functions right? I assumed that you did that, but I would still say that it is very hard to understand and especially to maintain. I think that the chance that the credentials are changed during provisioning is small enough so that it shouldn't matter.

Of course, it would be better to have it in every step. So do you have some idea how to make that transparent?

Dhairya-Arora01: Hmm.... if the chances are very less that the credentials will change between provisioning states then


Lets just add it to these handlers

- [`actionPreparing`](https://github.com/syself/cluster-api-provider-hetzner/blob/b2d9c777a3711c6b2672654b25659d5d2f2cde1e/pkg/services/baremetal/host/host.go#L181) : Which is invoked as the first step of provisioning i.e. before registering. 
- [`actionDeprovisioning`](https://github.com/syself/cluster-api-provider-hetzner/blob/b2d9c777a3711c6b2672654b25659d5d2f2cde1e/pkg/services/baremetal/host/host.go#L2355) : Which is invoked on deprovisioning of the server.

janiskemper: okay

janiskemper (commented):
are you sure that we should re-use the existing condition? It seems risky to me honestly.. With some edge cases based on existing usage and future usage.. Maybe we should have a separate conditon.. 

I do agree that the name "CredentialsAvailable" is too generic for current usage, but we can change that in v1beta2. WDYT?

Dhairya-Arora01:
@janiskemper the only other place where we set the `RobotCredentialsInvalid` condition is - when in the secret referenced in `HetznerCluster.Spec.HetznerSecret.Name` the specified robot credentials `username` and `password` are empty/not set.

This is the same as for HCloudTokenInvalid for hcloud machine.

We made the same changes there. Therefore I don't see a risk.

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -835,6 +835,93 @@ var _ = Describe("HetznerBareMetalHostReconciler - missing secrets", func() {
...
+				Data: map[string][]byte{
+					"private-key": []byte("os-ssh-secret-private-key"),
+					"sshkey-name": []byte("ssh-key-name"),
+					"public-key":  []byte("my-public-key"),
+				},
+			}
+			Expect(testEnv.Create(ctx, osSSHSecret)).To(Succeed())
+		})
+
+		It("should set CredentialsAvailable condition to false if Robot API returned unauthorized", func() {
```
janiskemper: is this test not gonna panic? Because we first apply the objects and then add the mock later?

pkg/services/baremetal/host/host.go
```diff
@@ -183,6 +183,23 @@ func (s *Service) actionPreparing(ctx context.Context) actionResult {
 
 	server, err := s.scope.RobotClient.GetBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID)
 	if err != nil {
+		// If Robot API returned "unauthorized" error - mark condition CredentialsAvailable as false
```
janiskemper: let's add this also to actionProvisioned

janiskemper (approved):

PR #1910, opened 2026-03-27
[codex] Backport #1752 to v1.1.x

guettli:
## What changed
Backports the fix from #1752 to `v1.1.x` by defaulting SSH ports when copying the machine `SSHSpec` into the baremetal host status.

The effective change on `v1.1.x` is limited to `pkg/services/baremetal/baremetal/baremetal.go`.

## Why it changed
If the SSH port is not specified, `PortAfterInstallImage` can remain `0`. On `v1.1.x`, `ensure-provisioned` then tries to use that zero port and provisioning never finishes.

## Impact
When the SSH port is omitted:
- `PortAfterInstallImage` now falls back to `22`
- `PortAfterCloudInit` now falls back to `PortAfterInstallImage`

This keeps the host status usable for the provisioning flow on `v1.1.x`.

## Validation
- `go test ./pkg/services/baremetal/baremetal ./pkg/services/baremetal/host`

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -65,6 +65,8 @@ const (
 
 	// FailureMessageMaintenanceMode indicates that host is in maintenance mode.
 	FailureMessageMaintenanceMode = "host machine in maintenance mode"
+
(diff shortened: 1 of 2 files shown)


guettli:
not needed.

PR #1916, opened 2026-04-02
[codex] show deduped CCM logs during e2e

guettli:
## What changed
This updates the continuous e2e status logger so workload-cluster CCM logs are printed during the run, not only collected afterward.

The logger now:
- discovers CCM deployments in `kube-system`
- reads logs from all matching pods and all containers
- prints each formatted line only once across polling cycles
- highlights lines containing `E####`

I also added focused helper tests for CCM deployment matching, container enumeration, deduplication, and highlighting.

## Why
The existing continuous e2e logging showed cluster status and CCM deployment metadata, but not the CCM logs themselves. That made CCM-related failures harder to diagnose while the test was still running.

## Impact
Developers running or reviewing e2e runs get CCM errors in the live stream with much less repeated noise.

## Validation
- `go test ./test/e2e -run TestDoesNotExist`
- `go test ./test/e2e -run 'Test(IsCCMDeploymentName|CCMContainerNames|CollectNewCCMLogLinesDedupesAndHighlights)$'`

## Notes
I did not run the full e2e suite locally because it requires the normal `e2e.config` setup.


guettli:
Superseded by #1917. The earlier PR was based on unrelated branch history and included a large unrelated diff.

PR #1918, opened 2026-04-02
🌱 Add CAPI version verification and align e2e config

guettli:
## What changed
- add `make verify-capi-version` backed by `hack/check-capi-version.sh`
- include the new target in `make verify`
- run the CAPI version check in `pr-verify`
- align the stale Cluster API provider URLs in `test/e2e/config/hetzner.yaml` with the version from `go.mod`

## Why
The repo already had a script to detect Cluster API version drift, but it was not enforced in CI. That allowed the e2e config to keep referencing `v1.8.10` while `go.mod` had already moved to `v1.10.9`.

## Impact
Pull requests now fail when CAPI-related versions drift from `go.mod`, and the current e2e config is back in sync.

## Validation
- `make verify-capi-version`
- `make lint-yaml-ci`
- `git diff --check`

### DIFF
--- a/hack/check-capi-version.sh
+++ b/hack/check-capi-version.sh
@@ -67,9 +67,14 @@ if ! awk \
     -v desired_major="${DESIRED_MAJOR}" \
     -v desired_minor="${DESIRED_MINOR}" \
     '
-        match($0, /major:[[:space:]]*([0-9]+)/, captures) { major = captures[1] }
-        match($0, /minor:[[:space:]]*([0-9]+)/, captures) {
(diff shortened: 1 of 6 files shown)


batistein (approved):

hack/check-capi-version.sh
```diff
@@ -1,4 +1,18 @@
 #!/usr/bin/env bash
+# Copyright 2026 The Kubernetes Authors.
```
janiskemper, line 2: @guettli here we introduced a duplicate boilerplate.

PR #1914, opened 2026-04-01
:seedling: Rename workload secret reconciliation helper

guettli:
Previous PR: [:seedling: Support upstream providerId format (Step1) #1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850)

Next PR: [:sparkles: Step 2: support upstream hcloud-ccm secret/key conventions #1703](https://github.com/syself/cluster-api-provider-hetzner/pull/1703)

This PR extracts the `reconcileTargetSecret` to `reconcileWorkloadClusterSecrets` rename from [#1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850), following review feedback to keep the ProviderID migration PR focused.

# Changes

- rename `reconcileTargetSecret` to `reconcileWorkloadClusterSecrets`
- rename local variables in that helper so workload-cluster and management-cluster scope are clearer
- keep behavior unchanged

This is a refactor-only PR. No functional change is intended.

PR #1850, opened 2026-03-09
:seedling: Support upstream providerId format (Step1)

guettli:
Next PR: [:seedling: Step 2: support upstream hcloud-ccm secret/key conventions (II) by guettli #1920](https://github.com/syself/cluster-api-provider-hetzner/pull/1920)

This PR is step 1 of the work to support the upstream [hcloud CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager/) for bare-metal clusters.

Until now, CAPH relied on our Syself fork, the [Syself "hetzner" CCM](https://github.com/syself/hetzner-cloud-controller-manager/). The main difference relevant here is the ProviderID format for bare-metal nodes:

- legacy format: `hcloud://bm-NNNN`
- upstream format: `hrobot://NNNN`

The upstream CCM now supports bare metal, so we can gradually phase out our fork. This PR contains the basic, non-opinionated changes needed for that migration. The more opinionated compatibility changes stay in [#1703](https://github.com/syself/cluster-api-provider-hetzner/pull/1703).

If you want CAPH to create new bare-metal ProviderIDs in the upstream format, set the annotation `capi.syself.com/use-hrobot-provider-id-for-baremetal: "true"` on the `HetznerCluster`.

Important: CAPH and the CCM must agree on the ProviderID format. If the annotation is enabled, use either the upstream HCloud CCM or the Syself CCM with `--use-hrobot-provider-id-for-baremetal`. Otherwise newly created nodes will not match and node bring-up will fail.

Existing nodes keep their current ProviderID. This change only affects newly created bare-metal nodes. HCloud machines are not affected.

# Changes

- add the `capi.syself.com/use-hrobot-provider-id-for-baremetal` annotation to `HetznerCluster`
- create bare-metal `ProviderID` values as `hrobot://NNNN` when that annotation is enabled, otherwise keep the legacy `hcloud://bm-NNNN` format
- validate that `HetznerBareMetalMachine.spec.providerID` is immutable once it has been set
- update the CSR controller to recognize both bare-metal ProviderID formats
- tighten HCloud ProviderID parsing so HCloud machines still require the `hcloud://NNNN` format
- update docs and cluster templates to describe the migration and default templates to the new annotation
- switch the bare-metal e2e CCM manifests to the upstream HCloud CCM and adjust tolerations/rendered YAML accordingly

# Out Of Scope

[...]

PR #1920, opened 2026-04-02
:seedling: Step 2: support upstream hcloud-ccm secret/key conventions (II)

guettli:
Next PR: [:seedling: Set v1.1.x to main branch, and release. #1781](https://github.com/syself/cluster-api-provider-hetzner/pull/1781)

(merged Previous PR: [:seedling: Support upstream providerId format (Step1) #1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850)) 

This PR is step 2 of the work to support the upstream [hcloud CCM](https://github.com/hetznercloud/hcloud-cloud-controller-manager/) for bare-metal clusters.

The core ProviderID migration work was moved to [#1850](https://github.com/syself/cluster-api-provider-hetzner/pull/1850). The refactor-only helper rename now lives in [#1914](https://github.com/syself/cluster-api-provider-hetzner/pull/1914).

This PR contains the follow-up changes that make the upstream CCM easier to use in practice.

# Changes

- mirror the HCloud token in the workload-cluster secret under both `token` and `hcloud`, in addition to the configured key, so both CCM variants can read it
- if `spec.hetznerSecret.name` is not `hcloud`, create an additional workload-cluster secret named `hcloud` for upstream compatibility
- update API comments and generated CRD docs to describe the upstream-compatible secret naming/key behavior
- keep recommending upstream defaults such as secret name `hcloud`, token key `token`, and Robot keys `robot-user` / `robot-password`
- add a `note` entry to the workload-cluster secret to make it obvious that CAPH reconciles it

# Comparing Secrets and Names of Keys

HCloud CCM Defaults ([Helm Chart](https://github.com/guettli/hcloud-cloud-controller-manager/blob/7452f13f7a196bc078b02f8aecb1a7b823cc36f7/chart/values.yaml#L54)):
```yaml
# Name of Secret: hcloud
token: ...
robot-user: ...
robot-password: ...
```

Hetzner Syself ccm ([Helm Chart](https://github.com/syself/charts/blob/01d909c820c458e4c548ff68f8fabdb06220cc3b/charts/ccm-hcloud/values.yaml#L29-L33)):
```yaml
# Name of Secret: hetzner
hcloud: ...
robot-user: ...
robot-password: ...
```

# This is version II

We started the PR again, because it was too big.

First: #1703


api/v1beta1/types.go
```diff
@@ -101,9 +101,19 @@ type HCloudPlacementGroupStatus struct {
 
 // HetznerSecretRef defines all the names of the secret and the relevant keys needed to access Hetzner API.
 type HetznerSecretRef struct {
-	// Name defines the name of the secret.
+	// Name defines the name of the secret. The name gets used for reading the credential in the
+	// mgt-cluster, and it gets used for creating a secret in the wl-cluster. About the secret in
+	// the wl-cluster: Attention, the upstream hcloud-ccm helm chart expects the name to be
+	// "hcloud". The Syself ccm fork defaults to "hetzner". To stay compatible with upstream
```
janiskemper: "To facilitate migrations" not "To stay compatible"

api/v1beta1/types.go
```diff
@@ -113,18 +123,38 @@ type HetznerSecretRef struct {
 // Need to specify either HCloudToken or both HetznerRobotUser and HetznerRobotPassword.
 type HetznerSecretKeyRef struct {
 	// HCloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
+	// It gets used for reading the credential in the mgt-cluster, and it gets used for creating a
+	// secret in the wl-cluster. We recommend to use "token", because this is the default of
+	// upstream hcloud-ccm, while the legacy Syself ccm fork uses "hcloud". For maximal
```
janiskemper: "To facilitate migrations" not for maximal compatibility.

api/v1beta1/types.go
```diff
@@ -113,18 +123,38 @@ type HetznerSecretRef struct {
...
+	// compatibility up to three keys get created in the secret for HCLOUD_TOKEN: "hcloud", "token"
```
janiskemper: we shouldn't do this. We should only create two maximum

api/v1beta1/types.go
```diff
@@ -113,18 +123,38 @@ type HetznerSecretRef struct {
...
+	// and the value of hcloudToken. This way we ensure that the ccm in the wl-cluster finds the
+	// secret.
+	//
 	// +optional
 	// +kubebuilder:default=hcloud-token
 	HCloudToken string `json:"hcloudToken"`
-	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is stored.
+
+	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is
```
janiskemper, line 141: why don't we create two keys here for compatibility?

api/v1beta1/types.go
guettli: @janiskemper why one should not be created? Please explain me why this should not be created.

api/v1beta1/types.go
guettli: done.

api/v1beta1/types.go
guettli: done.

api/v1beta1/types.go
guettli: @janiskemper I added a comparison to the PR description. Both CCMS use the same keys for robot-user and robot-password.

api/v1beta1/types.go
guettli: @janiskemper ok, I guess you mean when we create the secret called "hcloud", then create key "token", then we create secret "hetzner", then we create "hcloud". Yes, this makes sense. I updated the PR.

api/v1beta1/types.go
janiskemper: We only create one secret additionally, if the specified one is NOT the one the hcloud ccm wants. 

So we also only need to add another key to the secret that the hcloud ccm wants, right? We don't have to add another key to the secret that is irrelevant to hcloud ccm..

We only do this in the first place to facilitate migrations to the upstream hcloud ccm. Therefore, we can keep it lean and only focus on what that ccm wants, and that's all we need to do

guettli: @janiskemper ok, I adapted the PR accordingly.

janiskemper: Cool! I think I now have a clear picture here as well. I think we should revamp the whole comments here that we added in the PR and align it around the following principle:

> If the secret name is not according to what hcloud ccm expects, we create another secret with all keys as the hcloud ccm wants. The "main" secret, based on all this configuration, remains untouched. Only in the rare case where the main secret was called hcloud, but different keys where used than the ones the hcloud ccm expects, then one value will be added multiple times within that secret

guettli: @janiskemper I tried to implement your last comment in 6cea60a3. Please check it.

api/v1beta1/types.go
```diff
@@ -113,18 +123,38 @@ type HetznerSecretRef struct {
...
 	// HCloudToken defines the name of the key where the token for the Hetzner Cloud API is stored.
-	// +optional
-	// +kubebuilder:default=hcloud-token
+	// The controller reads the token from the management-cluster secret using this key and writes
+	// it to workload-cluster secrets using the same key. Upstream hcloud-ccm expects the key
+	// "token", while the legacy Syself ccm fork uses "hcloud". The configured workload-cluster
+	// secret remains untouched. CAPH only adds the compatibility key "token" inside the
+	// workload-cluster secret named "hcloud", and only if this field is not already "token". If
+	// the configured secret name is already "hcloud", the same token value is stored under both
+	// keys in that single secret. We recommend to use "token".
```
janiskemper, line 136: > The configured workload-cluster
	// secret remains untouched. CAPH only adds the compatibility key "token" inside the
	// workload-cluster secret named "hcloud", and only if this field is not already "token". If
	// the configured secret name is already "hcloud", the same token value is stored under both
	// keys in that single secret.

I would write instead of this:

To facilitate migrations to the upstream hcloud-ccm, it is ensured that the secret it expects is created in the proper way. This means that CAPH ensures that in the secret called "hcloud" there is the hcloud token stored under the key "token". This might lead to the same value being stored in two different keys of the same secret, depending on the configuration used here.

api/v1beta1/types.go
janiskemper: > @janiskemper I added a comparison to the PR description. Both CCMS use the same keys for robot-user and robot-password.

@guettli so that means that both CCMs use the same key, but here in our default we have a different one than that?

In the code I don't see that we hardcode the value that you mention "robot-user". Therefore, this secret would actually be created wrongly if I specify something else with this configuration than "robot-user".

janiskemper (changes requested):

api/v1beta1/types.go
guettli: @janiskemper I can do that, but I dont think this is needed. In all our templates, and in our commercial offering we use `robot-user` and `robot-password. Example: [hcloud-hetznerCluster.yaml](https://github.com/syself/cluster-api-provider-hetzner/blob/main/templates/cluster-templates/bases/hcloud-hetznerCluster.yaml#L18-L23)

janiskemper: okay, but we should still fix the code, because we cannot rely on people just using the templates, right? Currently, our code is wrong and we need to fix it by using the right configuration

api/v1beta1/types.go
janiskemper: @guettli this one you overlooked I believe

guettli: @janiskemper done: 2aeee8f7

api/v1beta1/types.go
janiskemper: @guettli then this is left before I review again (but I think we are almost there)

guettli: @janiskemper  done in https://github.com/syself/cluster-api-provider-hetzner/commit/50d838e124bf1a747c1a3581fefd04b32c109024 . The `hcloud` compatibility secret now also writes `robot-user` / `robot-password` aliases when the configured robot keys differ.

controllers/hetznercluster_controller.go
```diff
@@ -531,15 +533,49 @@ func reconcileWorkloadClusterSecrets(ctx context.Context, clusterScope *scope.Cl
...
+
+func workloadClusterSecretNames(secretName string) []string {
+	names := []string{secretName}
+	if secretName != "hcloud" {
+		names = append(names, "hcloud")
+	}
+	return names
+}
+
+func workloadClusterCompatibilityKeys(secretName, configuredKey, compatibilityKey string) []string {
```
janiskemper, line 553: plz explain this function in a comment. It's not trivial at all.

controllers/hetznercluster_controller.go
```diff
@@ -580,13 +622,29 @@ func reconcileWorkloadClusterSecrets(ctx context.Context, clusterScope *scope.Cl
 		wlSecret.Data["apiserver-host"] = []byte(clusterScope.HetznerCluster.Spec.ControlPlaneEndpoint.Host)
 		wlSecret.Data["apiserver-port"] = []byte(strconv.Itoa(int(clusterScope.HetznerCluster.Spec.ControlPlaneEndpoint.Port)))
 
+		notes := []string{
+			"This secret gets reconciled by Cluster API Provider Hetzner.",
+		}
+
+		if clusterScope.HetznerCluster.Spec.HetznerSecret.Name != "hcloud" {
```
janiskemper: I think this might be misleading to people. There is no action needed. We should rather invest in documentation rather than commenting here. People cannot easily change it here for example, but they need to go back to the mgt cluster. 

I'm fine with keeping the information about who owns the secret, but would remove these comments.

controllers/hetznercluster_controller_test.go
```diff
@@ -254,6 +259,296 @@ func TestIgnoreInsignificantHetznerClusterStatusUpdates(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := workloadClusterCompatibilityKeys(tc.secretName, tc.configuredKey, "token")
+			require.Equal(t, tc.want, got)
+		})
+	}
+}
+
+func TestReconcileOneWorkloadClusterSecretHetzner(t *testing.T) {
```
janiskemper, line 342: can you add comments to all tests to make sure that it is clear what this test is about?

janiskemper (dismissed):

controllers/hetznercluster_controller.go
guettli: @janiskemper I simplified the function and added a comment.

controllers/hetznercluster_controller.go
guettli: @janiskemper I am unsure what you want it to look like. Please elaborate.

controllers/hetznercluster_controller_test.go
guettli: done.

controllers/hetznercluster_controller.go
janiskemper: I would remove the "notes"

guettli: @janiskemper I removed "notes" completely.

janiskemper (commented):
LGTM. @Dhairya-Arora01 plz also review and tell us also if you don't understand something and want more explanation

controllers/hetznercluster_controller.go
```diff
@@ -489,9 +489,11 @@ func hcloudTokenErrorResult(
 	return res, nil
 }
 
-// reconcileWorkloadClusterSecrets ensures that the workload-cluster has the secret needed by the
-// ccm. The name of the secret is read from HetznerCluster.Spec.HetznerSecret.Name. Creating the
-// secret gets skipped, if HetznerCluster.Spec.SkipCreatingHetznerSecretInWorkloadCluster is set.
+// reconcileWorkloadClusterSecrets syncs Hetzner credentials from the management cluster into
+// workload-cluster secrets for CCM consumption. It creates the secret named in
```
Dhairya-Arora01: I would say: 
It creates a secret whose name matches `HetznerCluster.Spec.HetznerSecret.Name`. If that name is not "hcloud", a second secret named "hcloud" is also created

Dhairya-Arora01 (approved):
LGTM

controllers/hetznercluster_controller.go
guettli: @Dhairya-Arora01 please see: a158bef2 ok?

Dhairya-Arora01: works...

PR #1927, opened 2026-04-08
:seedling: Fix e2e tests on current main (v1.0.x)

guettli:
Fix e2e tests on current main (v1.0.x)

- Update from outdated hcloud machine types (like cpx31 to cpx32).


guettli:
v1.1.x is on main now, this PR is no longer needed.

PR #1930, opened 2026-04-08
:bug: Remove the check for `HCloudCredentialsInvalid` to unblock reconcilement

Dhairya-Arora01:
Currently, we skip reconciling HCloudMachine completely if the condition `HCloudTokenAvailable` is set to false with reason `HCloudCredentialsInvalid`. We added this check in this PR https://github.com/syself/cluster-api-provider-hetzner/pull/1888

This means that even when the CAPH controller restarts, the HCloudMachine is not reconciled. We don't want this behaviour, as on CAPH restart we want to reconcile all the related objects. For example, if Hetzner's API was temporarily unstable and returned an erroneous "unauthorized" response, the condition would be set incorrectly and without a restart triggering reconciliation, the machine would remain stuck indefinitely. Therefore, the early check that skips reconciliation based on the condition is removed.

Note that this does not cause repeated reconciliation attempts with invalid credentials as we still stop reconciling immediately after setting the `HCloudCredentialsInvalid` condition during the normal reconcile loop. The only difference is that we no longer prevent reconciliation from starting based on a previously set condition.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hcloudmachine_controller.go
+++ b/controllers/hcloudmachine_controller.go
@@ -100,15 +100,6 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 
 	log = log.WithValues("Machine", klog.KObj(machine))
 
-	// If the HCloudMachine has an unauthorized / HCloudCredentialsInvalid error, stop reconciling.
-	// There is no point in retrying API calls with an invalid token.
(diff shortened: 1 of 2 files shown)


janiskemper (approved):
let's add a unit test for the delete (that it doesn't try finding any server)

PR #1935, opened 2026-04-09
🌱 Introduce v1beta2 conditions for HCloudMachine

### LINKED ISSUE #1934: Introduce v1beta2 conditions for HCloudMachine
## Summary

Add v1beta2 conditions to `HCloudMachine` on `v1.1.x` while keeping the existing v1beta1 conditions unchanged.

- Base branch: `v1.1.x`

## What would you like to add

`HCloudMachine` is a primary infrastructure object that manages a real Hetzner Cloud server.

Today the machine lifecycle is surfaced through v1beta1 conditions such as:

| Condition | Purpose |
|---|---|
| `HCloudTokenAvailable` | Hetzner API credentials are valid and reachable |
| `BootstrapReady` | Bootstrap data has been provided by CAPI |
| `ServerCreateSucceeded` | Server was successfully created in HCloud |
| `ServerAvailable` | Server is running and usable |
| `HetznerAPIReachable` | The controller is not blocked by HCloud rate limiting |

These conditions are still needed and must keep their current behavior. At the same time, the CAPI v1beta2 transition requires `HCloudMachine` to expose v1beta2 conditions in parallel.

Deletion also needs to be visible more clearly. Today it is mostly inferred from `metadata.deletionTimestamp`, finalizers, and legacy `ServerAvailable=False` transitions. We want a dedicated v1beta2 `Deleting` condition for `HCloudMachine`.

## The Solution

Follow the same structure CAPV used:

- Add `status.v1beta2.conditions` to `HCloudMachine`.
- Keep writing the existing v1beta1 conditions exactly as today.
- Add direct `v1beta2conditions.Set(...)` calls in the same controller and service code paths that already set the v1beta1 conditions.
- Add a controller local patch step for `HCloudMachine` that computes:
  - the legacy v1beta1 summary
  - the v1beta2 `Ready` summary
  - the owned v1beta2 condition set for patching
- Add a dedicated v1beta2 `Deleting` condition.

This keeps the behavior explicit at each reconcile decision point and avoids a second translation layer that can drift from the real lifecycle logic.

## Final Conditions

The final v1beta2 conditions added for `HCloudMachine` are:

| Constant | Type | Polarity | Notes |
|---|---|---|---|
| `HCloudMachineReadyV1Beta2Condition` | `clusterv1.ReadyV1Beta2Condition` | Positive | Summary condition |
| `HCloudMachineHCloudTokenAvailableV1Beta2Condition` | `HCloudTokenAvailable` | Positive | True when HCloud credentials are valid and reachable |
| `HCloudMachineBootstrapReadyV1Beta2Condition` | `BootstrapReady` | Positive | True when bootstrap data is available |
| `HCloudMachineServerCreatedV1Beta2Condition` | `ServerCreated` | Positive | True when the HCloud server has been provisioned |
| `HCloudMachineServerAvailableV1Beta2Condition` | `ServerAvailable` | Positive | True when the server is running and usable |
| `HCloudMachineDeletingV1Beta2Condition` | `clusterv1.DeletingV1Beta2Condition` | Negative | Dedicated deletion signal for `HCloudMachine` |
| `HCloudMachineHCloudRateLimitExceededV1Beta2Condition` | `HCloudRateLimitExceeded` | Negative | True when HCloud API rate limit has been exceeded. Set to False on every successful reconcile (happy path setter). |

abdullah599:
Introduces v1beta2 conditions for HCloudMachine, following the upstream CAPI v1beta2 condition pattern. This adds a new `V1Beta2Status` field with `metav1.Condition` entries alongside the existing v1beta1 conditions.

Fixes #1934

- The v1beta2 conditions mirror the v1beta1 conditions at every call site
- `HandleRateLimitExceeded` and `reconcileRateLimit` use type assertions to conditionally set v1beta2 conditions only for `*HCloudMachine`
- `hcloudTokenErrorResult` similarly uses a type assertion since it handles both `HetznerCluster` and `HCloudMachine` setters
- `IgnoreTypesIfMissing` includes BootstrapReady, ServerCreated, ServerProvisioned, ServerAvailable, and HetznerAPIReachable to handle early-exit paths where those conditions may not yet be set
- The condition order in `HCloudMachineV1Beta2OwnedConditions` defines the Ready summary priority: HCloudTokenAvailable > HetznerAPIReachable > Deleting > BootstrapReady > ServerCreated > ServerProvisioned > ServerAvailable

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta1/hcloudmachine_types.go
```diff
@@ -165,6 +169,16 @@ type HCloudMachineStatus struct {
 	ExternalIDs HCloudMachineStatusExternalIDs `json:"externalIDs,omitempty"`
 }
 
+// HCloudMachineV1Beta2Status groups all the fields that will be added or modified in HCloudMachineStatus with the V1Beta2 version.
+type HCloudMachineV1Beta2Status struct {
+	// conditions represents the observations of a HCloudMachine's current state.
+	// Known condition types are Ready, HCloudTokenAvailable, BootstrapReady, ServerCreated, ServerAvailable, Deleting and HetznerAPIReachable.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
```
abdullah599, line 190: vshpere also added `// +kubebuilder:validation:MaxItems=32` should we do that?

api/v1beta1/conditions_const.go
```diff
@@ -251,3 +251,131 @@ const (
...
+	HCloudMachineNotReadyV1Beta2Reason = clusterv1.NotReadyV1Beta2Reason
+	// HCloudMachineReadyUnknownV1Beta2Reason surfaces when the HCloudMachine readiness is unknown.
+	HCloudMachineReadyUnknownV1Beta2Reason = clusterv1.ReadyUnknownV1Beta2Reason
+)
+
+const (
+	// HCloudMachineHCloudTokenAvailableV1Beta2Condition reports on whether the HCloud token is available.
+	HCloudMachineHCloudTokenAvailableV1Beta2Condition = "HCloudTokenAvailable"
+	// HCloudMachineTokenAvailableV1Beta2Reason surfaces when the HCloud token is available.
+	HCloudMachineTokenAvailableV1Beta2Reason = "TokenAvailable"
```
janiskemper: in general: I would name these conditions always "available" etc without duplicating the "token" or something else

api/v1beta1/conditions_const.go
```diff
@@ -251,3 +251,131 @@ const (
...
+	// HCloudMachineTokenSecretUnreachableV1Beta2Reason surfaces when the Hetzner secret cannot be reached.
+	HCloudMachineTokenSecretUnreachableV1Beta2Reason = "SecretUnreachable"
+	// HCloudMachineTokenInvalidV1Beta2Reason surfaces when the HCloud token is invalid.
+	HCloudMachineTokenInvalidV1Beta2Reason = "CredentialsInvalid"
+)
+
+const (
+	// HCloudMachineBootstrapReadyV1Beta2Condition reports on whether bootstrap data is ready.
+	HCloudMachineBootstrapReadyV1Beta2Condition = "BootstrapReady"
```
janiskemper: let's get rid of this and integrate it to "provisioned"

api/v1beta1/conditions_const.go
```diff
@@ -251,3 +251,131 @@ const (
...
+	HCloudMachineBootstrapReadyV1Beta2Reason = "BootstrapReady"
+	// HCloudMachineBootstrapNotReadyV1Beta2Reason surfaces when bootstrap data is not ready yet.
+	HCloudMachineBootstrapNotReadyV1Beta2Reason = clusterv1.WaitingForBootstrapDataV1Beta2Reason
+)
+
+const (
+	// HCloudMachineServerCreatedV1Beta2Condition reports on whether the HCloud server was created.
+	HCloudMachineServerCreatedV1Beta2Condition = "ServerCreated"
+	// HCloudMachineServerCreatedV1Beta2Reason surfaces when the HCloud server is provisioned.
+	HCloudMachineServerCreatedV1Beta2Reason = clusterv1.ProvisionedV1Beta2Reason
```
janiskemper: I think "provisioned" is not correct here

controllers/hetznercluster_controller.go
```diff
@@ -407,6 +408,14 @@ func reconcileRateLimit(setter conditions.Setter, rateLimitWaitTime time.Duratio
 		}
 		// Wait time is over, we continue
 		conditions.MarkTrue(setter, infrav1.HetznerAPIReachableCondition)
+		if hcloudMachine, ok := setter.(*infrav1.HCloudMachine); ok {
+			v1beta2conditions.Set(hcloudMachine, metav1.Condition{
+				Type:    infrav1.HCloudMachineHCloudRateLimitExceededV1Beta2Condition,
+				Status:  metav1.ConditionFalse,
+				Reason:  infrav1.HCloudMachineRateLimitNotExceededV1Beta2Reason,
+				Message: "rate limit wait time is over",
```
janiskemper: we can do this, but only if in the next reconcile this would be set back to some standard value.. If not, we should do something neutral here, maybe even without message. We have to remember that this will stay in the conditions

controllers/hetznercluster_controller.go
```diff
@@ -480,6 +492,35 @@ func hcloudTokenErrorResult(
...
 	}
 	conditions.SetSummary(setter)
+	if hcloudMachine, ok := setter.(*infrav1.HCloudMachine); ok {
+		v1beta2conditions.Set(hcloudMachine, metav1.Condition{
+			Type:    infrav1.HCloudMachineHCloudTokenAvailableV1Beta2Condition,
+			Status:  metav1.ConditionFalse,
+			Reason:  v1beta2Reason,
+			Message: conditions.GetMessage(hcloudMachine, infrav1.HCloudTokenAvailableCondition),
+		})
+		if !hcloudMachine.DeletionTimestamp.IsZero() {
```
janiskemper: don't we have another place where we already set this? why here?

api/v1beta1/conditions_const.go
```diff
@@ -251,3 +251,131 @@ const (
...
+	HCloudMachineServerNotFoundV1Beta2Reason = "ServerNotFound"
+	// HCloudMachineNetworkAttachFailedV1Beta2Reason surfaces a network attachment failure.
+	HCloudMachineNetworkAttachFailedV1Beta2Reason = "NetworkAttachFailed"
+	// HCloudMachineLoadBalancerAttachFailedV1Beta2Reason surfaces a load balancer attachment failure.
+	HCloudMachineLoadBalancerAttachFailedV1Beta2Reason = "LoadBalancerAttachFailed"
+)
+
+const (
+	// HCloudMachineDeletingV1Beta2Condition reports on HCloudMachine deletion progress.
+	HCloudMachineDeletingV1Beta2Condition = clusterv1.DeletingV1Beta2Condition
```
janiskemper: In case we don't ever use this for something useful, and set the Available condition anyway to "delete", we can consider this again whether we want it or not. 

We should mainly achieve that during a deletion process and there are issues, we show these issues. And also we should show that we are in a deletion process, which we would also show via InstanceState.

janiskemper (changes requested):

api/v1beta1/conditions_const.go
abdullah599: We’re using Deleting to surface specific error messages during the delete flow (e.g., when server deletion fails). ServerAvailable with reason Deleting shows the state, but Deleting carries the error details that bubble up to the Ready summary.

janiskemper: agreed, but which cases do we have? 
- not found -> ignore
- unauthorized -> we have a separate condition already
- rate limit exceeded -> we have a separate condition

I'm not able to come up with any other right now

abdullah599: hmm makes sense, I think we can remove it then and show deletion state through ServerAvailable (False, reason: Deleting). CAPV does the same i.e. no standalone Deleting condition, just a DeletingReason on existing conditions.

janiskemper: I'm totally open to adding it if necessary, but based on this PR and my analysis above I don't see a reason right now.

abdullah599: we can introduce it later if needed :)

api/v1beta1/conditions_const.go
abdullah599: For HCloudMachine, bootstrap data (user-data) is passed to the HCloud API during server creation itself, so without it the server can't even be created. The lifecycle is: check bootstrap → create server → provision (rescue boot, image install, OS startup) → available.

Since missing bootstrap data blocks server creation, not provisioning, I think ServerCreated=False, Reason=WaitingForBootstrapData is the right fit. It answers "why isn't the server created?" directly. WDYT?

janiskemper: sounds good

api/v1beta1/conditions_const.go
```diff
@@ -255,3 +255,113 @@ const (
...
+	HCloudMachineServerProvisioningV1Beta2Reason = "Provisioning"
+)
+
+const (
+	// HCloudMachineServerAvailableV1Beta2Condition reports on whether the HCloud server is available.
+	HCloudMachineServerAvailableV1Beta2Condition = "ServerAvailable"
+	// HCloudMachineServerAvailableV1Beta2Reason surfaces when the HCloud server is available.
+	HCloudMachineServerAvailableV1Beta2Reason = clusterv1.AvailableV1Beta2Reason
+	// HCloudMachineServerNotAvailableV1Beta2Reason surfaces when the HCloud server is not available.
+	HCloudMachineServerNotAvailableV1Beta2Reason = clusterv1.NotAvailableV1Beta2Reason
```
janiskemper: let's check that we keep only the ones that are used. This for example is not

controllers/hcloudmachine_controller_test.go
```diff
@@ -514,7 +529,8 @@ var _ = Describe("HCloudMachineReconciler", func() {
 
 			It("checks that ImageNotFound is visible in conditions if image does not exist", func() {
 				Eventually(func() bool {
-					return isPresentAndFalseWithReason(key, hcloudMachine, infrav1.ServerCreateSucceededCondition, infrav1.ImageNotFoundReason)
+					return isPresentAndFalseWithReason(key, hcloudMachine, infrav1.ServerCreateSucceededCondition, infrav1.ImageNotFoundReason) &&
+						isPresentAndFalseWithReasonV1Beta2(key, hcloudMachine, infrav1.HCloudMachineServerCreatedV1Beta2Condition, infrav1.HCloudMachineServerNotCreatedV1Beta2Reason)
```
janiskemper: this reason doesn't seem great. We should have the specific one. Let's check all cases again and whether we at least have as good conditions / reasons as before

controllers/hetznercluster_controller.go
```diff
@@ -407,6 +408,13 @@ func reconcileRateLimit(setter conditions.Setter, rateLimitWaitTime time.Duratio
 		}
 		// Wait time is over, we continue
 		conditions.MarkTrue(setter, infrav1.HetznerAPIReachableCondition)
+		if hcloudMachine, ok := setter.(*infrav1.HCloudMachine); ok {
```
janiskemper: this doesn't seem good. It applies also to multiple other objects. I think we should either say that we use a common condition (IMO it would be possible), or take this out of the generic function.

Same below with the hcloud token

janiskemper (changes requested):

controllers/hetznercluster_controller.go
abdullah599: i think we should extract the v1beta2 logic out of the generic functions

for reconcileRateLimit, the controller already sets HCloudRateLimitExceeded=False right after the call, so the type assertion inside is not needed. For hcloudTokenErrorResult, we can set the v1beta2 token condition in the HCloudMachine controller before calling it since we already know the error type there.

WDYT?

abdullah599: update: we decided to use common conditions

api/v1beta1/conditions_const.go
```diff
@@ -65,6 +65,10 @@ const (
 	ServerAvailableCondition clusterv1.ConditionType = "ServerAvailable"
 	// ServerTerminatingReason instance is in a terminated state.
 	ServerTerminatingReason = "InstanceTerminated"
+	// HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason indicates the image URL command is not accessible.
+	HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason = "ImageURLCommandNotAccessible"
```
Dhairya-Arora01: Maybe you can move these two conditions below as well, in the const block for `ServerAvailable` v1beta2 conditions.

api/v1beta1/conditions_const.go
```diff
@@ -255,3 +259,165 @@ const (
...
+)
+
+const (
+	// HCloudMachineServerProvisionedV1Beta2Condition reports on whether the HCloud server has completed
+	// boot-time provisioning (rescue boot, image install, OS startup).
+	HCloudMachineServerProvisionedV1Beta2Condition = "ServerProvisioned"
+	// HCloudMachineServerProvisionedV1Beta2Reason surfaces when the boot state machine has completed.
+	HCloudMachineServerProvisionedV1Beta2Reason = clusterv1.ProvisionedV1Beta2Reason
+	// HCloudMachineHandleBootStateUnsetV1Beta2Reason indicates the boot state is being initialized.
+	HCloudMachineHandleBootStateUnsetV1Beta2Reason = "HandleBootStateUnset"
```
Dhairya-Arora01: I'm not sure if `HandleBootStateUnset` is a good name for `reason`. If it indicates that the boot state is being initialized then how about renaming it as `BootStateInitializing` ?

pkg/scope/machine.go
```diff
@@ -121,7 +149,29 @@ func (m *MachineScope) Namespace() string {
...
+func SetHCloudMachineV1Beta2SummaryCondition(hcloudMachine *infrav1.HCloudMachine) error {
+	return v1beta2conditions.SetSummaryCondition(hcloudMachine, hcloudMachine, infrav1.HCloudMachineReadyV1Beta2Condition,
+		infrav1.HCloudMachineV1Beta2SummaryOpts()...,
+	)
+}
+
+// machinePatchOpts returns the list of patch.Option for HCloudMachine.
+func machinePatchOpts() []patch.Option {
+	return []patch.Option{
+		// owned v1beta2 conditions.
```
Dhairya-Arora01, line 175: shouldn't we include the owned v1beta1 conditions as well ?

abdullah599: v1beta1 conditions were not declared as owned anywhere in CAPH before this change, and I was trying to avoid modifying anything related to v1beta1. I can add it if you prefer :)

Dhairya-Arora01: Since, not multiple controllers set the conditions on HCloudMachine, it doesn't make any difference to specify owned v1beta1 conditions. But then we can apply to same logic for v1beta2 conditions as well :grinning: . So I think we are specifying owned v1beta2 conditions then we can specify owned v1beta1 conditions as well. Although it won't make a difference in our case.

janiskemper: I think though that we don't have to do it because we will deprecate it soon. Therefore no reason to spend time with it if it is not really relevant

Dhairya-Arora01: Fine with that too....

abdullah599: > Therefore no reason to spend time with it if it is not really relevant

Time was already spent :)


I think we can keep it since I have already made the change 🤷‍♂️

Dhairya-Arora01 (approved):
LGTM

api/v1beta1/conditions_const.go
```diff
@@ -255,3 +255,169 @@ const (
...
+
+const (
+	// HCloudTokenAvailableV1Beta2Condition reports on whether the HCloud Token is available.
+	HCloudTokenAvailableV1Beta2Condition = "HCloudTokenAvailable"
+	// HCloudTokenAvailableV1Beta2Reason indicates that the HCloudToken is available.
+	HCloudTokenAvailableV1Beta2Reason = clusterv1.AvailableV1Beta2Reason
+	// HCloudTokenInvalidV1Beta2Reason indicates that the HCloudToken is invalid.
+	HCloudTokenInvalidV1Beta2Reason = "Invalid"
+	// HetznerSecretUnreachableV1Beta2Reason indicates that Hetzner secret is unreachable.
+	HetznerSecretUnreachableV1Beta2Reason = "HetznerSecretUnreachable" // #nosec
```
janiskemper: let's actually remove the "Hetzner" from here. That is a naming pattern that you can in theory even configure with our objects. It doesn't have to be called "Hetzner"

pkg/scope/machine.go
```diff
@@ -98,10 +100,36 @@ type MachineScope struct {
...
 
-// Close closes the current scope persisting the cluster configuration and status.
+// Close closes the current scope persisting the machine configuration and status.
 func (m *MachineScope) Close(ctx context.Context) error {
+	// set summary for v1beta1 conditions.
 	conditions.SetSummary(m.HCloudMachine)
-	return m.patchHelper.Patch(ctx, m.HCloudMachine)
+
+	// set summary for v1beta2 conditions.
+	readyCondition, err := v1beta2conditions.NewSummaryCondition(
```
janiskemper, line 109: we use "NewSummaryCondition" here, but also have a function using "SetSummaryCondition" below. In the two summaries we use different "Ready" conditions - once with the HCloudMachine specific condition, one with the generic one.. 

There seems to be something wrong.

api/v1beta1/conditions_const.go
```diff
@@ -255,3 +255,169 @@ const (
...
+
+const (
+	// HCloudMachineServerCreatedV1Beta2Condition reports on whether the HCloud server was created.
+	HCloudMachineServerCreatedV1Beta2Condition = "ServerCreated"
+	// HCloudMachineServerCreatedV1Beta2Reason surfaces when the HCloud server has been created.
+	HCloudMachineServerCreatedV1Beta2Reason = "Created"
+	// HCloudMachineServerWaitingForBootstrapDataV1Beta2Reason surfaces when the server cannot be created because bootstrap data is not yet available.
+	HCloudMachineServerWaitingForBootstrapDataV1Beta2Reason = clusterv1.WaitingForBootstrapDataV1Beta2Reason
+	// HCloudMachineServerCreateFailedIrrecoverableV1Beta2Reason surfaces an irrecoverable create failure.
+	HCloudMachineServerCreateFailedIrrecoverableV1Beta2Reason = "CreateFailedIrrecoverable"
```
janiskemper: we should fix the grammar. It should be "irrecoverably" right?

api/v1beta1/conditions_const.go
```diff
@@ -255,3 +255,169 @@ const (
...
+	HCloudMachineUnknownHCloudStatusV1Beta2Reason = "UnknownHCloudStatus"
+	// HCloudMachineUnknownServerStatusV1Beta2Reason indicates an unknown server status.
+	HCloudMachineUnknownServerStatusV1Beta2Reason = "UnknownServerStatus"
+	// HCloudMachineImageURLSetButNoCommandProvidedV1Beta2Reason indicates imageURL is set but no command is provided.
+	HCloudMachineImageURLSetButNoCommandProvidedV1Beta2Reason = "ImageURLSetButNoCommandProvided"
+
+	// HCloudMachineWaitForRescueSystemV1Beta2Reason indicates waiting for the rescue system to be enabled.
+	HCloudMachineWaitForRescueSystemV1Beta2Reason = "WaitForRescueSystem"
+	// HCloudMachineEnablingRescueTimedOutV1Beta2Reason indicates enabling rescue system timed out.
+	HCloudMachineEnablingRescueTimedOutV1Beta2Reason = "EnablingRescueTimedOut"
```
janiskemper, line 334: this seems to not be used.. I checked it randomly because I wanted to understand how we use it.. Do you wanna check the other reasons as well, if they are not used?

api/v1beta1/hcloudmachine_types.go
```diff
@@ -213,6 +229,68 @@ func (r *HCloudMachine) SetConditions(conditions clusterv1.Conditions) {
...
+			HCloudRateLimitExceededV1Beta2Condition,
+		},
+		v1beta2conditions.IgnoreTypesIfMissing{
+			HCloudTokenAvailableV1Beta2Condition,
+			HCloudMachineServerCreatedV1Beta2Condition,
+			HCloudMachineServerProvisionedV1Beta2Condition,
+			HCloudMachineServerAvailableV1Beta2Condition,
+			HCloudRateLimitExceededV1Beta2Condition,
+		},
+		v1beta2conditions.CustomMergeStrategy{
```
janiskemper, line 282: what does this part mean exactly? can you maybe add comments?

controllers/hcloudmachine_controller.go
```diff
@@ -141,9 +145,22 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 
 	// Create the scope.
 	secretManager := secretutil.NewSecretManager(log, r, r.APIReader)
+
+	// Set ServerAvailable=False and InstanceState=Deleting before the token check
```
janiskemper: we don't have a patch helper created at this point. This change wouldn't have any effect therefore AFAIK

janiskemper (changes requested):

api/v1beta1/conditions_const.go
abdullah599: it is being used (i also verified other reasons): https://github.com/syself/cluster-api-provider-hetzner/blob/da4a77b289940c7d7eece5dd5d82178929543b77/pkg/services/hcloud/server/server.go#L520-L528

pkg/scope/machine.go
abdullah599: i'll drop the alias and switch all call sites to use clusterv1.ReadyV1Beta2Condition directly.

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+
+// v1beta2 conditions.
+
+// common conditions used across resource types.
+
+const (
+	// HCloudRateLimitExceededV1Beta2Condition reports on whether the HCloud API rate limit has been exceeded.
+	HCloudRateLimitExceededV1Beta2Condition = "HCloudRateLimitExceeded"
+	// HCloudRateLimitExceededV1Beta2Reason indicates that the HCloud API rate limit has been exceeded.
+	HCloudRateLimitExceededV1Beta2Reason = "RateLimitExceeded"
```
janiskemper:
```suggestion
	HCloudRateLimitExceededV1Beta2Reason = "Exceeded"
```

janiskemper: plz check all occurrences for this pattern, we have also done this in the other PRs. You can even go through the other PRs as well to ensure they are compliant.

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	// HCloudMachineServerProvisionedV1Beta2Reason surfaces when the boot state machine has completed.
+	HCloudMachineServerProvisionedV1Beta2Reason = clusterv1beta1.ProvisionedV1Beta2Reason
+	// HCloudMachineBootStateInitializingV1Beta2Reason indicates the boot state is being initialized.
+	HCloudMachineBootStateInitializingV1Beta2Reason = "BootStateInitializing"
+	// HCloudMachineBootStateInitializingTimedOutV1Beta2Reason indicates the boot state initialization timed out.
+	HCloudMachineBootStateInitializingTimedOutV1Beta2Reason = "BootStateInitializingTimedOut"
+	// HCloudMachineProvisioningServerV1Beta2Reason indicates the server is being provisioned.
+	HCloudMachineProvisioningServerV1Beta2Reason = "Provisioning"
+	// HCloudMachineServerNotRunningYetV1Beta2Reason indicates the server is not running yet.
+	HCloudMachineServerNotRunningYetV1Beta2Reason = "NotRunningYet"
```
janiskemper: this should actually be "ServerNotRunningYet" because it refers to the actual HCloud server

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	// HCloudMachineUnknownHCloudStatusV1Beta2Reason indicates an unknown HCloud server status.
+	HCloudMachineUnknownHCloudStatusV1Beta2Reason = "UnknownHCloudStatus"
+	// HCloudMachineUnknownServerStatusV1Beta2Reason indicates an unknown server status.
```
janiskemper: what's the difference between this and the one before? Seems exactly the same

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	HCloudMachineUnknownServerStatusV1Beta2Reason = "UnknownServerStatus"
+	// HCloudMachineImageURLSetButNoCommandProvidedV1Beta2Reason indicates imageURL is set but no command is provided.
+	HCloudMachineImageURLSetButNoCommandProvidedV1Beta2Reason = "ImageURLSetButNoCommandProvided"
```
janiskemper: is this possible? I know we changed it from flag to property in API.. This should be validated now in the API layer

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+
+	// HCloudMachineWaitForRescueSystemV1Beta2Reason indicates waiting for the rescue system to be enabled.
+	HCloudMachineWaitForRescueSystemV1Beta2Reason = "WaitForRescueSystem"
```
janiskemper: this should be "waiting" not "wait" right? Considering how we did it in the other conditions?

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	HCloudMachineWaitingForEnablingRescueActionV1Beta2Reason = "WaitingForEnablingRescueAction"
+	// HCloudMachineEnablingRescueActionFailedV1Beta2Reason indicates the rescue enable action failed.
+	HCloudMachineEnablingRescueActionFailedV1Beta2Reason = "EnablingRescueActionFailed"
+	// HCloudMachineEnablingRescueActionDoneV1Beta2Reason indicates the rescue enable action is done.
+	HCloudMachineEnablingRescueActionDoneV1Beta2Reason = "EnablingRescueActionDone"
+	// HCloudMachineRescueNotEnabledYetV1Beta2Reason indicates the rescue system is not enabled yet.
+	HCloudMachineRescueNotEnabledYetV1Beta2Reason = "RescueNotEnabledYet"
+
+	// HCloudMachineGetSSHPrivateKeyFailedV1Beta2Reason indicates getting the SSH private key failed.
+	HCloudMachineGetSSHPrivateKeyFailedV1Beta2Reason = "GetSSHPrivateKeyFailed"
```
janiskemper: get -> getting? (same below)

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	// HCloudMachineGetSSHClientFailedV1Beta2Reason indicates getting the SSH client failed.
+	HCloudMachineGetSSHClientFailedV1Beta2Reason = "GetSSHClientFailed"
+	// HCloudMachineRetryingSSHConnectionV1Beta2Reason indicates the SSH connection is being retried.
+	HCloudMachineRetryingSSHConnectionV1Beta2Reason = "RetryingSSHConnection"
+	// HCloudMachineRebootViaSSHFailedV1Beta2Reason indicates rebooting via SSH failed.
+	HCloudMachineRebootViaSSHFailedV1Beta2Reason = "RebootViaSSHFailed"
+	// HCloudMachineGetHostnameFailedV1Beta2Reason indicates getting the hostname failed.
+	HCloudMachineGetHostnameFailedV1Beta2Reason = "GetHostnameFailed"
```
janiskemper: get -> getting?

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	// HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason indicates the image URL command is not accessible.
+	HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason = "ImageURLCommandNotAccessible"
+	// HCloudMachineStartImageURLCommandFailedV1Beta2Reason indicates starting the image URL command failed.
+	HCloudMachineStartImageURLCommandFailedV1Beta2Reason = "StartImageURLCommandFailed"
+	// HCloudMachineStartImageURLCommandNoZeroExitCodeV1Beta2Reason indicates the image URL command returned a non-zero exit code.
+	HCloudMachineStartImageURLCommandNoZeroExitCodeV1Beta2Reason = "StartImageURLCommandNoZeroExitCode"
+	// HCloudMachineHCloudImageURLCommandRunningV1Beta2Reason indicates the image URL command is running.
+	HCloudMachineHCloudImageURLCommandRunningV1Beta2Reason = "HCloudImageURLCommandRunning"
+	// HCloudMachineRunningImageCommandTimedOutV1Beta2Reason indicates the running image command timed out.
+	HCloudMachineRunningImageCommandTimedOutV1Beta2Reason = "RunningImageCommandTimedOut"
```
janiskemper: ImageURLCommand here and below

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+	// HCloudMachineGetServerImageFailedV1Beta2Reason indicates getting the server image failed.
+	HCloudMachineGetServerImageFailedV1Beta2Reason = "GetServerImageFailed"
+	// HCloudMachineGetRawBootstrapDataFailedV1Beta2Reason indicates getting the raw bootstrap data failed.
+	HCloudMachineGetRawBootstrapDataFailedV1Beta2Reason = "GetRawBootstrapDataFailed"
+	// HCloudMachinePowerOnServerFailedV1Beta2Reason indicates powering on the server failed.
+	HCloudMachinePowerOnServerFailedV1Beta2Reason = "PowerOnServerFailed"
+	// HCloudMachineServerOffV1Beta2Reason indicates the server is off.
+	HCloudMachineServerOffV1Beta2Reason = "ServerOff"
+	// HCloudMachineServerOffTimeoutV1Beta2Reason indicates the server off timeout was reached.
+	HCloudMachineServerOffTimeoutV1Beta2Reason = "ServerOffTimeout"
```
janiskemper: TimeoutReached

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,157 @@ const (
...
+
+const (
+	// HCloudMachineServerAvailableV1Beta2Condition reports on whether the HCloud server is available.
+	HCloudMachineServerAvailableV1Beta2Condition = "ServerAvailable"
+	// HCloudMachineServerAvailableV1Beta2Reason surfaces when the HCloud server is available.
+	HCloudMachineServerAvailableV1Beta2Reason = clusterv1beta1.AvailableV1Beta2Reason
+	// HCloudMachineServerNotFoundV1Beta2Reason surfaces when the HCloud server cannot be found.
+	HCloudMachineServerNotFoundV1Beta2Reason = "NotFound"
+	// HCloudMachineNetworkAttachFailedV1Beta2Reason surfaces a network attachment failure.
+	HCloudMachineNetworkAttachFailedV1Beta2Reason = "NetworkAttachFailed"
```
janiskemper: AttachingToNetworkFailed (similar below)

api/v1beta1/hcloudmachine_types.go
janiskemper: can you add the explanation everywhere?

janiskemper (changes requested):

api/v1beta1/conditions_const.go
abdullah599: Both v1beta1 reasons existed before this PR and were ported as-is into v1beta2.

both describe the same case. UnknownHCloudStatus is set in the early checks at server.go:492 when the hcloud server status is not Starting, Initializing, or Running. UnknownServerStatus is set later in a boot state branch at server.go:1081 when the status is unexpected at that stage. 

I merged both into one reason at both call sites. v1beta1 now uses "ServerStatusUnknown" and v1beta2 uses HCloudMachineServerStatusUnknownV1Beta2Reason = "ServerStatusUnknown". The v1beta1 severities (Info in the early check, Warning in the boot state branch) stay as they were.

api/v1beta1/conditions_const.go
```diff
@@ -295,3 +295,153 @@ const (
...
+	HCloudMachineBootingToRescueTimedOutV1Beta2Reason = "BootingToRescueTimedOut"
+	// HCloudMachineWaitForRescueEnabledToBeFalseV1Beta2Reason indicates waiting for rescue enabled to become false.
+	HCloudMachineWaitForRescueEnabledToBeFalseV1Beta2Reason = "WaitingForRescueEnabledToBeFalse"
+
+	// HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason indicates the image URL command is not accessible.
+	HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason = "ImageURLCommandNotAccessible"
+	// HCloudMachineStartImageURLCommandFailedV1Beta2Reason indicates starting the image URL command failed.
+	HCloudMachineStartImageURLCommandFailedV1Beta2Reason = "StartImageURLCommandFailed"
+	// HCloudMachineStartImageURLCommandNoZeroExitCodeV1Beta2Reason indicates the image URL command returned a non-zero exit code.
+	HCloudMachineStartImageURLCommandNoZeroExitCodeV1Beta2Reason = "StartImageURLCommandNoZeroExitCode"
```
Dhairya-Arora01: Shouldn't this be NonZeroExitCode instead of NoZerodExitCode ?

api/v1beta1/hcloudmachine_types.go
Dhairya-Arora01: I think we should add here that the only reason we use CustomMergeStrategy is to have reason aligned with CAPI upstream instead of CAPI's generic defaults. 

Currently, it looks like that without this even negative-polarity conditions won't be handled as well but that is not true.

abdullah599: > it looks like that without this even negative-polarity conditions won't be handled as well but that is not true.

That is the case. We had a discussion about this:
image


Am I missing something?

references:
capi: https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/cluster/cluster_controller_status.go#L1181C1-L1187C5

ref: https://pkg.go.dev/sigs.k8s.io/cluster-api/util/conditions#GetDefaultMergePriorityFunc

Dhairya-Arora01: see: https://github.com/syself/cluster-api-provider-hetzner/blob/56c6b415e115babe5fc38dd9c495f2856e250903/vendor/sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions/v1beta2/summary.go#L61-L64

if we don't specify the merge strategy then it already takes the DefaultMergeStrategy and specify the negative polarity conditions there - in the same way as we specify in our custom merge strategy.

That's why I said, the only reason that we are using CustomMergeStrategy is because we want to manipulate the reason in the summary condition. But since we have to specify CustomMergeStrategy, we need to specify our negative polarity conditions there with the default merge strategy.

I propose that we should add this as a note.

api/v1beta1/hcloudmachine_types.go
```diff
@@ -292,13 +285,18 @@ func HCloudMachineV1Beta2SummaryOpts() []v1beta2conditions.SummaryOption {
...
+		// here. When a CustomMergeStrategy is provided, NewSummaryCondition
+		// skips the path that wires up the NegativePolarityConditionTypes
+		// SummaryOption into the default strategy, so the negative-polarity
+		// types must be specified explicitly inside the strategy.
 		v1beta2conditions.CustomMergeStrategy{
 			MergeStrategy: v1beta2conditions.DefaultMergeStrategy(
-				// Register the rate-limit condition as negative-polarity (True = problem)
-				// so it's bucketed as an issue and drives the Ready summary when firing.
 				v1beta2conditions.GetPriorityFunc(v1beta2conditions.GetDefaultMergePriorityFunc(
 					HCloudRateLimitExceededV1Beta2Condition,
```
Dhairya-Arora01, line 302: can you add above this (HCloudRateLimitExceededV1Beta2Condition) - "conditions with negative polarity"

Dhairya-Arora01 (dismissed):

pkg/services/hcloud/server/server.go
janiskemper, line 550: is it possible that we don't have the error handling that we should have for the API call? With authentication and rate limit?

pkg/services/hcloud/server/server.go
```diff
@@ -273,6 +318,12 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 				"Server creation failed with an irrecoverable error: %s. If the requested resources (server type or location) become available again, delete the Machine to trigger a new creation attempt.",
 				err.Error(),
 			)
+			v1beta2conditions.Set(s.scope.HCloudMachine, metav1.Condition{
+				Type:    infrav1.HCloudMachineServerCreatedV1Beta2Condition,
+				Status:  metav1.ConditionFalse,
+				Reason:  infrav1.HCloudMachineServerCreateFailedIrrecoverablyV1Beta2Reason,
+				Message: err.Error(),
```
janiskemper: should we add the full message from above?

pkg/services/hcloud/server/server.go
```diff
@@ -333,25 +400,84 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
...
+			Status:  metav1.ConditionFalse,
+			Reason:  infrav1.HCloudMachineBootStateInitializingTimedOutV1Beta2Reason,
+			Message: msg,
+		})
 		return reconcile.Result{}, nil
 	}
 
 	updateHCloudMachineStatusFromServer(hm, server)
 
+	// This is a new machine with imageURL. The webhook validates that ImageURLCommand is set
```
janiskemper: is this supposed to be part of this PR?

janiskemper (dismissed):

janiskemper (approved):

PR #1944, opened 2026-04-10
:seedling: Remove redundant SSH-based reboot check from actionProvisioned

guettli:
**Summary:**

In `actionProvisioned`, after a reboot is triggered (Phase 1), the subsequent reconcile loop (Phase 2) falls back to `handleIncompleteBoot` when `SSHAfterInstallImage=true` and the BootID hasn't changed yet. `handleIncompleteBoot` uses SSH hostname comparison to detect whether a reboot started, and escalates through SSH -> software -> hardware reboots if the hostname doesn't match.
This logic was designed for `actionRegistering` and `actionEnsureProvisioned`, where the hostname changes across reboots (e.g. old OS -> rescue; or rescue -> provisioned OS), making hostname mismatch a reliable signal. In `actionProvisioned`, the node reboots back into the same OS and  the hostname is identical before and after. So the mismatch inference doesn't hold. This can cause unnecessary reboot escalations that interrupt an already in-progress reboot.

Additionally, we have added comments to the `actionProvisioned` function to make it more readable.

Also the field `HetznerBaremetalHost.Spec.Status.Rebooted` has been marked as deprecated as it was superseded by `HetznerBaremetalHost.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID`

ref: [zulip discussion](https://chat.syself.org/#narrow/stream/20-provider.2Fhetzner/topic/actionProvisioned.20-.3E.20handleIncompleteBoot/near/533458)


pkg/services/baremetal/host/host.go
```diff
@@ -2125,6 +2128,8 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 		return actionContinue{delay: 10 * time.Second}
 	}
 
+	// With SSH access enabled, fall back to direct host checks while the BootID still has not
```
janiskemper: This doesn't seem correct. But also the code seems to not make sense. We check for the hostname. How does this help to find out whether the server rebooted successfully or never rebooted in the first place?

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
```diff
@@ -2011,8 +2043,8 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 		return actionError{err: err}
 	}
 
-	// get the current BootId from the wl-cluster. If it has changed, we know that the reboot was
-	// successful.
+	// The BootID is a random UUID the kubelet sets on startup. Comparing the value
+	// before and after a reboot is the way to confirm the node rebooted.
```
janiskemper, line 2047: what would be a use case where the bootID is empty? I think either we should not do anything and just ignore the fact that it is empty (the rest of the code would still work), or we should set a permanent error. WDYT?

pkg/services/baremetal/host/host.go
```diff
@@ -2026,12 +2058,18 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 
 	isRebooted := host.Spec.Status.Rebooted
```
janiskemper: do we still need this, or would it be possible to replace it with the BootID? It seems redundant for me right now

pkg/services/baremetal/host/host.go
```diff
@@ -2081,24 +2122,25 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
-
-		createSSHRebootEvent(ctx, host, msg)
-
 		conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
-			"RebootingMachine",
+			"WaitingForNodeToBeRebooted",
 			clusterv1.ConditionSeverityInfo, "%s",
 			msg)
 		host.Spec.Status.Rebooted = true
 		return actionContinue{delay: 10 * time.Second}
```
janiskemper: should we really reconcile that early? We know that we just triggered a reboot right? It won't reboot in below a minute anyway right?

pkg/services/baremetal/host/host.go
```diff
@@ -2133,25 +2180,27 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
+		// The BootID change (checked at the top of Phase 2) is the sole completion signal.
+		conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
+			"WaitingForBootIDToChange",
+			clusterv1.ConditionSeverityInfo,
+			"Node is up via SSH but BootID has not yet changed")
+		return actionContinue{delay: 10 * time.Second}
 	}
-	// Reboot has been ongoing
+
+	// Hostname did not match: the node is not yet back on the expected OS.
```
janiskemper: this is misleading. Hostname didn't match can also mean that there is no hostname because the server is not reachable (yet) via ssh

pkg/services/baremetal/host/host.go
```diff
@@ -2133,25 +2180,27 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
+	// analyzeSSHOutputProvisioned classifies the SSH output into one of:
+	//   - timeout (SSH connection timed out: node is mid-reboot)
+	//   - connection refused (SSH not listening yet: node is mid-reboot)
+	//   - unexpected hostname / empty stdout / stderr: something is wrong
 	isTimeout, isSSHConnectionRefusedError, err := analyzeSSHOutputProvisioned(out)
 	if err != nil {
 		if errors.Is(err, errUnexpectedHostName) {
```
janiskemper: we should set a permanent error in this case, stating that the hostname is unexpected (nothing more than that)

pkg/services/baremetal/host/host.go
```diff
@@ -2167,6 +2216,8 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 		return actionError{err: err}
 	}
 
+	// handleIncompleteBoot decides whether the reboot is still in progress or has failed (failed=true).
+	// A definitive failure triggers a PermanentError so that the remediation controller can take over.
 	failed, err := s.handleIncompleteBoot(ctx, false, isTimeout, isSSHConnectionRefusedError)
```
janiskemper: this part also works again with timeouts etc right? That seems redundant taking into account that we have a timeout for the whole process already.

are there any benefits for this function apart from the timeouts?

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
Dhairya-Arora01: I searched about this but didn't find anything useful so asked AI - BootID can be empty in these 2 cases
- Node has just registered but the kubelet hasn't yet populated status.nodeInfo . As there's a brief window during initial node registration where the status fields may not be fully reported yet.
- The kubelet is unable to read /proc/sys/kernel/random/boot_id — this is where the kubelet sources the value on Linux. If the file is missing, unreadable, or the proc filesystem isn't mounted properly, it would report an empty string.

In practice, the most common scenario you'd encounter is the first one

I think returning an error (what we currently do) is better. With permanent error - it would get remediated.

pkg/services/baremetal/host/host.go
Dhairya-Arora01: I don't think so it would be replaceable with BootID. If the BootID didn;t changed - would you trigger a reboot ?  It might be that the reboot hasn't been started yet.

pkg/services/baremetal/host/host.go
Dhairya-Arora01: We can increase the delay then to 1 minute.

pkg/services/baremetal/host/host.go
janiskemper: we are doing a remediation already 😅 I think, as we said yesterday already, there is no point in preventing this from being deleted.. It is already in a way - we only try to fix the problem with a reboot.

pkg/services/baremetal/host/host.go
janiskemper: don't we set the bootID exactly when we restart? I don't understand it. Why would it NOT be possible to know based on the BootID (which you also remove after the reboot is done) - if there is a reboot ongoing?

pkg/services/baremetal/host/host.go
Dhairya-Arora01: I would like to know the consequences of this.
If the host is taken by some other cluster, then the node there would be deprovisioned right ?

janiskemper: yeah, but that's not our problem. Another case would be that the user triggered the rescue system over UI and that's why we end up there.

api/v1beta1/hetznerbaremetalhost_types.go
janiskemper, line 334: another thing I realized: Why don't we name it "NodeBootID" and "RebootSince"? Why do we have "annotation" in there?

janiskemper, line 338: IMO we should move this up to the main status. There is no "externalID" in this. It's just a timestamp for the reboot that we trigger without any external systems. 

If you agree with this and the other comment, we can create an issue

pkg/services/baremetal/host/host.go
```diff
@@ -2011,27 +2040,33 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
-		s.scope.Error(err, "")
+		msg := "node.Status.NodeInfo.BootID is empty"
 		conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
 			"NodeInWorkloadClusterHasEmptyBootID",
 			clusterv1.ConditionSeverityWarning, "%s",
-			err.Error())
-		return actionError{err: err}
+			msg)
+
+		s.scope.HetznerBareMetalHost.SetError(infrav1.PermanentError, msg)
```
janiskemper: I just realized that we are already in a remediation process and then set a permanent error. Would this work as expected?

pkg/services/baremetal/host/host.go
```diff
@@ -2151,44 +2150,14 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
+	// command hasn't taken effect yet.
 	conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
 		"WaitingForNodeToBeRebooted",
 		clusterv1.ConditionSeverityInfo,
-		"Waiting for BootID of Node (in wl-cluster) to change (%s)",
-		time.Since(host.Spec.Status.ExternalIDs.RebootAnnotationSince.Time).Round(time.Second))
+		"Waiting for BootID of Node in workload cluster to change",
+	)
 
 	return actionContinue{delay: 10 * time.Second}
```
janiskemper, line 2168: let's open a discussion around this. I find also the 1 min and then 10 sec reconcile a bit random. We should agree also with asking Thomas again. Plz initiate that discussion

api/v1beta1/hetznerbaremetalhost_types.go
Dhairya-Arora01: Because we set it only when reboot happens due to reboot annotation, not on reboot as part of normal provisioning.

janiskemper: okay, understood. But then we would also not need to do it differently compared to previously right? We can use the existing properties also to mark the timeout. AND we could think about always setting the nodeBootID to make it easier to understand. Then we should of course bring back the boolean indicating that the reboot happened

Dhairya-Arora01: okay lets create an issue for this and the comment below than @janiskemper

janiskemper: plz do it then

Dhairya-Arora01: https://github.com/syself/cluster-api-provider-hetzner/issues/1974

janiskemper: Based on that other issue (my comment in it), we would bring back the "reboot" boolean that you marked as deprecated, because we would always set the NodeBootID. If you agree to that, then plz revert the change you made with deprecating the boolean.

janiskemper: #1974

pkg/services/baremetal/host/host.go
```diff
@@ -1985,18 +2011,24 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
 	machine, err := util.GetOwnerMachine(ctx, s.scope.Client, s.scope.HetznerBareMetalMachine.ObjectMeta)
 	if err != nil {
 		err = fmt.Errorf("actionProvisioned (Reboot via Annotation), GetOwnerMachine failed: %w",
 			err)
 		return actionError{err: err}
 	}
 
+	// NodeRef is expected to be set once the Machine has successfully joined the cluster.
+	// If it is still nil at this stage, it likely indicates that the node never registered
+	// (e.g. kubelet failed to join, bootstrap issues, etc.) which we treat as a permanent error.
```
janiskemper: fatal error (check everywhere)

janiskemper (approved):

PR #1946, opened 2026-04-10
Increase boot-to-rescue timeout to 8m

guettli:
Increase the baremetal SSH reboot-to-rescue timeout from 5 minutes to 8 minutes.

Why:
- Hetzner baremetal rescue boots can take around 5 minutes 30 seconds in practice.
- The baremetal host flow currently escalates an SSH-triggered reboot after 5 minutes.
- That can trigger an unnecessary follow-up reboot even though the rescue system becomes reachable shortly after.

What changes:
- Add a dedicated baremetal rescue SSH reboot timeout. Other timeouts are already 10 minutes.
- Add tests that cover the rescue vs non-rescue timeout behavior.

Evidence:
- `caphcli check-bm-servers` reached rescue after about 5 minutes 30 seconds on `bm-e2e-1763731`.

Related: [Increase rescue wait timeout and warn at 80 percent by guettli · Pull Request #3 · syself/caphcli](https://github.com/syself/caphcli/pull/3)

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -137,6 +137,23 @@ func (s *Service) actionPreparing(ctx context.Context) actionResult {
 
 	server, err := s.scope.RobotClient.GetBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID)
 	if err != nil {
+		// If Robot API returned "unauthorized" error - mark condition RobotCredentialsAvailable as false
+		// with reason RobotCredentialsInvalid and stop reconciling.
+		if models.IsError(err, models.ErrorCodeUnauthorized) {
+			msg := "Robot API returned unauthorized; verify the credentials in the referenced secret are correct"
+			conditions.MarkFalse(
+				s.scope.HetznerBareMetalHost,
+				infrav1.RobotCredentialsAvailableCondition,
+				infrav1.RobotCredentialsInvalidReason,
+				clusterv1.ConditionSeverityError,
+				"%s",
+				msg,
+			)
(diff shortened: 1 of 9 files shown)


janiskemper (changes requested):
there are multiple other occurrences of "6*time.Minute" with what looks like the same use case. 

I vote for creating a variable so that this is transparent in the future, which values belong together.

pkg/services/baremetal/host/host.go
```diff
@@ -413,7 +414,11 @@ func (s *Service) handleErrorTypeSSHRebootFailed(ctx context.Context, isSSHTimeo
 	if wantsRescue {
 		rebootInto = "rescue mode"
 	}
-	if !isSSHTimeoutError || hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.LastUpdated, sshResetTimeout) {
+	sshRebootTimeout := sshResetTimeout
```
janiskemper: why do you add this new code here? Why would a reset to rescue take longer than others? Why don't we just increase the timeout of the ssh reset in general?

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
janiskemper: I think this line can be removed

janiskemper: @guettli

guettli: @janiskemper Good idea, then then PR gets very small. Just increase the integer.

janiskemper (approved):

PR #1949, opened 2026-04-10
:seedling: Move SSH private key availability check before creating the server

Dhairya-Arora01:
Move the SSH private key availability preflight check before creating the server.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -230,7 +230,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		return reconcile.Result{RequeueAfter: requeueImmediately}, nil
 	}
 
-	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.                    ───────
+	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
 	// Check that we have valid SSH private key in the secret. A failure could also mean there is a
 	// network failure while trying to access the api-server.
 	if hm.Spec.ImageURL != "" {
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -230,18 +230,20 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		return reconcile.Result{RequeueAfter: requeueImmediately}, nil
 	}
 
-	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.                    ───────
 	// Check that we have valid SSH private key in the secret. A failure could also mean there is a
 	// network failure while trying to access the api-server.
-	_, err := s.getSSHPrivateKey(ctx)
-	if err != nil {
-		s.scope.Error(err, "")
-		if errors.Is(err, errSSHKeyMisconfigured) {
-			return reconcile.Result{}, nil
+	if hm.Spec.ImageURL != "" {
+		_, err := s.getSSHPrivateKey(ctx)
+		if err != nil {
+			s.scope.Error(err, "")
+			if errors.Is(err, errSSHKeyMisconfigured) {
+				return reconcile.Result{}, nil
+			}
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
 		}
-		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
+		v1beta1conditions.MarkTrue(s.scope.HCloudMachine, infrav1.SSHPrivateKeyAvailableCondition)
 	}
-	v1beta1conditions.MarkTrue(s.scope.HCloudMachine, infrav1.SSHPrivateKeyAvailableCondition)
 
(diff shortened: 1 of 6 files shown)


pkg/services/hcloud/server/server.go
janiskemper: another point I had was that we check here whether the secret exists. In case we get a not found, we should also show that in the preflight check and stop

pkg/services/hcloud/server/server.go
```diff
@@ -233,6 +233,19 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		return reconcile.Result{}, nil
 	}
 
+	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+	// Check that we have valid SSH private key in the secret. A failure could also mean there is a
+	// network failure while trying to access the api-server.
+	_, err := s.getSSHPrivateKey(ctx)
+	if err != nil {
+		err = fmt.Errorf("getSSHPrivateKey failed: %w", err)
```
janiskemper: we should check for a more specific error. In case the secret is empty or doesn't contain proper information, we should maybe say that we requeue (or also not? In other cases, e.g. with the hcloud token, we stop requeuing). 

In case it is a normal API error, we should return the error.

janiskemper (changes requested):

pkg/services/hcloud/server/server.go
Dhairya-Arora01: Okay,
We already check that the secret exists and also if the privateKey specified in the secret is empty. On such an error we reque.

IMO, we should introduce a new condition here.
In case of an error in getSSHPrivateKey, we set condition `ServerAvailable` as false with reason `GetSSHPrivateKeyFailed` - The condition is very generic but the reason is associated more with the code.

Instead we can set a condition like `SSHPrivateKeyAvailable` with a reason like `InvalidSSHPrivateKey` or `SSHPrivateKeyNotFound` etc.
WDYT ?

janiskemper: okay

pkg/services/hcloud/server/server.go
```diff
@@ -222,6 +222,34 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
+		// No need for Requeue, because adding the command line argument to the caph deployment,
+		// will restart the controller, and all resources will be reconciled.
+		return reconcile.Result{}, nil
+	}
+
+	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+	// Check that we have valid SSH private key in the secret. A failure could also mean there is a
+	// network failure while trying to access the api-server.
+	_, err := s.getSSHPrivateKey(ctx)
+	if err != nil {
```
janiskemper, line 229: there might be multiple possible root causes for this. Shouldn't we distinguish them?

janiskemper (changes requested):

guettli:
@Dhairya-Arora01 Is this PR still valid? I think that PR solved it: [:seedling: When image-url-command is missing, don't create server. by guettli · Pull Request #1995 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1995)

Dhairya-Arora01:
> @Dhairya-Arora01 Is this PR still valid? I think that PR solved it: [🌱 When image-url-command is missing, don't create server. by guettli · Pull Request #1995 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1995)

@guettli , the title for this PR is currently misleading (I will update that). So with this PR, I want to add a preflight check that SSH private key must be available before creating the server. Currently, we do this check after creating the server.

pkg/services/hcloud/server/server.go
```diff
@@ -1599,9 +1595,17 @@ func updateHCloudMachineStatusFromServer(hm *infrav1.HCloudMachine, server *hclo
 
 // getSSHPrivateKey retrieves the SSH private key used for connecting to the rescue systems.
 // It reads the key from the Kubernetes secret referenced by HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+// On failure it sets SSHPrivateKeyAvailableCondition with a specific reason describing the root cause.
 func (s *Service) getSSHPrivateKey(ctx context.Context) (string, error) {
 	robotSecretName := s.scope.HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.Name
 	if robotSecretName == "" {
+		v1beta1conditions.MarkFalse(
```
janiskemper, line 1609: if this happens, then it doesn't make sense to requeue at all - what we do right now though for one min

janiskemper (approved):

pkg/services/hcloud/server/server.go
```diff
@@ -230,18 +230,20 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		return reconcile.Result{RequeueAfter: requeueImmediately}, nil
 	}
 
-	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+	// Fetch the SSH private key from the secret referenced in HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.                    ───────
```
janiskemper: I think there is something wrong here

janiskemper (approved):

PR #1950, opened 2026-04-11
🌱 Split v1beta1 ServerAvailable into ServerProvisioned and ServerAvailable conditions for HCloudMachine

### LINKED ISSUE #1947: Split v1beta1 `ServerAvailable` condition into `ServerProvisioned` and `ServerAvailable` for HCloudMachine
base branch: v1.1.x

## Summary

Split the existing `ServerAvailable` condition into `ServerProvisioned` (tracks provisioning) and `ServerAvailable` (tracks post-provisioning health).

## The Problem

We handle the HCloudMachine provisioning steps manually through a boot state machine. The HCloud server creation (API call) happens early, and afterwards there is a whole provisioning process: enabling rescue, SSH, running image commands, booting the OS.

The current `ServerAvailable` condition covers both provisioning progress and post-provisioning health. When a user sees `ServerAvailable=False`, they cannot tell whether the server is still being provisioned or whether it was provisioned and then hit a problem (server not found, network attachment failure, load balancer issue, server not found).

## The Solution

Split the current `ServerAvailable` condition to match the actual lifecycle:

* **`ServerProvisioned`**: This condition tracks the full provisioning state machine. It becomes `True` when the OS is running and the machine reaches `OperatingSystemRunning` boot state (before network/LB attachment).
* ** `ServerAvailable`**: This condition will now track post-provisioning health. It becomes `True` after network attachment and load-balancer attachment succeed. It can go `False` if the server is powered off, network fails, or LB detaches after initial provisioning.

This gives users three distinct signals:
| Condition | Meaning |
|-----------|---------|
| `ServerCreated=True` | HCloud API call succeeded, server exists |
| `ServerProvisioned=True` | Full boot state machine complete, OS is running |
| `ServerAvailable=True` | Network and LB attached, machine is fully operational |

abdullah599:
Splits the overloaded v1beta1 `ServerAvailableCondition` into two distinct conditions:
- `ServerProvisionedCondition`: tracks boot time provisioning progress (rescue boot, image install, OS startup)
- `ServerAvailableCondition`: indicates the server is running and ready (network attached, LB attached)
This ensures provisioning status is tracked separately from post provisioning availability, and that post provisioning operations never mutate the provisioned condition.

Fixes #1947

- `ServerProvisionedCondition` is only set to `False` during provisioning and `True` once the OS is running
- Post provisioning code (`handleOperatingSystemRunning`, delete path) only touches `ServerAvailableCondition`

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -197,7 +197,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 			return reconcile.Result{}, err
 		}
 		s.scope.Error(nil, msg)
-		conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
+		conditions.MarkFalse(hm, infrav1.ServerCreateSucceededCondition,
 			"HandleBootStateUnsetTimedOut", clusterv1.ConditionSeverityWarning,
 			"%s", msg)
 		return reconcile.Result{}, nil
@@ -216,9 +216,6 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		msg = fmt.Sprintf("Updating old resource (pre BootState) %s", hm.Status.BootState)
 
 		s.scope.Info(msg)
-		conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
-			"HandleBootStateUnset", clusterv1.ConditionSeverityInfo,
-			"%s", msg)
 		return reconcile.Result{RequeueAfter: requeueImmediately}, nil
 	}
 
@@ -256,7 +253,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		if errors.Is(err, errServerCreateNotPossible) {
 			err = fmt.Errorf("createServerFromImageNameOrURL failed: %w", err)
 			s.scope.Error(err, "")
-			conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
+			conditions.MarkFalse(hm, infrav1.ServerCreateSucceededCondition,
 				"ServerCreateNotPossible", clusterv1.ConditionSeverityWarning,
 				"%s", err.Error())
 			return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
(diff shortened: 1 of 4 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -256,7 +256,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		if errors.Is(err, errServerCreateNotPossible) {
 			err = fmt.Errorf("createServerFromImageNameOrURL failed: %w", err)
 			s.scope.Error(err, "")
-			conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+			conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
```
janiskemper: this should be "server created" actually.. I think it was wrong before no?

pkg/services/hcloud/server/server.go
```diff
@@ -197,7 +197,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 			return reconcile.Result{}, err
 		}
 		s.scope.Error(nil, msg)
-		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+		conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
```
janiskemper: this should also be "server create" in theory, no? The pretty much only case is that the create timed out for some reason..

pkg/services/hcloud/server/server.go
```diff
@@ -216,7 +216,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		msg = fmt.Sprintf("Updating old resource (pre BootState) %s", hm.Status.BootState)
 
 		s.scope.Info(msg)
-		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+		conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
```
janiskemper: I would remove this condition. It is misleading and shouldn't ever be visible.

pkg/services/hcloud/server/server.go
```diff
@@ -741,13 +741,14 @@ func (s *Service) handleBootingToRealOS(ctx context.Context, server *hcloud.Serv
...
 	case hcloud.ServerStatusStarting, hcloud.ServerStatusInitializing:
-		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+		conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
 			"BootingToRealOS", clusterv1.ConditionSeverityInfo,
 			"Operating system of node is booting")
 		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 
 	case hcloud.ServerStatusRunning:
 		hm.SetBootState(infrav1.HCloudBootStateOperatingSystemRunning)
+		conditions.MarkTrue(hm, infrav1.ServerProvisionedCondition)
```
janiskemper, line 748: It seems pretty useless to set the status here on false (below this line) right?

abdullah599: yes

janiskemper (approved):

PR #1953, opened 2026-04-13
:seedling: Compare v1.1.x to main

guettli:
Compare v1.1.x to main.

Ignore changes to vendor directory.

## Related

[:seedling: Set v1.1.x to main branch, and release (Overview)](https://github.com/syself/cluster-api-provider-hetzner/pull/1781)


controllers/hcloudmachine_controller.go
```diff
@@ -175,10 +239,16 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 		return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
-	if !hcloudMachine.ObjectMeta.DeletionTimestamp.IsZero() {
+	if !hcloudMachine.DeletionTimestamp.IsZero() {
 		return r.reconcileDelete(ctx, machineScope)
 	}
 
+	if hcloudMachine.Status.BootState == infrav1.HCloudBootStateProvisioningFailed {
```
janiskemper, line 246: why don't we do this in the normal reconcile in server.go? I don't see a benefit of having it here

pkg/services/hcloud/server/server.go
```diff
@@ -91,135 +101,682 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			}
+
+			return reconcile.Result{}, fmt.Errorf("findServer: %w", err)
+		}
+
+		conditions.MarkTrue(s.scope.HCloudMachine, infrav1.HCloudTokenAvailableCondition)
+
+		// findServer will return both server and error as nil, if the server was not found.
+		if server == nil {
+			// The server did disappear in HCloud? Maybe it was delete via web-UI.
```
janiskemper, line 138: let's improve the wording. If the server was not found, someone deleted it. We can just write it in a straightforward way

pkg/services/hcloud/server/server.go
```diff
@@ -91,135 +101,682 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			// We set MachineError. CAPI will delete machine.
+			msg := fmt.Sprintf("hcloud server (%q) no longer available. Setting MachineError.",
+				*s.scope.HCloudMachine.Spec.ProviderID)
+
+			s.scope.Error(errors.New(msg), msg,
+				"ProviderID", *s.scope.HCloudMachine.Spec.ProviderID,
+				"BootState", s.scope.HCloudMachine.Status.BootState,
```
janiskemper, line 145: boot state seems irrelevant in this case. Also bootStateSince. We can remove it from the log

pkg/services/hcloud/server/server.go
```diff
@@ -91,135 +101,682 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+			"GetSSHClientFailed", clusterv1.ConditionSeverityWarning,
+			"%s", err.Error())
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	}
 
-			return reconcile.Result{}, fmt.Errorf("failed to create server: %w", err)
+	err = sshClient.Reboot().Err
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) {
+			// ssh connection refused is common while the rescue system starts.
```
janiskemper, line 496: what do you mean "when the rescue system starts"? let's write a better comment before this code block describing what you do. 

I understand the following: We enable the rescue system in Hetzner via API. For some reason that I don't understand this leads to ssh connections failing with "connection refused" - independent of the reboot. 

Is that the case?

pkg/services/hcloud/server/server.go
```diff
@@ -91,135 +101,682 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 		}
+		s.scope.Error(nil, msg)
+		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+			"BootingToRescueTimedOut", clusterv1.ConditionSeverityWarning,
+			"%s", msg)
+		return reconcile.Result{}, nil
 	}
 
-	s.scope.SetProviderID(server.ID)
+	if server.RescueEnabled {
```
janiskemper, line 542: why do we do this? Can you explain it? We should ideally also change the reason and message of the condition to something more understandable. Why would we want it to be false?

pkg/services/hcloud/server/server.go
```diff
@@ -91,135 +101,682 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
+
+	switch state {
+	case sshclient.ImageURLCommandStateRunning:
+		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
+			"HCloudImageURLCommandRunning", clusterv1.ConditionSeverityInfo,
+			"hcloud-image-url-command running")
+		return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+
+	case sshclient.ImageURLCommandStateFinishedSuccessfully:
+		record.Event(hm, "ImageURLCommandSuccessful", logFile)
```
janiskemper, line 679: do we really wanna do this when it was successful in a prod system?

janiskemper (commented):
let's implement the comments in another PR after we agree on it

api/v1beta1/hcloudremediation_webhook.go
```diff
@@ -17,42 +17,50 @@ limitations under the License.
...
-func (r *HCloudRemediation) ValidateUpdate(runtime.Object) (admission.Warnings, error) {
+// ValidateUpdate implements webhook.CustomValidator so a webhook will be registered for the type.
+func (*hcloudRemediationWebhook) ValidateUpdate(_ context.Context, _ runtime.Object, _ runtime.Object) (admission.Warnings, error) {
 	return nil, nil
 }
 
-// ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
-func (r *HCloudRemediation) ValidateDelete() (admission.Warnings, error) {
+// ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type.
+func (*hcloudRemediationWebhook) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
```
Dhairya-Arora01, line 64: nit: can we just remove "_" as we are not using the parameters just to remove the visual noise.
Same in other webhook files as well.

api/v1beta1/hetznerbaremetalmachine_validation.go
```diff
@@ -86,5 +86,14 @@ func validateHetznerBareMetalMachineSpecUpdate(oldSpec, newSpec HetznerBareMetal
 		)
 	}
 
+	if oldSpec.ProviderID != nil && *oldSpec.ProviderID != "" {
+		// once the ProviderID was set, the value must not change.
+		if newSpec.ProviderID == nil || *oldSpec.ProviderID != *newSpec.ProviderID {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec", "providerID"), newSpec.ProviderID, "providerID immutable"),
```
Dhairya-Arora01, line 93: Instead of field.Invalid error lets return field.Forbidden for immutability checks.

Aligned with how upstream K8s does this. Ex
```
Pod "nginx-ingress-controller-w2cgl" is invalid: spec: Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds` or `spec.tolerations` (only additions to existing tolerations)
```

api/v1beta1/hetznercluster_webhook.go
```diff
@@ -151,14 +163,18 @@ func isNetworkZoneSameForAllRegions(regions []Region, defaultNetworkZone *string
 	return nil
 }
 
-// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
-func (r *HetznerCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
+// ValidateUpdate implements webhook.CustomValidator so a webhook will be registered for the type.
+func (*hetznerClusterWebhook) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+	r, ok := newObj.(*HetznerCluster)
+	if !ok {
+		return nil, fmt.Errorf("expected an HetznerCluster object but got %T", r)
```
Dhairya-Arora01, line 170: shouldn't we return apierrors.NewBadRequest here as well, if we are returning it below for the oldObj

Same for other webhooks....

api/v1beta1/types.go
```diff
@@ -113,18 +127,38 @@ type HetznerSecretRef struct {
...
+	// workload-cluster secret named "hcloud", and only if this field is not already "token". If
+	// the configured secret name is already "hcloud", the same token value is stored under both
+	// keys in that single secret. We recommend to use "token".
+	//
+	// +optional +kubebuilder:default=hcloud-token
 	HCloudToken string `json:"hcloudToken"`
-	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is stored.
+
+	// HetznerRobotUser defines the name of the key where the username for the Hetzner Robot API is
+	// stored.  It gets used for reading the credential in the mgt-cluster, and it gets used for
```
Dhairya-Arora01, line 142: nit: Double spaces between "." and "It"

controllers/hetznercluster_controller.go
```diff
@@ -508,73 +522,121 @@ func reconcileTargetSecret(ctx context.Context, clusterScope *scope.ClusterScope
...
+	if secretName != "hcloud" {
+		names = append(names, "hcloud")
+	}
+	return names
+}
+
+// workloadClusterCompatibilityKeys returns the configured key for every secret.
+// When reconciling the secret named "hcloud", it also returns the compatibility
+// key unless it matches the configured key.
+func workloadClusterCompatibilityKeys(secretName, configuredKey, compatibilityKey string) []string {
```
Dhairya-Arora01, line 553: The term compatibility in the function name seems weird to me. Can we rename this to `keysForWorkloadClusterSecret`

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -703,3 +755,131 @@ func (c *sshClient) ExecutePreProvisionCommand(ctx context.Context, command stri
...
+	}()
+
+	scpClient, err := scp.NewClientBySSH(client)
+	if err != nil {
+		return 0, "", fmt.Errorf("couldn't create a new scp client: %w", err)
+	}
+
+	defer scpClient.Close()
+
+	if command == "" {
```
Dhairya-Arora01, line 789: Shouldn't we move this guard to top, before we create the ssh client

pkg/services/baremetal/client/ssh/ssh_client.go
```diff
@@ -703,3 +755,131 @@ func (c *sshClient) ExecutePreProvisionCommand(ctx context.Context, command stri
...
+	err = scpClient.CopyFromFile(ctx, *fdCommand, dest, "0700")
+	if err != nil {
+		return 0, "", fmt.Errorf("error copying file %q to %s:%d:%s %w", command, c.ip, c.port, dest, err)
+	}
+
+	reader := bytes.NewReader(bootstrapData)
+	dest = "/root/bootstrap.data"
+	err = scpClient.CopyFile(ctx, reader, dest, "0700")
+	if err != nil {
+		return 0, "", fmt.Errorf("error copying boostrap data to %s:%d:%s %w", c.ip, c.port, dest, err)
```
Dhairya-Arora01, line 814: typo boostrap

pkg/services/baremetal/host/host.go
```diff
@@ -1864,26 +2200,56 @@ func (s *Service) actionDeprovisioning(_ context.Context) actionResult {
...
-			IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
-		})
-		out := sshClient.ResetKubeadm()
-		s.scope.V(1).Info("Output of ResetKubeadm", "stdout", out.StdOut, "stderr", out.StdErr, "err", out.Err)
-		if out.Err != nil {
-			record.Warnf(s.scope.HetznerBareMetalHost, "FailedResetKubeAdm", "failed to reset kubeadm: %s", out.Err.Error())
+	conditions.MarkTrue(s.scope.HetznerBareMetalHost, infrav1.RobotCredentialsAvailableCondition)
+
+	if s.scope.SSHAfterInstallImage {
+		// If has been provisioned completely, stop all running pods
```
Dhairya-Arora01, line 2236: If *it* has been provisioned completely

pkg/services/hcloud/server/server.go
Dhairya-Arora01: @guettli ?

guettli: @Dhairya-Arora01 @janiskemper 


I am looking at:

```go 

	// Now we know that the rescue-system was enabled. Up to now the PreRescueOS is running. Next
	// step is to reboot the server into the rescue system.

	// Reboot via ssh, avoid API calls to hcloud (rate-limit)
	sshClient, err := s.getSSHClient(ctx)
	if err != nil {
		err = fmt.Errorf("getSSHClient failed: %w", err)
		s.scope.Error(err, "")
		v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
			"GetSSHClientFailed", clusterv1beta1.ConditionSeverityWarning,
			"%s", err.Error())
	}

	err = sshClient.Reboot().Err
	if err != nil {
		if errors.Is(err, syscall.ECONNREFUSED) {
			// ssh connection refused is common while the rescue system starts.
			// Provide a nice message.
			v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
				"RetryingSSHConnection",
				clusterv1beta1.ConditionSeverityInfo, "Rebooting")
		}

		err = fmt.Errorf("reboot to rescue: reboot via ssh failed: %w", err)
		s.scope.Error(err, "")
		v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
			"RebootViaSSHFailed",
			clusterv1beta1.ConditionSeverityWarning, "%s", err.Error())
	}

```

Overall goal: The hcloud machine is currently in the preRescueOS (aka Vanilla Ubuntu).
The hcloud api to enable the rescue system on the next boot was already set. Now we need to reboot the server. We do that via ssh to save api requests.

ECONNREFUSED is overall a good signal. We know that there is something reachable. The server is alive. That's why an error message like "... failed" gets avoided. "Rebooting" matches better.

@janiskemper does this answer your question?

---

BTW, looking at my own code after some weeks. I think every handleXXX() function should have a better comment at the top.

janiskemper: @guettli it doesn't fully answer my question. I basically want to understand how this situation looks like. Why do we expect that there is "connection refused" after activating rescue via api and before rebooting via ssh? Why can we confidently say that this "connection refused" error is gone after a few seconds, so that we can finally trigger the reboot via ssh?

I only want to understand it to then write better comments in the code.

guettli: @janiskemper in the previous step, then controller waits until hcloud api says that the server is ready. But it can take some time until ssh is reachable.

See the code in the above function:

```go 
	// analyze status of server
	switch server.Status {
	case hcloud.ServerStatusStarting, hcloud.ServerStatusInitializing:
		v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
			"ServerNotRunningYet", clusterv1beta1.ConditionSeverityInfo,
			"hcloud server is %q", server.Status)
	case hcloud.ServerStatusRunning:
		// execute below code
	default:
		// some temporary status
		s.scope.Info("Unknown hcloud server status", "status", server.Status)
		v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
			"UnknownHCloudStatus", clusterv1beta1.ConditionSeverityInfo, "hcloud server has unknown status: %q", server.Status)
	}
```

There is a delay between ServerStatusRunning and ssh reachable. And in that phase Connection Refused is common.

janiskemper: ah okay. Perfect, then let's document this @Dhairya-Arora01 :)

guettli:
Reset of `main` was done, this is no longer needed.

PR #1954, opened 2026-04-13
:seedling: Move hcloud image-url command into HCloudMachine spec

guettli:
## What changed

This moves hcloud custom image provisioning from the controller CLI into the `HCloudMachine` API.

- add `spec.imageURLCommand` to `HCloudMachineSpec`
- make hcloud provisioning read the command from `HCloudMachine.spec`
- remove the `--hcloud-image-url-command` controller flag and its wiring
- regenerate the hcloud CRDs
- update the image-url-command and HCloudMachineTemplate docs

## Related BM PR

#1955

PR #1955, opened 2026-04-13
:seedling: Move baremetal image-url command into HetznerBareMetalMachine spec

guettli:
## What changed

This moves baremetal custom image provisioning from the controller CLI into the `HetznerBareMetalMachine` API.

- add `spec.installImage.imageURLCommand` to `InstallImage`
- make baremetal provisioning read the command from `HetznerBareMetalMachine.spec`
- align the field naming with `#1954` by using `imageURLCommand`
- remove the `--baremetal-image-url-command` controller flag and its wiring
- regenerate the baremetal CRDs
- update the image-url-command and HetznerBareMetalMachineTemplate docs

## Related HCloud PR

#1954


api/v1beta1/hetznerbaremetalmachine_types.go
```diff
@@ -157,6 +157,18 @@ type InstallImage struct {
...
+	// provisions a machine from Image.URL. CAPH copies that command into the rescue system and
+	// executes it there.
+	//
+	// Docs: https://syself.com/docs/caph/developers/image-url-command
+	//
+	// ImageURLCommand must be set if the machine should be provisioned from Image.URL without
+	// installimage. The command name must start with image-url-command-.
+	// +kubebuilder:validation:Optional
+	// +optional
+	ImageURLCommand string `json:"imageURLCommand,omitempty"`
```
janiskemper, line 170: stupid question: But why do we call it ImageURLCommand?

pkg/services/baremetal/host/host.go
```diff
@@ -1252,15 +1254,25 @@ func (s *Service) actionImageInstallingCustomImageURLCommand(ctx context.Context
...
+		if command == "" {
+			err = errors.New("internal error: spec.status.installImage.imageURLCommand is not set")
 			s.scope.Error(err, "")
 			conditions.MarkFalse(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition,
 				"ImageURLCommandMissing",
 				clusterv1.ConditionSeverityError,
 				"%s", err.Error())
-			// this can only be changed by updating the controller. This will make the
-			// controller reconcile all resources.
+			return actionContinue{delay: time.Hour}
```
janiskemper: we shouldn't continue here.. This is something that cannot be fixed. Same below. 

But also: this cannot happen because of the webhooks. Therefore we have to remove this whole code block IMO. 

I don't wanna keep it for sure, because otherwise we need to properly handle it and there is no point.

janiskemper (changes requested):

api/v1beta1/hetznerbaremetalmachine_validation.go
```diff
@@ -18,32 +18,70 @@ package v1beta1
...
-		spec.InstallImage.Image.Path == "" {
-		allErrs = append(allErrs,
-			field.Invalid(field.NewPath("spec", "installImage", "image"), spec.InstallImage.Image,
-				"have to specify either image name and url or path"),
-		)
-	}
+	if installImage.UsesImageURLCommand() {
+		if image.URL == "" {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec", "installImage", "image", "url"), image.URL,
```
Dhairya-Arora01, line 40: shouldn't this be field.Required ?

api/v1beta1/hetznerbaremetalmachine_types.go
guettli: It is the command which receives the image-url as argument and then fetches the image.

Do you want to change it?

If you want a better comment, then tell me what text would make sense here.

pkg/services/baremetal/host/host.go
guettli: @janiskemper ok, using `actionStop{}` now.

api/v1beta1/hetznerbaremetalmachine_validation.go
guettli: @Dhairya-Arora01 

looking at:

```go 
func validateHetznerBareMetalMachineSpecCreate(spec HetznerBareMetalMachineSpec) field.ErrorList {
	var allErrs field.ErrorList
	installImage := spec.InstallImage
	image := installImage.Image

	if installImage.UsesImageURLCommand() {
		if image.URL == "" {
			allErrs = append(allErrs,
				field.Invalid(field.NewPath("spec", "installImage", "image", "url"), image.URL,
					"url is required when imageURLCommand is set"),
			)
		} 
...
```

It  (image.url) is not always required, only if UsesImageURLCommand() is true ... Or I don't understand your question.

api/v1beta1/hetznerbaremetalmachine_types.go
janiskemper: I was thinking about the name in general... If you have a better name, then good. Otherwise ignore it

guettli: I dont have a better name.

api/v1beta1/hetznerbaremetalmachine_validation.go
Dhairya-Arora01: I was talking about the error type - currently we return field.Invalid , IMO we should return field.Required. But this case is conditional - so return field.Required might not be the best option. Okay then I agree with you that we can return field.Invalid error

Dhairya-Arora01 (approved):
LGTM, though you need to rebase

janiskemper (approved):

PR #1958, opened 2026-04-13
:seedling: Fix unit-tests on v1.1.x.

guettli:
Fix unit-tests on v1.1.x.

Failed before:

> Summarizing 2 Failures:
  [FAIL] HCloudMachineReconciler Basic hcloudmachine test wrong server [It] checks that ImageNotFound is visible in conditions if image does not exist
  /home/guettli/syself/caph-fix-unit-tests/controllers/hcloudmachine_controller_test.go:518
  [FAIL] HCloudMachineReconciler various specs without placement groups, but with placement group in hcloudMachine spec [It] should show the expected reason for server not created
  /home/guettli/syself/caph-fix-unit-tests/controllers/hcloudmachine_controller_test.go:625
> 
> Ran 119 of 119 Specs in 60.768 seconds
FAIL! -- 117 Passed | 2 Failed | 0 Pending | 0 Skipped

- Fixed the test, by not overwriting the condition.
- Updated the test to provide a better err msg, when the test fails.
- Provide better err msg, when running envtest fails. For example, when running like `DEBUG=1 ginkgo run --focus "without placement groups"    ./controllers/... | ./hack/filter-caph-controller-manager-logs.py -`


pkg/services/hcloud/server/server.go
```diff
@@ -255,9 +255,12 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 		if errors.Is(err, errServerCreateNotPossible) {
 			err = fmt.Errorf("createServerFromImageNameOrURL failed: %w", err)
 			s.scope.Error(err, "")
-			conditions.MarkFalse(hm, infrav1.ServerCreateSucceededCondition,
-				"ServerCreateNotPossible", clusterv1.ConditionSeverityWarning,
-				"%s", err.Error())
+			if !conditions.IsFalse(hm, infrav1.ServerCreateSucceededCondition) {
```
Dhairya-Arora01: This `if` conditional seems to be unnecessary.

test/helpers/envtest.go
```diff
@@ -221,6 +222,21 @@ func NewTestEnvironment() *TestEnvironment {
...
 }
 
+func wrapEnvtestStartError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	errText := err.Error()
+	if !strings.Contains(errText, "$PATH") {
+		// This is not: unable to start control plane itself: failed to start the controlplane.
```
Dhairya-Arora01: The comment seems weird with ":" and the grammar particularly.

guettli: @Dhairya-Arora01 please check again.

pkg/services/hcloud/server/server.go
janiskemper: is this relevant? To me it sounds like we should stick to one version (in the utility function) or here, but not both

controllers/hcloudmachine_controller_test.go
```diff
@@ -513,9 +513,29 @@ var _ = Describe("HCloudMachineReconciler", func() {
 			})
 
 			It("checks that ImageNotFound is visible in conditions if image does not exist", func() {
-				Eventually(func() bool {
-					return isPresentAndFalseWithReason(key, hcloudMachine, infrav1.ServerCreateSucceededCondition, infrav1.ImageNotFoundReason)
-				}, timeout, interval).Should(BeTrue())
+				Eventually(func() error {
```
janiskemper: why do we do this? If the utility function has some issues, we oculd also solve it there probably no? I'm not able to judge on first glance what changed.

guettli: @janiskemper before we got an meaningless "false is not true" error. We get much better error now, when the test fails:

> expected ServerCreateSucceeded to be False with reason "ImageNotFound", 
> 
> got status="False" reason="ServerCreateNotPossible" message="createServerFromImageNameOrURL failed: create server from imageName (\"my-control-plane-2\"): no image found with name my-control-plane-2: server create not possible - need action"

---


For new code I prefer `Eventually(func() error` to `Eventually(func() bool`

---

But I can revert that, too. Or we write a helper which makes getting a good error message easier.

janiskemper: well this applies to all cases where we use this utility function. Do we have an idea how to improve that?

pkg/services/hcloud/server/server.go
guettli: @janiskemper what is the desired result?

Currently (pre PR):

> reason="ServerCreateNotPossible" message="createServerFromImageNameOrURL failed: create server from imageName ("my-control-plane-2"): no image found with name my-control-plane-2: server create not possible - need action"


But test expected reason="ImageNotFound".

---

I slightly prefer ImageNotFound, but the more general reason is fine, too.

janiskemper: I also do that. So we should not overwrite it. But my point is that we should find a better solution ideally, not this if statement. @Dhairya-Arora01 can also look into this if you want.

guettli: @janiskemper what about merging it like it is now, and make it nice in a second PR. The CI fails for other PRs now, and I would like to fix v1.1.x soon.

controllers/hcloudmachine_controller_test.go
guettli: Added a reusable helper in `test/helpers/conditions.go` and switched only this `Eventually` to use it.

I reverted the broader rollout so this PR stays scoped to the original failing case. If we want to apply the same pattern to the other `isPresentAndFalseWithReason` usages, I can do that in a separate follow-up PR.

pkg/services/hcloud/server/server.go
janiskemper: @abdullah599 then plz look into that. If we have another issue, we can merge this

controllers/hcloudmachine_controller_test.go
janiskemper: @guettli can you plz write down properly what your problem is with the isPresentAndFalseWithReason utility and then why you wanna change it and what you propose? I think that we either properly change the original utility function, we introduce a separate utility function, or we do it differently in this one case. 

If no change is needed here for this PR, I would leave the topic, revert the changes, and create an issue for it.

pkg/services/hcloud/server/server.go
abdullah599: https://github.com/syself/cluster-api-provider-hetzner/pull/1935/changes/6d2a932db84217508baad3b260229a1245a88a11

I did this (i'll revert it). I did not know we fixed it here

reason: 
ServerCreateSucceededCondition are already set with the specific reason by the inner function that returned errServerCreateNotPossible. imo we must not overwrite them here.


The guard isn't needed imo. Every code path that returns `errServerCreateNotPossible` already sets `ServerCreateSucceededCondition` to False with a specific reason before returning:

1. `createServer` → `InstanceHasNonExistingPlacementGroupReason`
2. `getSSHKeys` → `SSHKeyNotFoundReason`
3. `getServerImage` → `ServerTypeNotFoundReason`
4. `getServerImage` → `ImageAmbiguousReason`
5. `getServerImage` → `ImageNotFoundReason`

So `conditions.IsFalse(hm, infrav1.ServerCreateSucceededCondition)` is always true when this handler runs.

janiskemper: so we can just remove this right?

controllers/hcloudmachine_controller_test.go
guettli: @janiskemper I extracted that to an own PR: [:seedling: test: use ConditionFalseWithReasonAtKey in controllers by guettli · Pull Request #1962 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/1962)

pkg/services/hcloud/server/server.go
guettli: @abdullah599 yes, maybe it works for now to remove that block completely. But who know if the future paths all set a condition. I recommend to leave it like this.

janiskemper: @guettli I wouldn't leave it like this, because the problem shouldn't exist in the first place, if this code part is removed.

guettli: done.

janiskemper (commented):
LGTM from my side. @abdullah599 is that what you had in mind as well?

abdullah599 (approved):
> LGTM from my side. @abdullah599 is that what you had in mind as well?

Yes :)

PR #1962, opened 2026-04-14
:seedling: test: use ConditionFalseWithReasonAtKey in controllers

guettli:
use ConditionFalseWithReasonAtKey in controllers

In tests, the failure message of an Eventually() is usualy not easy to understand, when the function returns `bool`. It is easier to understand, when the Eventually checks for the absencse of an error.


Dhairya-Arora01 (commented):
We should use `ConditionFalseWithReasonAtKey` func at other places like server_test.go as well.

PR #1965, opened 2026-04-14
:seedling: Move SSHAfterInstallImage from cli argument to HetznerBaremetalMachine

guettli:
Moves the SSH-after-install-image setting from the controller CLI into HetznerBareMetalMachine.spec.sshSpec.noSSHAfterInstallImage. 

This makes the behavior configurable per HBMM or template while keeping the current default behavior, and updates the CRDs and tests accordingly.


pkg/scope/baremetalhost.go
```diff
@@ -170,3 +167,14 @@ func (s *BareMetalHostScope) hasConstantHostname() bool {
 	return s.Cluster.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true" ||
 		s.HetznerBareMetalMachine != nil && s.HetznerBareMetalMachine.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true"
 }
+
+// SSHAfterInstallImageEnabled returns the effective SSH-after-installimage setting for the host.
+func (s *BareMetalHostScope) SSHAfterInstallImageEnabled() bool {
+	if s.HetznerBareMetalHost.Spec.Status.SSHSpec != nil {
```
janiskemper: why do you have these two if statements? AFAIK the business logic would make it impossible to have one but not the other

guettli: @janiskemper good question. I simplified it to:

```go 
// SSHAfterInstallImageEnabled returns the effective SSH-after-installimage setting for the host.
func (s *BareMetalHostScope) SSHAfterInstallImageEnabled() bool {
	return !s.HetznerBareMetalMachine.Spec.SSHSpec.NoSSHAfterInstallImage
}
```

janiskemper: since we currently copy paste it to the host and use these values from the host controller always, I would say we should use the one from the host.

guettli: But tests fail now. But this is a bug in the way the test data gets created....

guettli: tests are fixed now.

pkg/services/baremetal/host/host_suite_test.go
```diff
@@ -132,6 +132,9 @@ func newTestService(
 			Namespace: host.Namespace,
 		},
 	}
+	if host.Spec.Status.SSHSpec != nil {
```
janiskemper, line 135: I think you can remove this again. That was only necessary because of a code snippet we removed now

janiskemper: @guettli this is still open

guettli: @janiskemper done

janiskemper (approved):

PR #1968, opened 2026-04-15
:bug: requeue instead of silent return when rate-limited when 'HCloudMachine' already has 'Status.Ready=true'

### LINKED ISSUE #1966: HCloudMachine stuck with stale error conditions after rate-limit event when Status.Ready=true
/kind bug

During a sustained Hetzner API rate-limit event (e.g. caused by high-frequency HCCM `--route-reconciliation-period`, many concurrent CAPH reconciles, or CSI volume polling), the following sequence leads to permanently stuck `HCloudMachine` objects:

1. The rate limit is hit while `findServer` is called for a machine that already has `Status.Ready=true`.
2. CAPH enters the branch in `Reconcile()` (server.go) that handles this case:
   ```go
   // If HCloudMachine is ready, stop reconciling.
   ```
3. Because `Result{}` carries no `RequeueAfter`, the reconciler **never re-schedules** this machine — it only re-reconciles if an external event (annotation, spec change) triggers a new watch event.
4. Meanwhile, stale error conditions written during the rate-limit storm (e.g. `ServerCreateSucceeded=False` with reason `ServerCreateFailedReason`) persist on `status.conditions` indefinitely.
5. The parent `MachineDeployment` observes `availableReplicas < desiredReplicas` (because `maxUnavailable=0` treats any machine with a False condition as unavailable) and remains `Available=False` — even though every underlying Hetzner server is running healthy and the node is registered in Kubernetes.

The stuck state is permanent: the machines self-heal only if CAPH is restarted (which triggers a full re-list) or if conditions are manually patched via `kubectl patch --subresource=status`.

When `findServer` hits a rate limit on a machine that is already `Status.Ready=true`, CAPH should schedule a delayed requeue (e.g. `RequeueAfter: 5m`) so that once the rate limit lifts, a normal reconcile runs, clears any stale conditions, and the `MachineDeployment` recovers automatically without operator intervention.

The affected code path in `pkg/services/hcloud/server/server.go`:

```go
// Current (buggy) - origin/main and v1.1.x
if !s.scope.HCloudMachine.Status.Ready {
    hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
}
// If HCloudMachine is ready, stop reconciling.
return reconcile.Result{}, nil   // ← no requeue; stale conditions never cleared
```

Proposed fix:
```go
if !s.scope.HCloudMachine.Status.Ready {
    hcloudutil.HandleRateLimitExceeded(s.scope.HCloudMachine, err, "findServer")
}
// HCloudMachine is ready so we skip setting a misleading rate-limit condition.
// However, we still requeue so that any stale error conditions left over from
// a previous rate-limit storm can be cleared once the rate limit lifts.
return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```

This also interacts with issue #1934 / PR #1935: the proposed `HCloudRateLimitExceeded` v1beta2 negative-polarity condition will have the same silent-no-requeue problem on this code path, since `handleRateLimit` (which sets the condition) is bypassed for ready machines. The fix above ensures the reconciler comes back and `reconcileRateLimit` eventually deletes the v1beta2 condition after the wait time expires.

We observed this affecting 31 HCloudMachines across 8 clusters simultaneously after a ~24-hour rate-limit event.

**Environment:**

- cluster-api-provider-hetzner version: v1.0.12 (main), confirmed present in v1.1.0-alpha.2 (v1.1.x)
- Kubernetes version: v1.32.x (management cluster)
- OS: Linux (Hetzner Cloud, Ubuntu 24.04)

**AI Disclaimer**
Claude Sonnet 4.6 through Github Copilot did the troubleshooting and provided the fix.

pidreher:
**AI Disclaimer**
This PR was created by Claude Sonnet 4.6 through GitHub Copilot

When `findServer` returns a rate-limit error but the `HCloudMachine` already has `Status.Ready=true`, the previous code silently returned `reconcile.Result{}` with no `RequeueAfter`. This meant the reconciler never re-scheduled the machine after the rate limit lifted.

As a result, stale error conditions written during the rate-limit event (e.g. `ServerCreateSucceeded=False`, `HetznerAPIReachable=False`) persisted indefinitely on `status.conditions`. The parent `MachineDeployment` observed fewer `availableReplicas` than desired (with `maxUnavailable=0`) and stayed `Available=False` even though every underlying Hetzner server was healthy and its node was registered in Kubernetes.

The fix changes the silent return to `RequeueAfter: 5 * time.Minute`, so once the rate limit lifts the reconciler runs the happy path and clears any stale conditions automatically — no operator intervention required.

Fixes #1966

The same bug exists on `v1.1.x` and has been ported there in a separate branch (`fix/rate-limit-silent-return-v1.1.x`). That branch also interacts with the upcoming v1beta2 `HCloudRateLimitExceeded` condition from PR #1935 — without this requeue, a stuck `HCloudRateLimitExceeded=True` condition on a ready machine would similarly never be cleared by `reconcileRateLimit`.

---
**TODOs**:

- [x] squash commits
- [x] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -116,8 +116,14 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
 		}
 
-		// If HCloudMachine is ready, stop reconciling.
-		return reconcile.Result{}, nil
+		// HCloudMachine is ready so we skip setting a misleading rate-limit condition.
+		// However, we still requeue so that any stale error conditions (e.g. ServerCreateSucceeded=False
+		// left over from a previous rate-limit storm) can be cleared once the rate limit lifts.
+		s.scope.Info("rate limit exceeded during findServer, requeueing to clear stale conditions",
+			"namespace", s.scope.HCloudMachine.Namespace,
+			"name", s.scope.HCloudMachine.Name,
+		)
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
 	// if no server is found we have to create one.


janiskemper:
I still have doubts. I raised them in the issue: https://github.com/syself/cluster-api-provider-hetzner/issues/1966#issuecomment-4255109046

PR #1978, opened 2026-04-17
:seedling: fix(hcloudmachine): reconcile on control-plane lb target changes

guettli:
## Introduction

`reconcileLoadBalancerAttachment()` (in the hcloud server service) decides whether a
control-plane node should be added to, or removed from, the Hetzner Load Balancer. To make
that decision it looks at two things:

* Two signals on the related Cluster API `Machine`:
  * `Machine.DeletionTimestamp` — set once a machine starts being deleted, i.e. it's safe to
    remove it from the LB. (Deliberately not using the deprecated
    `PreDrainDeleteHookSucceededV1Beta1Condition` — v1beta1 conditions are going away.)
  * `controlplanev1.KubeadmControlPlaneMachineAPIServerPodHealthyCondition` — set once a new
    control-plane node's kube-apiserver pod is healthy, i.e. it's safe to add it to the LB.
* `HetznerCluster.Status.ControlPlaneLoadBalancer.Target`, which HCloudMachine's own
  reconcile loop uses to know which servers are already attached.

The problem: both of these only ever change as part of a `Machine.Status`/`ObjectMeta` or
`HetznerCluster.Status` update. Before this PR:

* `IgnoreInsignificantMachineStatusUpdates()` treated **all** `Machine.Status`-only changes as
  insignificant, so it filtered out exactly the Condition transition above.
* `IgnoreInsignificantHetznerClusterUpdates()` did the same for
  `Status.ControlPlaneLoadBalancer.Target` changes.

In both cases the HCloudMachine controller was never triggered by the event that actually
matters for LB attachment — it only picked up the change incidentally, on whatever its next
unrelated reconcile happened to be. In practice: when a new control-plane node's apiserver pod
becomes healthy, it should be added to the LB right away, but the controller could sit for up
to ~30s (the requeue interval in `reconcileLoadBalancerAttachment`, see #1988 — already merged,
it fixed a related bug where that requeue result was silently dropped) before noticing.

**Fix**: both predicates now explicitly check for changes to these signals and mark the event
as significant, so HCloudMachine reconciles immediately instead of waiting for an incidental
trigger. `Machine.DeletionTimestamp` changes are ObjectMeta changes and are already caught by
the predicate's generic old/new comparison, so only the apiserver-health Condition needs an
explicit check.

## Why bare metal doesn't need the same fix

`HetznerBareMetalMachine` has an equivalent `reconcileLoadBalancerAttachment()`
(in `pkg/services/baremetal/baremetal`) that reacts to the same two signals. However,
`HetznerBareMetalMachineReconciler.SetupWithManager` (`controllers/hetznerbaremetalmachine_controller.go`)
watches `clusterv1.Machine` and `HetznerCluster` without any `IgnoreInsignificant*`-style
predicate on those watches — the only global filter it applies is
`predicates.ResourceNotPausedAndHasFilterLabel`, which only inspects the pause annotation and
watch-filter label on the incoming object and never diffs old vs. new state. So it can't
suppress a Condition-only or Target-only change the way the two `IgnoreInsignificant*`
predicates fixed here do.

In other words, the bug this PR fixes is specific to the extra `IgnoreInsignificantMachineStatusUpdates`/
`IgnoreInsignificantHetznerClusterUpdates` filtering that only exists on the HCloudMachine
watches. Bare metal never filters these events out in the first place, so
`HetznerBareMetalMachine` already reconciles immediately on both signals. No change needed there.


controllers/hcloudmachine_controller.go
```diff
@@ -401,6 +403,42 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
 	}
 }
 
+// These Machine conditions directly control control-plane load balancer membership,
+// but they arrive as status-only updates. We must treat them as significant so the
+// HCloudMachine controller reconciles when a replacement control plane becomes safe
+// to add to the LB or when an old node becomes ready to be removed from it.
+func machineConditionsAffectingLoadBalancerReconcileChanged(oldMachine, newMachine *clusterv1.Machine) bool {
+	for _, conditionType := range []clusterv1.ConditionType{
+		clusterv1.PreDrainDeleteHookSucceededV1Beta1Condition,
```
janiskemper: I think we should check here the v1beta2 conditions. Otherwise I would even expect this to fail because the v1beta1 conditions are not in the same list AFAIK

janiskemper (dismissed):

controllers/hcloudmachine_controller.go
guettli: @janiskemper The first version of the PR followed the code:

```go
if deprecatedv1beta1conditions.Has(s.scope.Machine, clusterv1.PreDrainDeleteHookSucceededV1Beta1Condition)
```

The new way (Checking the DeletionTimestamp) is more robust.

Please check again.

janiskemper (commented):
looks good to me. @abdullah599 can you also have a look - especially also with regards to the conditions package we use?

controllers/hcloudmachine_controller.go
```diff
@@ -401,6 +403,42 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
 	}
 }
 
+// machineConditionsAffectingLoadBalancerReconcileChanged reports whether the Machine's
+// KubeadmControlPlaneMachineAPIServerPodHealthyCondition changed. reconcileLoadBalancerAttachment
+// (pkg/services/hcloud/server/server.go) uses this condition to decide whether to attach the
+// server to the load balancer. Without this check, a condition change alone would not trigger
+// a reconcile.
+func machineConditionsAffectingLoadBalancerReconcileChanged(oldMachine, newMachine *clusterv1.Machine) bool {
+	return !reflect.DeepEqual(
```
abdullah599: This compares the whole condition, so a change in Reason or Message alone still triggers a reconcile plus a live GetServer call, even though reconcileLoadBalancerAttachment only reads the status via conditions.IsTrue. Can we compare just the status: conditions.IsTrue(oldMachine, ...) != conditions.IsTrue(newMachine, ...) or is there something I missed?

controllers/hcloudmachine_controller.go
```diff
@@ -401,6 +403,42 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
 	}
 }
 
+// machineConditionsAffectingLoadBalancerReconcileChanged reports whether the Machine's
+// KubeadmControlPlaneMachineAPIServerPodHealthyCondition changed. reconcileLoadBalancerAttachment
+// (pkg/services/hcloud/server/server.go) uses this condition to decide whether to attach the
+// server to the load balancer. Without this check, a condition change alone would not trigger
+// a reconcile.
+func machineConditionsAffectingLoadBalancerReconcileChanged(oldMachine, newMachine *clusterv1.Machine) bool {
```
abdullah599: This checks a single condition, so the plural Conditions in the name reads a bit off. Could we rename to machineAPIServerPodHealthyConditionChanged

controllers/hcloudmachine_controller.go
```diff
@@ -401,6 +403,42 @@ func (r *HCloudMachineReconciler) HetznerClusterToHCloudMachines(_ context.Conte
...
+
+	if oldCluster.Status.ControlPlaneLoadBalancer != nil {
+		oldTargets = oldCluster.Status.ControlPlaneLoadBalancer.Target
+	}
+
+	if newCluster.Status.ControlPlaneLoadBalancer != nil {
+		newTargets = newCluster.Status.ControlPlaneLoadBalancer.Target
+	}
+
+	// ignore empty vs nil slice.
```
abdullah599: nit: slices.Equal already treats nil and empty slices as equal, so we could drop this guard and the DeepEqual below and just return !slices.Equal(oldTargets, newTargets). LoadBalancerTarget is comparable, and it removes the need for this comment too


<details>

<summary>Ref</summary>

Image

</details>

pkg/services/hcloud/server/server.go
```diff
@@ -1531,11 +1530,11 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, server *h
 		return reconcile.Result{}, nil
 	}
 
-	// remove server from load balancer if it's being deleted.
-	// s.scope.Machine is a CAPI core v1beta2 Machine, so its legacy conditions
-	// live at Status.Deprecated.V1Beta1.Conditions — use deprecatedv1beta1conditions,
-	// not v1beta1conditions. See .golangci.yaml for the full mapping.
-	if deprecatedv1beta1conditions.Has(s.scope.Machine, clusterv1.PreDrainDeleteHookSucceededV1Beta1Condition) {
+	// remove server from load balancer if it's being deleted. The Machine controller only
```
abdullah599: I do not think equivalent is quite right. DeletionTimestamp is set the moment deletion starts, which is earlier and broader than the old condition, so the server leaves the load balancer sooner. Also pre-drain hook is not a term used anywhere in CAPH, so it is hard to follow. Can we just describe the current behavior: remove the server as soon as the Machine starts deleting

controllers/hcloudmachine_controller_test.go
```diff
@@ -232,6 +285,50 @@ func TestIgnoreInsignificantMachineStatusUpdates(t *testing.T) {
 	}
 }
 
+func TestIgnoreInsignificantHetznerClusterUpdates_TargetChanges(t *testing.T) {
```
abdullah599, line 288: This only covers the case where a target is added. Could we also add an unchanged targets case that expects false and a nil ControlPlaneLoadBalancer case, so a future change that always returns true would be caught

abdullah599 (commented):
minor nits else LGTM :)

controllers/hcloudmachine_controller.go
guettli: You're right, `reconcileLoadBalancerAttachment` only ever
calls `conditions.IsTrue` on this condition, so Reason/Message churn shouldn't cause a reconcile.
Switched to exactly the comparison you suggested:

```go
func machineAPIServerPodHealthyConditionChanged(oldMachine, newMachine *clusterv1.Machine) bool {
	return conditions.IsTrue(oldMachine, controlplanev1.KubeadmControlPlaneMachineAPIServerPodHealthyCondition) !=
		conditions.IsTrue(newMachine, controlplanev1.KubeadmControlPlaneMachineAPIServerPodHealthyCondition)
}
```

controllers/hcloudmachine_controller.go
guettli: Agreed, renamed to `machineAPIServerPodHealthyConditionChanged` (and updated the call site and
comment accordingly). Turns out this rename also fits naturally with the fix for your comment
above.

controllers/hcloudmachine_controller.go
guettli: Nice simplification, done:

```go
func controlPlaneLoadBalancerTargetsChanged(oldCluster, newCluster *infrav1.HetznerCluster) bool {
	var oldTargets []infrav1.LoadBalancerTarget
	var newTargets []infrav1.LoadBalancerTarget

	if oldCluster.Status.ControlPlaneLoadBalancer != nil {
		oldTargets = oldCluster.Status.ControlPlaneLoadBalancer.Target
	}

	if newCluster.Status.ControlPlaneLoadBalancer != nil {
		newTargets = newCluster.Status.ControlPlaneLoadBalancer.Target
	}

	return !slices.Equal(oldTargets, newTargets)
}
```

Removed the "ignore empty vs nil slice" comment and the guard since `slices.Equal` already
handles that, and dropped `reflect.DeepEqual` in favor of `slices.Equal` (kept the `reflect`
import since it's still used elsewhere in the file).

pkg/services/hcloud/server/server.go
guettli: Good point, "equivalent" was misleading and the pre-drain-hook framing doesn't match how the
rest of CAPH talks about this. Simplified to just describe the behavior:

```go
// remove the server as soon as the Machine starts deleting.
if !s.scope.Machine.DeletionTimestamp.IsZero() {
```

(`pkg/services/hcloud/server/server.go`)

controllers/hcloudmachine_controller_test.go
guettli: Good catch, the existing test would have passed even if the function always returned `true`.
Added two subtests to `TestIgnoreInsignificantHetznerClusterUpdates_TargetChanges`:

* `unchanged targets` — same targets on both sides (only `ResourceVersion` differs), expects
  `false`.
* `nil ControlPlaneLoadBalancer` — `ControlPlaneLoadBalancer` is `nil` on both old and new
  object, expects `false`.

Both fail if `controlPlaneLoadBalancerTargetsChanged` is changed to unconditionally return
`true`. Ran `go test ./controllers/... -run TestIgnoreInsignificant -v` locally — all pass.

abdullah599 (approved):
LGTM! :)

PR #1981, opened 2026-04-18
:seedling: Set Host.Spec.Status.NodeBootID on every reboot

### LINKED ISSUE #1974: Improve BootID tracking and fix LastUpdated timeout reliability
## Background

During review of #1944, few design issues were identified with how reboot state is tracked on `HetznerBareMetalHost`.

---

## Problem 1: `RebootAnnotationNodeBootID` only set during annotation-triggered reboots

`RebootAnnotationNodeBootID` currently lives under `Spec.Status.ExternalIDs`. Currently this field only exists in the context of the reboot annotation, and `ExternalIDs` implies it references an external system. However we propose to store the BootID read from the workload cluster node on every reboot.
The BootID reflects the node's last boot, and should always be populated, first when the node successfully comes up during provisioning (`actionEnsureProvisioned`), and updated on every subsequent reboot.

### Proposed changes

- Rename `RebootAnnotationNodeBootID` to `NodeBootID`
- Move `NodeBootID` to top-level `ControllerGeneratedStatus` (remove the `ExternalIDs` struct entirely)
- Populate `NodeBootID` when the node first comes up during provisioning
- Update `NodeBootID` on every confirmed reboot (when BootID changes)

## Problem 2: `LastUpdated` is used for reboot timeouts but resets unreliably

`LastUpdated` is set in the host controller's defer block whenever the host object changes (`!cmp.Equal(initialHost, bmHost)`). It is used as a timeout reference for hetznerbaremetalhost's provisioning:

The problem is that `LastUpdated` is not a stable anchor. It changes when conditiions are updated, which results - timeout window being reset/sliding forward.

## Problem 3: `RebootAnnotationSince` does not clearly communicate its purpose, and `LastUpdated` should be deprecated

`RebootAnnotationSince` (currently under `ExternalIDs`) is a dedicated timestamp that records when a reboot annotation was first noticed. It is written exactly once per reboot cycle and is not affected by condition changes, making it a reliable anchor for the 5-minute reboot timeout in `actionProvisioned`.

However, its name is tied to the annotation concept and it lives in the wrong place.

`LastUpdated` is a general timestamp and should not be used for reboot timeout logic, it resets in ways that break timeout correctness.

### Proposed changes

- Introduce a new field `RebootTriggeredAt *metav1.Time` at top-level `ControllerGeneratedStatus`
  - Set when a reboot is triggered (Phase 1 of the annotation reboot flow, and during provisioning reboots)
  - Cleared when the reboot is confirmed (BootID changes)
  - Use this field for all reboot-related timeout checks instead of `LastUpdated`
- Deprecate `LastUpdated`. It can be removed in v1beta2.

---

## Summary of field changes

| Old | New | Notes |
|-----|-----|-------|
| `Spec.Status.ExternalIDs.RebootAnnotationNodeBootID` | `Spec.Status.NodeBootID` | Always set, not just during annotation reboots |
| `Spec.Status.ExternalIDs.RebootAnnotationSince` | `Spec.Status.RebootTriggeredAt` | Replaces hacky `LastUpdated` usage for reboot timeouts |
| `Spec.Status.ExternalIDs` | removed | Struct serves no purpose once fields are moved |
| `Spec.Status.LastUpdated` | deprecated | Too broad for timeout logic; remove in v1beta2 |

Dhairya-Arora01:

Key changes made:
- Spec.Status.ExternalIDs.RebootAnnotationNodeBootID has been renamed to Spec.Status.NodeBootID.
  - Populate NodeBootID when the node first comes up during provisioning.
  - Update NodeBootID on every confirmed reboot.
- Spec.Status.ExternalIDs.RebootAnnotationSince has been renamed to Spec.Status.RebootTriggeredAt.
  - Set when the reboot is triggered and used to check if the reboot timed-out.
- Deprecate Spec.Status.LastUpdatedAt and instead use RebootTriggeredAt to detect reboot timeouts.

Fixes #1974

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


pkg/services/baremetal/host/host.go
```diff
@@ -2058,16 +2045,50 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 		return actionStop{}
 	}
 
-	if host.Spec.Status.ExternalIDs.RebootAnnotationNodeBootID == "" {
+	if !rebootDesired {
```
janiskemper, line 2073: we need to find a new way of working with the conditions. Before this line you set RebootSucceeded condition on false - independent on whether there is a reboot or not

pkg/services/baremetal/host/host.go
```diff
@@ -229,6 +230,7 @@ func (s *Service) actionPreparing(ctx context.Context) actionResult {
 			}
 			msg := "Rebooting into rescue mode."
 			createSSHRebootEvent(ctx, s.scope.HetznerBareMetalHost, msg)
+			s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
```
janiskemper, line 233: I have the feeling that you don't always set this. E.g. I searched for ".Reboot()" and found some occurrences without the timestamp, if I'm not mistaken.

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
Dhairya-Arora01: I missed it here https://github.com/syself/cluster-api-provider-hetzner/blob/e28ea22f8efcf491c6fdc70d2214ad040fb6e2bb/pkg/services/baremetal/host/host.go#L1913

pkg/services/baremetal/host/host.go
```diff
@@ -1961,51 +1984,14 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
 	if err != nil {
-		err = fmt.Errorf("actionProvisioned (Reboot via Annotation), failed to get wlClient: %w", err)
-		conditions.MarkFalse(host, infrav1.RebootSucceededCondition,
-			"GetWorkloadClusterClientFailed",
-			clusterv1.ConditionSeverityWarning, "%s",
+		err = fmt.Errorf("actionProvisioned, failed to get wlClient: %w", err)
+		conditions.MarkFalse(host,
+			infrav1.NodeBootIDRetrievedCondition,
+			infrav1.GetWorkloadClusterClientFailedReason,
+			clusterv1.ConditionSeverityError, "%s",
```
janiskemper: let's do a warning here. this can be temporary.

pkg/services/baremetal/host/host.go
```diff
@@ -2047,26 +2041,64 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
...
-			"NodeInWorkloadClusterHasEmptyBootID",
-			clusterv1.ConditionSeverityWarning, "%s",
+		conditions.MarkFalse(host,
+			infrav1.NodeBootIDRetrievedCondition,
+			infrav1.BootIDEmptyReason,
+			clusterv1.ConditionSeverityWarning,
+			"%s",
 			msg)
 
 		s.scope.HetznerBareMetalHost.SetError(infrav1.FatalError, msg)
```
janiskemper: we shouldn't set this fatal error anymore.. Since we now check directly after provisioning, it is for sure more likely to happen. I would even say that we need to see what happens and maybe say we requeue or something - or just leave it empty. 

Currently, I would say let's leave it empty. However, only in cases where we are not supposed to reboot. If so, we have to actually give this as an error - even fatal

janiskemper (commented):
can you write down which searches you did to find all reboots? Just to review that part.

pkg/services/baremetal/host/host.go
Dhairya-Arora01: Should we do the same, when machine.status.nodeRef is nil ?
https://github.com/syself/cluster-api-provider-hetzner/blob/bff384634835369e7e99a08d609f29b8bcaf5606/pkg/services/baremetal/host/host.go#L2017-L2022

janiskemper: I probably missed that, so yes.

Dhairya-Arora01:
> can you write down which searches you did to find all reboots? Just to review that part.

1. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L228
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L233
2. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1241
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1247
3. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1557
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1563
4. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1865
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1871
5. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1913
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L1918
6. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L2123
  Timestamp set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L2173

Dhairya-Arora01:
For reboots via Robot API

1. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L244
  Timestamp Set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L249
  
2. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L431
  Timestamp Set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L435
  
3. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L479
  Timestamp Set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L483
  
4. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L513
  Timestamp Set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L517
  
5. https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L2136
  Timestamp Set: https://github.com/syself/cluster-api-provider-hetzner/blob/d844d40672353b6dee803764b4c7a97e1f79454e/pkg/services/baremetal/host/host.go#L2173

pkg/services/baremetal/host/host.go
Dhairya-Arora01: okay

pkg/services/baremetal/host/host.go
```diff
@@ -1112,6 +1119,9 @@ func (s *Service) actionPreProvisioning(ctx context.Context) actionResult {
 		return actionStop{}
 	}
 
+	// we are in resuce mode i.e. reboot was successful, now clear the RebootTriggeredAt timestamp.
```
janiskemper: I think this should be in actionRegistering not here...

pkg/services/baremetal/host/host.go
```diff
@@ -1817,6 +1835,7 @@ func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult)
 	record.Event(s.scope.HetznerBareMetalHost, "ServerProvisioned", "server successfully provisioned")
 	conditions.MarkTrue(s.scope.HetznerBareMetalHost, infrav1.ProvisionSucceededCondition)
 	s.scope.HetznerBareMetalHost.ClearError()
+	s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = nil
```
janiskemper: I think you should do that right after you realized that the reboot worked.. Not after the validation that the provisioning worked

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
Dhairya-Arora01: makes sense

pkg/services/baremetal/host/host.go
Dhairya-Arora01: done

janiskemper (approved):

PR #1982, opened 2026-04-19
fix: nil pointer panic in createOptsFromSpec when ControlPlaneEndpoint is unset

The127:
## Summary

- `createOptsFromSpec` dereferences `hc.Spec.ControlPlaneEndpoint.Port` unconditionally, but `ControlPlaneEndpoint` is a pointer (`*clusterv1.APIEndpoint`) that is `nil` during the first reconciliation — before CAPH has created the LB and populated the field.
- This causes a controller crash loop (`nil pointer dereference` at `loadbalancer.go:326`) on every fresh `HetznerCluster` with `controlPlaneLoadBalancer.enabled: true`.
- The official cluster template uses `controlPlaneEndpoint: {host: "", port: 443}` as a workaround, but Kubernetes strips the field due to `omitempty` serialization on the pointer type, so the nil dereference is unavoidable following the documented bootstrap flow.

## Fix

Fall back to `ControlPlaneLoadBalancer.Port` when `ControlPlaneEndpoint` is nil or has no port set. This is the correct value anyway since the LB listen port is configured via `controlPlaneLoadBalancer.port`.

## How to reproduce

1. Apply a `HetznerCluster` with `controlPlaneLoadBalancer.enabled: true` and no `controlPlaneEndpoint` (or `host: ""` which gets stripped by Kubernetes).
2. CAPH controller enters crash loop immediately on first reconciliation.

## Test plan

- [ ] Fresh `HetznerCluster` with LB enabled and no `controlPlaneEndpoint` — controller no longer crashes
- [ ] `controlPlaneEndpoint.port` is respected when explicitly set (non-nil, non-zero)
- [ ] `controlPlaneLoadBalancer.port` is used as fallback when endpoint is nil

### DIFF
--- a/pkg/services/hcloud/loadbalancer/loadbalancer.go
+++ b/pkg/services/hcloud/loadbalancer/loadbalancer.go
@@ -323,7 +323,12 @@ func createOptsFromSpec(hc *infrav1.HetznerCluster) hcloud.LoadBalancerCreateOpt
 		network = &hcloud.Network{ID: hc.Status.Network.ID}
 	}
 
-	listenPort := int(hc.Spec.ControlPlaneEndpoint.Port)
+	// ControlPlaneEndpoint is set by CAPH after LB creation, so it may be nil on
+	// the first reconciliation. Fall back to ControlPlaneLoadBalancer.Port in that case.
+	listenPort := int(hc.Spec.ControlPlaneLoadBalancer.Port)
+	if hc.Spec.ControlPlaneEndpoint != nil && hc.Spec.ControlPlaneEndpoint.Port != 0 {
+		listenPort = int(hc.Spec.ControlPlaneEndpoint.Port)
+	}
 	publicInterface := true
 	return hcloud.LoadBalancerCreateOpts{
 		LoadBalancerType: &hcloud.LoadBalancerType{Name: hc.Spec.ControlPlaneLoadBalancer.Type},


The127:
Workaround i used: let the controller crash once (the LB still gets created), then grab the LB IP with hcloud load-balancer list and patch it in  manually:                                                                                                                                   
```bash
kubectl patch hetznercluster <name> --type=merge -p '{"spec":{"controlPlaneEndpoint":{"host": <LB_IP>","port":6443}}}'
kubectl patch cluster <name> --type=merge -p '{"spec":{"controlPlaneEndpoint":{"host":"<LB_IP>","port":6443}}}'
kubectl delete secret <cluster-name>-kubeconfig
```
The controller stabilizes after that and provisioning continues normally.

guettli:
@The127 have you looked at the v1.1.x branch? Does it have the same issue?

The127:
@guettli will do. and report back :) 
TY for repsonding so quickly and pointing me towards a potentially fixed branch. 
I have only used v1.0.12 which is the latest release.

guettli:
We work heavily on releasing v1.1.0 soon

The127:
I couldnt easily try it out again with the unreleased version, but i can tell that the code looks to be doing the correct thing in v1.1.0. 
I will close this PR.
Ty again for the quick responses.

PR #1987, opened 2026-04-20
:seedling: Gate bare-metal LB attachment on API server health

### LINKED ISSUE #1986: Reconcile of baremetal control-plane LoadBalancer
/kind feature

For hcloud CP LBs, we check if the Pod on the related capi machine is ready. But for bm we do not do that.

Here is the hcloud code, which we miss for bm CP.

```go 
	// attach only if server has private IP or public IPv4, otherwise Hetzner cannot handle it
	if server.PublicNet.IPv4.IP == nil && !hasPrivateIP {
	}

	apiServerPodHealthy := s.scope.Cluster.Spec.ControlPlaneRef == nil ||
		s.scope.Cluster.Spec.ControlPlaneRef.Kind != "KubeadmControlPlane" ||
		conditions.IsTrue(s.scope.Machine, controlplanev1.MachineAPIServerPodHealthyCondition)

	// we attach only nodes with kube-apiserver pod healthy to avoid downtime, skipped for the first node
	if len(s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.Target) > 0 && !apiServerPodHealthy {
		conditions.MarkFalse(hm, infrav1.ServerAvailableCondition,
			"WaitingForAPIServer", clusterv1.ConditionSeverityInfo,
			"reconcile LoadBalancer: apiserver pod not healthy yet.")
	}
```

guettli:
Gate bare-metal control-plane load balancer target attachment on `APIServerPodHealthy` once the load balancer already has a target

Keep the first control-plane target unblocked and cover both behaviors with unit tests

Bare-metal control-plane load balancer reconciliation attached new targets as soon as the host IPs were known. Unlike the hcloud implementation, it did not wait for the kube-apiserver pod on the corresponding CAPI machine to become healthy before adding subsequent control-plane nodes, which risks API downtime during scale-up or replacement.

Fixes #1986


Dhairya-Arora01 (approved):
LGTM

janiskemper (changes requested):
we have to evaluate this first based on the similar PR for hcloud. Probably we need to change the order of things. We at least need to have a similar comment as in the hcloud PR why this order is necessary.

cc @abdullah599

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -571,9 +578,28 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
 
 	// if both IPs are already added as target, then do nothing
 	if foundIPv4 && foundIPv6 {
+		conditions.MarkTrue(s.scope.BareMetalMachine, infrav1.ServerAvailableCondition)
```
janiskemper: from what I can see in the PR, this condition seems to not be set until now on this object. Currently, it would also only be set for control planes, which doesn't make a lot of sense.

I am not a huge fan of the amount of places where we set it on "true". Is it maybe possible to do it more on specific points?

janiskemper (approved):

PR #1988, opened 2026-04-21
:seedling: Return early from handleOperatingSystemRunning on non-zero reconcile.Result

abdullah599:

`handleOperatingSystemRunning` calls `reconcileLoadBalancerAttachment` but then ignores its returned `reconcile.Result`. When that result carries a `RequeueAfter` (e.g. the kube-apiserver pod is not yet healthy and the server should not be attached to the load balancer yet), `handleOperatingSystemRunning` would still fall through to:

~~~go
s.scope.SetReady(true)
conditions.MarkTrue(hm, infrav1.ServerAvailableCondition)
~~~

overwriting the `ServerAvailableCondition=False/WaitingForAPIServer` that `reconcileLoadBalancerAttachment` had just set, and marking the server Ready prematurely.

This PR adds the missing early return so the requeue propagates and the condition is preserved:

~~~go
if res != (reconcile.Result{}) {
    return res, nil
}
~~~

This fix was originally bundled in #1978; it is split out here so it can land on its own.

None

none.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -822,11 +822,12 @@ func (s *Service) handleOperatingSystemRunning(ctx context.Context, server *hclo
 		return res, reterr
 	}
 
+	s.scope.SetReady(true)
+
 	if res != (reconcile.Result{}) {
 		return res, nil
 	}
 
-	s.scope.SetReady(true)
 	conditions.MarkTrue(hm, infrav1.ServerAvailableCondition)
 	return reconcile.Result{}, nil
 }
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -823,9 +823,14 @@ func (s *Service) handleOperatingSystemRunning(ctx context.Context, server *hclo
 	}
 
 	// Order matters:
-	// 1. SetReady(true) first so CAPI propagates ProviderID to the Machine.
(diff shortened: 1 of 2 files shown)


guettli (approved):
lgtm, thank you!

janiskemper (changes requested):
can you add a comment in the code about the order of things here? That's pretty complicated actually and also hard to follow

guettli (approved):
lgtm

pkg/services/hcloud/server/server.go
```diff
@@ -822,7 +822,17 @@ func (s *Service) handleOperatingSystemRunning(ctx context.Context, server *hclo
 		return res, reterr
 	}
 
+	// Order matters:
+	// 1. SetReady(true) first so CAPI propagates ProviderID to the Machine.
```
janiskemper, line 826: plz add that this is necessary for the Machine to become ready, and for the node to be created. Also mention that otherwise we would have a deadlock, because we wait for the node to become ready to add the control planes as load balancer targets, but the node never becomes ready because we don't set it here.

So plz bring in the context from reconcileLoadBalancerAttachment

janiskemper (changes requested):

janiskemper (approved):

PR #1990, opened 2026-04-21
⚠️ Bump CAPI to v1.11 (keep v1beta1 contract)

abdullah599:

This PR bumps CAPH from CAPI v1.10 to v1.11, along with the related Kubernetes and controller runtime updates. CAPH continues to implement the v1beta1 CAPI contract. The move to the v1beta2 contract for our own CRDs is planned as a follow up, before the v1beta1 compatibility layer is removed upstream (tentatively August 2026).

---

## Background

CAPI v1.11 introduces the new v1beta2 API. Providers can choose to stay on the v1beta1 contract for now and move to v1beta2 later. This PR takes that path, following CAPA PR #5720 as the primary reference. 

Migration guide: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11

## Scope of this PR

In scope:
* Bump CAPI, controller runtime, and Kubernetes dependencies.
* Update Go imports for CAPI core types from v1beta1 to v1beta2.
* Adapt to contract changes that are required at compile time.
* Switch condition writes on CAPI core Machine to the Metal3 pattern.
* Keep our CRDs on the v1beta1 contract shape.

Out of scope (Phase B, follow up PR):
* Migrate our CRDs to `[]metav1.Condition`.
* Add a v1beta2 API version for our CRDs.
* Declare `contract: v1beta2` in metadata.yaml.

### 1. CRD contract stays on v1beta1

Our CRDs keep the same shape. `metadata.yaml` still declares `contract: v1beta1`. No field types change on our side. This matches what CAPA did.

### 2. Go imports for CAPI core move to v1beta2

This is separate from the contract choice. The migration guide recommends switching Go imports to v1beta2 when bumping to CAPI v1.11. All CAPI helper functions in v1.11 (like `util.IsControlPlaneMachine`, `predicates.*`, `kubeconfig.*`, the test framework) now accept only v1beta2 Go types. Keeping v1beta1 Go imports would require forking every helper we use.

Quote from the guide:
> in v1.11 all the CAPI utils like e.g. IsControlPlaneMachine are now using v1beta2 API types. This might lead to additional work for providers to keep using v1beta1 types from CAPI v1.11 (you have to fork all utils the provider is using from an older CAPI release or replace them with something else).

### 3. Conditions on our own CRDs

Our CRDs still have `Status.Conditions` as the legacy `clusterv1beta1.Conditions` slice. So we use the deprecated helper packages for them:
* `sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions` for MarkFalse, MarkTrue, Set, Has, Get, IsTrue, SetSummary.
* `sigs.k8s.io/cluster-api/util/deprecated/v1beta1/patch` as the patch helper.

Both are imported under prefixed aliases (`v1beta1conditions`, `v1beta1patch`) so they cannot be confused with the non deprecated packages. The migration guide warns that using the wrong patch helper can silently drop conditions at runtime with no compile error. The aliases plus `depguard` rules in `.golangci.yaml` prevent that.

### 4. Conditions on CAPI core Machine (remediation): Metal3 pattern

When our remediation controllers need to write the `OwnerRemediated` condition on a CAPI core Machine, we follow the Metal3 pattern. We write to the legacy shadow field `status.deprecated.v1beta1.conditions`, using `sigs.k8s.io/cluster-api/util/conditions/deprecated/v1beta1`. This keeps us fully on v1beta1 style condition types and works with the CAPI 1.11 compatibility layer, which MHC uses to observe those conditions.

Reference: https://github.com/metal3-io/cluster-api-provider-metal3/blob/v1.12.3/baremetal/metal3remediation_manager.go#L349 (`SetOwnerRemediatedConditionNew`).

### 5. `InfrastructureRef` becomes `ContractVersionedObjectReference`

Every site that referenced `Cluster.Spec.InfrastructureRef`, `Machine.Spec.InfrastructureRef`, or similar refs on CAPI core objects had to be updated.

**What changed upstream:**

In CAPI v1.11, these refs stopped being `*corev1.ObjectReference` and became a new purpose built type `ContractVersionedObjectReference`. The new type has a smaller, stricter surface:

~~~go
type ContractVersionedObjectReference struct {
    APIGroup string
    Kind     string
    Name     string
}
~~~

No `Namespace`, no `UID`, no `APIVersion`, no `ResourceVersion`, no `FieldPath`.

**How we adapted:**

* **Presence checks.** The old `ref == nil` or `ref != nil` pattern does not apply. The new type is a value, not a pointer. To check if a ref is set, we use the new `ref.IsDefined()` method.
* **Kind access.** The old `ref.GroupVersionKind().Kind` is now simply `ref.Kind`. The old `ref.GroupVersionKind().GroupKind()` becomes `schema.GroupKind{Group: ref.APIGroup, Kind: ref.Kind}`.
* **APIGroup instead of APIVersion.** The new type stores only the API group (no version). This is intentional: CAPI resolves the version at runtime from the contract, so providers cannot hardcode a version.
* **Namespace is dropped.** The ref is always resolved in the same namespace as the owner object. Every place that previously set `InfrastructureRef.Namespace` had that field removed. This also applies to test fixtures that built these refs by hand.
* **Construction in tests.** Test fixtures that previously used `InfrastructureRef: &corev1.ObjectReference{...}` are now `InfrastructureRef: clusterv1.ContractVersionedObjectReference{APIGroup, Kind, Name}`. 

### 6. Alias scheme (mostly taken from CAPA)

To keep the code readable and safe, we use a consistent alias scheme everywhere:

| Alias | Package | Purpose |
|---|---|---|
| `clusterv1` | `sigs.k8s.io/cluster-api/api/core/v1beta2` | CAPI core types |
| `clusterv1beta1` | `sigs.k8s.io/cluster-api/api/core/v1beta1` | Legacy types we still reference |
| `bootstrapv1` | `sigs.k8s.io/cluster-api/api/bootstrap/kubeadm/v1beta2` | kubeadm bootstrap |
| `controlplanev1` | `sigs.k8s.io/cluster-api/api/controlplane/kubeadm/v1beta2` | kubeadm control plane |
| `v1beta1patch` | `util/deprecated/v1beta1/patch` | Patch helper for our CRDs |
| `v1beta1conditions` | `util/deprecated/v1beta1/conditions` | Condition utils for our CRDs |
| `deprecatedv1beta1conditions` | `util/conditions/deprecated/v1beta1` | Legacy shadow conditions on CAPI core v1beta2 objects (Metal3 pattern) |
| `conditions` | `util/conditions` | Read only, in `pkg/services/hcloud/server/server.go`, for CAPI owned conditions |

## References

* CAPA PR (primary reference, same contract choice): https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5720
* CAPI v1.10 to v1.11 migration guide: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.10-to-v1.11
* Metal3 remediation manager: https://github.com/metal3-io/cluster-api-provider-metal3/blob/v1.12.3/baremetal/metal3remediation_manager.go#L349

none

**TODOs**:
- [ ] remove vendor directory changes before requesting review
- [ ] squash commits
- [ ] add unit tests where missing


controllers/hcloudmachine_controller_test.go
```diff
@@ -324,8 +324,14 @@ var _ = Describe("HCloudMachineReconciler", func() {
 	Context("Basic hcloudmachine test", func() {
 		Context("correct server", func() {
 			BeforeEach(func() {
-				// remove bootstrap infos
-				capiMachine.Spec.Bootstrap = clusterv1.Bootstrap{}
+				// clear DataSecretName to simulate "bootstrap not ready"
+				capiMachine.Spec.Bootstrap = clusterv1.Bootstrap{
```
janiskemper, line 328: this doesn't look like clearing to me

pkg/scope/machine.go
```diff
@@ -177,7 +177,7 @@ func (m *MachineScope) SetProviderID(serverID int64) {
 
 // ServerIDFromProviderID converts the ProviderID (hcloud://NNNN) to the ServerID.
 func (m *MachineScope) ServerIDFromProviderID() (int64, error) {
-	if m.HCloudMachine.Spec.ProviderID == nil || m.HCloudMachine.Spec.ProviderID != nil && *m.HCloudMachine.Spec.ProviderID == "" {
```
janiskemper, line 180: I don't think this is correct

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -224,10 +225,14 @@ func (s *Service) setOwnerRemediatedConditionToFailed(ctx context.Context, msg s
 	// When machine is still unhealthy after remediation, setting of OwnerRemediatedCondition
 	// moves control to CAPI machine controller. The owning controller will do
 	// preflight checks and handles the Machine deletion.
-	conditions.MarkFalse(
+	//
+	// Write the legacy-shape condition into status.deprecated.v1beta1.conditions
+	// so the v1beta1 compat layer in MachineHealthCheck sees it. Mirrors the
+	// cluster-api-provider-metal3 approach in metal3remediation_manager.go.
```
janiskemper: let's not write this reference to the code. In the PR description that's enough

pkg/services/hcloud/server/server.go
```diff
@@ -143,7 +144,7 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 				*s.scope.HCloudMachine.Spec.ProviderID)
 
 			s.scope.Error(errors.New(msg), msg,
-				"ProviderID", *s.scope.HCloudMachine.Spec.ProviderID,
+				"ProviderID", s.scope.HCloudMachine.Spec.ProviderID,
```
janiskemper: I think this is wrong

pkg/services/hcloud/server/server.go
```diff
@@ -960,7 +960,7 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, server *h
 	}
 
 	// remove server from load balancer if it's being deleted
-	if conditions.Has(s.scope.Machine, clusterv1.PreDrainDeleteHookSucceededCondition) {
+	if conditions.Has(s.scope.Machine, string(clusterv1beta1.PreDrainDeleteHookSucceededCondition)) {
```
janiskemper: is this correct? To use the latest conditions package together with a v1beta1 condition?

pkg/services/hcloud/server/server_test.go
```diff
@@ -821,13 +822,20 @@ var _ = Describe("Reconcile", func() {
...
-				InfrastructureRef: corev1.ObjectReference{
-					Name:       hcloudMachine.Name,
-					Kind:       hcloudMachine.Kind,
-					APIVersion: hcloudMachine.APIVersion,
+				InfrastructureRef: clusterv1.ContractVersionedObjectReference{
+					APIGroup: infrav1.GroupVersion.Group,
+					Name:     hcloudMachine.Name,
+					Kind:     hcloudMachine.Kind,
+				},
+				Bootstrap: clusterv1.Bootstrap{
```
janiskemper, line 830: adding this seems to change the test setup. is this on purpose?

test/e2e/config/hetzner.yaml
```diff
@@ -149,7 +149,7 @@ variables:
 
   # NOTE: INIT_WITH_BINARY and INIT_WITH_KUBERNETES_VERSION are only used by the clusterctl upgrade test to initialize
   # the management cluster to be upgraded.
-  INIT_WITH_BINARY: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.10.10/clusterctl-linux-amd64"
+  INIT_WITH_BINARY: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.11.8/clusterctl-linux-amd64"
   INIT_WITH_PROVIDERS_CONTRACT: "v1beta1"
   INIT_WITH_KUBERNETES_VERSION: "v1.33.6"
```
janiskemper: I suppose we can update this as well. Can you check the past bumps of the CAPI version to see which versions to update?

.golangci.yaml
```diff
@@ -97,10 +97,30 @@ linters:
...
+          alias: controlplanev1
+        # Deprecated v1beta1 patch + conditions helpers. Prefixed aliases prevent
+        # silent swaps with the non-deprecated util/patch and util/conditions packages
+        # (see CAPI 1.10->1.11 migration guide; wrong helper can drop conditions at
+        # runtime with no compile error).
+        - pkg: sigs.k8s.io/cluster-api/util/deprecated/v1beta1/patch
+          alias: v1beta1patch
+        - pkg: sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions
+          alias: v1beta1conditions
+        # Legacy-shadow conditions helper for v1beta2 CAPI core objects (Metal3
```
janiskemper: let's remove metal3

metadata.yaml
```diff
@@ -6,6 +6,9 @@
 apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
 kind: Metadata
 releaseSeries:
+  - major: 1
```
janiskemper: not sure, what is this about?

controllers/controllers_suite_test.go
```diff
@@ -437,30 +439,46 @@ func getDefaultHetznerBareMetalMachineSpec() infrav1.HetznerBareMetalMachineSpec
...
 	return objectCondition.Status == corev1.ConditionFalse &&
 		objectCondition.Reason == reason
 }
 
-func isPresentAndTrue(key types.NamespacedName, getter conditions.Getter, condition clusterv1.ConditionType) bool {
+// isPresentAndFalseWithReasonV2 reads a legacy-shape condition from a v1beta2
+// CAPI core object (Cluster, Machine) via GetV1Beta1Conditions(), i.e. the
+// status.deprecated.v1beta1.conditions field. This is how CAPI 1.11 exposes
+// legacy conditions on v1beta2 objects under the v1beta1 contract compat layer.
+func isPresentAndFalseWithReasonV2(key types.NamespacedName, obj client.Object, condition clusterv1.ConditionType, reason string) bool {
```
janiskemper, line 460: why don't we check v1beta2 conditions for CAPI objects directly?

janiskemper (changes requested):

test/e2e/config/hetzner.yaml
```diff
@@ -14,8 +14,8 @@ providers:
   - name: cluster-api
     type: CoreProvider
     versions:
-      - name: v1.10.10
-        value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.10.10/core-components.yaml"
+      - name: v1.11.8
```
janiskemper, line 17: I think you should also change the version of the CAPI objects you apply to v1beta2 in the templates

test/e2e/config/hetzner.yaml
```diff
@@ -31,8 +31,8 @@ providers:
   - name: kubeadm
     type: BootstrapProvider
     versions:
-      - name: v1.10.10
-        value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.10.10/bootstrap-components.yaml"
+      - name: v1.11.8
+        value: "https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.11.8/bootstrap-components.yaml"
         type: "url"
         contract: "v1beta1"
```
janiskemper, line 37: I think this should be updated as well

controllers/hcloudmachine_controller.go
```diff
@@ -200,8 +200,8 @@ func (r *HCloudMachineReconciler) Reconcile(ctx context.Context, req reconcile.R
 			}
 		}
 
-		readyReason := conditions.GetReason(machineScope.HCloudMachine, clusterv1.ReadyCondition)
-		readyMessage := conditions.GetMessage(machineScope.HCloudMachine, clusterv1.ReadyCondition)
+		readyReason := v1beta1conditions.GetReason(machineScope.HCloudMachine, clusterv1.ReadyCondition)
```
janiskemper: this mixes v1beta2 and v1beta1

controllers/hcloudmachinetemplate_controller_test.go
```diff
@@ -54,25 +55,25 @@ var _ = Describe("HCloudMachineTemplateReconciler", func() {
 
 	Context("Basic hcloudmachinetemplate test", func() {
 		Context("ClusterClass test", func() {
-			var capiClusterClass *clusterv1.ClusterClass
+			var capiClusterClass *clusterv1beta1.ClusterClass
 
 			BeforeEach(func() {
-				capiClusterClass = &clusterv1.ClusterClass{
+				capiClusterClass = &clusterv1beta1.ClusterClass{
```
janiskemper: why do we use the old clusterClass?

controllers/hetznerbaremetalmachine_controller_test.go
```diff
@@ -698,7 +708,7 @@ var _ = Describe("HetznerBareMetalMachineReconciler", func() {
 					if err != nil {
 						return err
 					}
-					c := conditions.Get(host, clusterv1.ReadyCondition)
+					c := v1beta1conditions.Get(host, clusterv1.ReadyCondition)
```
janiskemper: there are multiple occurrences of ReadyCondition in combination with v1beta2

test/helpers/envtest.go
```diff
@@ -82,6 +83,7 @@ func init() {
 	// Calculate the scheme.
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(clusterv1.AddToScheme(scheme))
+	utilruntime.Must(clusterv1beta1.AddToScheme(scheme))
```
janiskemper: IMO this shouldn't be necessary. Or should it?

janiskemper (changes requested):

controllers/controllers_suite_test.go
abdullah599: This function is mostly used for things like remediation where we write conditions. Those test reads from the same place our remediation writes. Our remediation calls `util/conditions/deprecated/v1beta1.MarkFalse` with `MachineOwnerRemediatedV1Beta1Condition`, which writes to `Status.Deprecated.V1Beta1.Conditions`. The test helper reads the same field. If the test read the v1beta2 primary list (`Status.Conditions []metav1.Condition`) it would always return false since we never write there.

janiskemper: Maybe I'm missing something - but I'm rather sure I saw this used for CAPI machines.

controllers/hcloudmachine_controller_test.go
abdullah599: The Machine CRD in CAPI v1.11 marks spec.bootstrap as a required field. A zero value Bootstrap{} is dropped by the API server, so creating the Machine fails with "spec.bootstrap: Required value".

I verified this by pushing a version with Bootstrap{} instead of the ConfigRef block. CI failed with that exact error (run 24766940909).

The ConfigRef block is a placeholder only. DataSecretName stays nil, which is what IsBootstrapDataReady checks, so the reconciler still treats the Machine as "bootstrap not ready" and the test still exercises the same path.

janiskemper: interesting

controllers/controllers_suite_test.go
abdullah599: Yes, `isPresentAndFalseWithReasonV2` is used on CAPI Machine objects. Machines are v1beta2, but our remediation writes conditions to the legacy shadow field `Status.Deprecated.V1Beta1.Conditions` via `util/conditions/deprecated/v1beta1.MarkFalse`, not to the primary `Status.Conditions []metav1.Condition` list.

The helper reads from the same shadow field where we wrote. If it read from the primary list instead, it would always return nil since we never write there.

Moving the reads and writes to the primary list is the v1beta2 contract migration.

janiskemper: Okay so we use it only in the context of remediation right? Then we can resolve this.

pkg/scope/machine.go
abdullah599: the `!= nil` in the second clause is redundant since the first clause already handles the nil case ,no?

janiskemper: ah I think you are right, but can you just quickly verify that in a simple program?

abdullah599: Done!

<details>
<summary>Go code</summary>

```go
package main

import "fmt"

func strPtr(s string) *string { return &s }

// original: if m.HCloudMachine.Spec.ProviderID == nil || m.HCloudMachine.Spec.ProviderID != nil && *m.HCloudMachine.Spec.ProviderID == ""
func original(providerID *string) bool {
    return providerID == nil || providerID != nil && *providerID == ""
}

// simplified: if m.HCloudMachine.Spec.ProviderID == nil || *m.HCloudMachine.Spec.ProviderID == ""
func simplified(providerID *string) bool {
    return providerID == nil || *providerID == ""
}

func main() {
    cases := []struct {
        name       string
        providerID *string
    }{
        {"nil pointer", nil},
        {"empty string", strPtr("")},
        {"valid hcloud", strPtr("hcloud://12345")},
        {"invalid prefix", strPtr("aws://12345")},
        {"just prefix", strPtr("hcloud://")},
    }

    fmt.Printf("%-20s %-10s %-10s %-10s\n", "case", "original", "simplified", "match")
    fmt.Println("---------------------------------------------------------")
    allMatch := true
    for _, c := range cases {
        o := original(c.providerID)
        s := simplified(c.providerID)
        match := o == s
        if !match {
            allMatch = false
        }
        fmt.Printf("%-20s %-10v %-10v %-10v\n", c.name, o, s, match)
    }
    fmt.Println("---------------------------------------------------------")
    if allMatch {
        fmt.Println("RESULT: all cases match, original and simplified are equivalent")
    } else {
        fmt.Println("RESULT: mismatch found")
    }
}

```

</details>

image

pkg/services/hcloud/server/server_test.go
abdullah599: Same reason as: https://github.com/syself/cluster-api-provider-hetzner/pull/1990#discussion_r3122483558

test/e2e/config/hetzner.yaml
abdullah599: Quick note on scope, since "move to v1beta2" can mean a few different things in this PR:

1. Go imports for CAPI core moved to v1beta2. Done in this PR. This is what required adapting to the new ContractVersionedObjectReference type on InfrastructureRef and ControlPlaneRef (covered in Key Decision 5 in the PR body).

2. Our CRD contract. Stays on v1beta1 for this PR. metadata.yaml still declares contract: v1beta1.

3. CAPI core object apiVersions in the e2e templates (what you're pointing at here). Still v1beta1 in this PR, grouped with 2.

Because 1 is done, our controllers already read Cluster, Machine, etc. as v1beta2 through CAPI's conversion webhook, regardless of which apiVersion the template YAML uses. So the v1beta1 templates work end to end today.

Flipping the template apiVersions isn't a string swap. v1beta2 changes the spec shape too (InfrastructureRef moves to {apiGroup, kind, name} with no namespace, similar changes on Machine.Bootstrap and NodeRef), so every ref block in the templates would need rewriting. That belongs with the CRD contract move in the follow up PR, before the v1beta1 compat layer is removed upstream around August 2026.

janiskemper: I disagree here, but we can do it in a follow-up PR. Then you should explicitly point that we wanna update the templates. 

We use already the v1beta2 CAPI objects in the code, therefore there is no reason at all to not switch to them in the templates as well. 

The conclusion is not correct that it belongs to the v1beta2 contract, because that is about using v1beta2 CAPH objects - not v1beta2 CAPI objects. We already started using v1beta2 CAPI objects in this PR, therefore we should soon switch everywhere. If not in this PR, then in a follow-up one.

abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/1992

test/e2e/config/hetzner.yaml
abdullah599: https://github.com/syself/cluster-api-provider-hetzner/pull/1990#discussion_r3123350729

controllers/hcloudmachinetemplate_controller_test.go
```diff
@@ -64,29 +65,17 @@ var _ = Describe("HCloudMachineTemplateReconciler", func() {
 					},
 					Spec: clusterv1.ClusterClassSpec{
 						ControlPlane: clusterv1.ControlPlaneClass{
-							MachineInfrastructure: &clusterv1.LocalObjectTemplate{
```
Dhairya-Arora01, line 67: why was MachineInfrastructure removed ?

Dhairya-Arora01 (commented):
Looks good overall. cc @janiskemper

controllers/hcloudmachinetemplate_controller_test.go
abdullah599: This test only asserts that machineTemplate.Status.OwnerType == "ClusterClass", which is derived from the direct OwnerReference on the machineTemplate, not from the ClusterClass spec. When migrating the fixture to v1beta2, I dropped MachineInfrastructure because it does not affect the test. I'll add it to remove confusion :)

janiskemper (approved):

PR #1995, opened 2026-04-24
:seedling: When image-url-command is missing, don't create server.

guettli:
When image-url-command is missing, don't create server.

Otherwise:

> InfrastructureReady: failed to create HCloud server foo in LOCATION (type TYPE): server name is already used (uniqueness_error, ...)

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -63,6 +63,8 @@ var hcloudImageURLCommandDir = "/shared"
 
 var errServerCreateNotPossible = errors.New("server create not possible - need action")
 
+var errServerCreateStopReconcile = errors.New("stopped Reconciling")
+
 // Service defines struct with machine scope to reconcile HCloudMachines.
 type Service struct {
 	scope *scope.MachineScope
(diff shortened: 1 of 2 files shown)


janiskemper (commented):
LGTM. @Dhairya-Arora01 for a final review

pkg/services/hcloud/server/server.go
```diff
@@ -1039,9 +1024,32 @@ func (s *Service) createServerFromImageNameOrURL(ctx context.Context) (*hcloud.S
...
+	}
+
+	// The webhook already validates this, but we check again before any file access because we
+	// never want a HCloudMachine to make CAPH copy arbitrary controller files into the rescue
+	// system, for example service-account tokens. The webhook could also have been disabled
+	// temporarily, so this runtime check is still meaningful.
+	if _, err := utils.ResolveImageURLCommandPath(hcloudImageURLCommandDir, imageURLCommandName); err != nil {
+		err = fmt.Errorf("imageURLCommand %q is invalid or not accessible by the controller pod: %w", imageURLCommandName, err)
+		s.scope.Error(err, "")
+		v1beta1conditions.MarkFalse(s.scope.HCloudMachine, infrav1.ServerAvailableCondition,
```
Dhairya-Arora01: Why do we mark ServerAvailable condition as false here but ServerProvisionedCondition above ?

guettli: @Dhairya-Arora01 I moved to code block from an other location to that place.

Do you think I should change it? I don't have an opinion here at the moment.

Dhairya-Arora01: I think we should align it @guettli

Dhairya-Arora01 (approved):

PR #1998, opened 2026-04-25
🐛 Retry on conflict when claiming the shared Hetzner Secret

### LINKED ISSUE #1997: Wrap `client.Update` in `retry.RetryOnConflict` in `claimSecret` to stop 409 conflicts on the shared Hetzner Secret
## Summary

Stop 409 "object has been modified" errors on the `hetzner` Secret by wrapping the `client.Update` call in `pkg/secrets/secret.go` with `retry.RetryOnConflict`, re-fetching the Secret on conflict and re-applying the same changes.

## The Problem

In each org namespace, many CAPH controllers share the same `hetzner` Secret (HetznerCluster, HCloudMachineTemplate, HCloudMachine, HetznerBareMetalMachine, HetznerBareMetalHost, and HCloudRemediation). All of them call `SecretManager.AcquireSecret`, which calls `claimSecret` in `pkg/secrets/secret.go`. `claimSecret` ensures the Secret has three pieces of metadata set:

- A cache label `LabelEnvironmentName`.
- An owner reference pointing to the `HetznerCluster`.
- The `SecretFinalizer` so the Secret cannot be deleted while in use.

`claimSecret` has guards (`HasLabel`, `alreadyOwned`, `AddFinalizer`) that return early when the field is already set. In steady state, `needsUpdate` stays `false` and no write is made. This part works fine.

**Why we see it on the hetzner Secret?**
The conflict only appears during the very first setup. At that point, the Secret has no label, no owner, and no finalizer yet. Multiple controllers may read the Secret at the same moment, all see the fields missing, and all try to add them. One wins, the others get the 409.

## The Solution

Wrap the `client.Update` call in `retry.RetryOnConflict`. On conflict, re-fetch the Secret from the API server and let the retry re-apply the changes. The guards inside `claimSecret` are idempotent, so the next retry either sees that all fields are already set (and returns with no write) or applies just the missing fields (and writes cleanly).

abdullah599:
This PR wraps the `client.Update` call in `claimSecret` with `retry.RetryOnConflict`. On conflict, we re-fetch the Secret so the next retry sees the latest state, and let the idempotent guards (`HasLabel`, `alreadyOwned`, `AddFinalizer`) either fill in whatever the winner did not set or return early. The backoff matches the one used in `sigs.k8s.io/cluster-api/util/patch/patch.go` (5 steps, 100ms, full jitter)

Fixes #1997 

None.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/secrets/secret.go
+++ b/pkg/secrets/secret.go
@@ -20,12 +20,15 @@ package secretutil
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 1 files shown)


pkg/secrets/secret.go
```diff
@@ -62,51 +65,75 @@ func NewSecretManager(log logr.Logger, cacheClient client.Client, apiReader clie
...
+		Steps:    5,
+		Duration: 100 * time.Millisecond,
+		Jitter:   1.0,
 	}
-	if owner != nil {
-		if ownerIsController {
-			if !metav1.IsControlledBy(secret, owner) {
-				if err := controllerutil.SetControllerReference(owner, secret, sm.client.Scheme()); err != nil {
-					return fmt.Errorf("failed to set secret controller reference: %w", err)
+	err := retry.RetryOnConflict(backoff, func() error {
```
janiskemper: it seems as if the wait package is easier and makes retry obsolete.. https://github.com/kubernetes-sigs/cluster-api/blob/cf3d8f81e95cc8b7305daabbd0ad2c8440e83c83/util/patch/patch.go#L351

janiskemper (approved):

PR #2004, opened 2026-05-02
🌱 Improve Quick Start Guide 

abdullah599:

The Quick Start Guide does not produce a working cluster as written. The main reason is that the cluster template, the secret the guide tells the user to create, and the Syself CCM all expect a different secret name or key. A user who follows the guide ends up with `HCloudCredentialsInvalid` on the HetznerCluster and a Syself CCM that crash loops with `secret "hetzner" not found`.

This PR makes the guide work end to end by aligning everything on one secret shape: `name: hcloud`, `key: hcloud`. This is what the Syself CCM (the default in the guide) supports. CAPH already has a helper that, when the secret is named `hcloud`, also writes the token under key `token` in the workload cluster, so the upstream HCloud CCM and the hcloud-csi chart also work without any extra setup.

The PR also fixes a few smaller issues in the same flow that were tripping up new users.

<details>
<summary><strong>Why secret name <code>hcloud</code> and key <code>hcloud</code> (and not the upstream pattern <code>hcloud</code> / <code>token</code>)</strong></summary>

The upstream Hetzner pattern is `hcloud` / `token`, and that was the first proposal in the team discussion. The blocker is that the Syself CCM v2 chart hardcodes the token file name to `hcloud` and does not let the user override it. Since the quickstart installs the Syself CCM by default, we have to use a key the chart can actually read.

CAPH's compatibility helper in `controllers/hetznercluster_controller.go` (`keysForWorkloadClusterSecret`) saves us from picking sides: when the secret is named `hcloud`, it writes the token under both the configured key and a hardcoded `token` key in the workload cluster. Note that this duplication only triggers when the configured key is not already `token`, which is why we set the key to `hcloud` instead. So one secret with one key satisfies all three downstream consumers without any per component overrides:

- Syself CCM reads the file `hcloud`
- Upstream HCloud CCM reads env var `HCLOUD_TOKEN` from key `token`
- hcloud-csi reads env var `HCLOUD_TOKEN` from key `token`
</details>

<details>
<summary><strong>List of changes</strong></summary>

**Templates**

- `templates/cluster-templates/v1beta1/bases/hcloud-hetznerCluster.yaml` and `hcloud-hetznerCluster-network.yaml`: change `hcloudToken: token` to `hcloudToken: hcloud` so the template matches the secret the guide creates.

**Page 02 (`02-management-cluster-setup.md`)**

- Fix the secret create and patch commands to use name `hcloud` and key `hcloud`.
- Mark the SSH key step as required (the template hardcodes `${SSH_KEY_NAME}`).
- Add a short paragraph explaining that CAPH auto syncs this secret to the workload cluster, so the user knows they do not need to create it again later.
- Add a small verify step that prints the token length.

**Page 03 (`03-creating-a-workload-cluster.md`)**

- Tell the user to wait for `INITIALIZED: True` on the KCP before fetching the workload kubeconfig, with a callout for the `EOF` error users hit when they fetch too early.
- Replace `make install-ccm-in-wl-cluster` with the equivalent `helm` command. The make target reads the wrong env var and forces the user to be inside a repo checkout.
- Install the Cilium values from a raw GitHub URL so the user does not need a repo checkout.
- Add a short subsection on using the upstream Hetzner CCM, which works as is because of CAPH's key duplication.
- CSI: drop the `existingSecret` overrides (chart defaults already match) and add the missing `helm repo add hcloud` step.

**Page 04 (`04-next-steps.md`)**

- Make the `~/.kube/config` assumption explicit when switching back to the management cluster.
</details>

Fixes #

See the dropdowns above.

**TODOs**:
- [ ] squash commits
- [ ] include documentation

### DIFF
--- a/docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
+++ b/docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
@@ -22,7 +22,7 @@ There are several tasks that have to be completed before a workload cluster can 
 
 1. Create a new [HCloud project](https://console.hetzner.cloud/projects).
 1. Generate an API token with read and write access. You'll find this if you click on the project and go to "security".
-1. If you want to use it, generate an SSH key, upload the public key to HCloud (also via "security"), and give it a name. Read more about [Managing SSH Keys](/docs/caph/02-topics/01-managing-ssh-keys.md).
+2. Generate an SSH key, upload the public key to HCloud (also via "security"), and give it a name. Read more about [Managing SSH Keys](/docs/caph/02-topics/01-managing-ssh-keys.md).
 
 ## Bootstrap or Management Cluster Installation
 
@@ -107,7 +107,9 @@ Optional Variables:
   - WORKER_MACHINE_COUNT         (defaults to 0)
 ```
 
-## Create a secret for hcloud only
+## Create the HCloud secret
+
+This guide covers the pure HCloud flow (no bare metal), so we only create the HCloud secret here.
 
 In order for the provider integration hetzner to communicate with the Hetzner API ([HCloud API](https://docs.hetzner.cloud/)), we need to create a secret with the access data. The secret must be in the same namespace as the other CRs.
 
@@ -118,13 +120,23 @@ In order for the provider integration hetzner to communicate with the Hetzner AP
 Use the below command to create the required secret with the access data:
 
 ```shell
-kubectl create secret generic hetzner --from-literal=hcloud=$HCLOUD_TOKEN
+kubectl create secret generic hcloud --from-literal=hcloud=$HCLOUD_TOKEN
 ```
 
 Patch the created secret so that it can be automatically moved to the target cluster later. The following command helps you do that:
 
 ```shell
-kubectl patch secret hetzner -p '{"metadata":{"labels":{"clusterctl.cluster.x-k8s.io/move":""}}}'
+kubectl patch secret hcloud -p '{"metadata":{"labels":{"clusterctl.cluster.x-k8s.io/move":""}}}'
 ```
 
-The secret name and the tokens can also be customized in the cluster template.
+Why one secret with one key (`hcloud`) is enough: the cluster template's `hetznerSecretRef` is configured to read the token from key `hcloud`. CAPH then auto-syncs this secret into the workload cluster's `kube-system` namespace and, for backward compatibility, also exposes the token under key `token`. That single secret therefore satisfies the Syself CCM (reads file `hcloud`), the upstream HCloud CCM, and the Hetzner CSI driver (both expect key `token`) — without any per-component overrides. You do not need to create a secret in the workload cluster manually.
+
+## Verify the secret
+
+Before applying the cluster, confirm the secret exists with a non-empty value:
+
+```shell
+kubectl get secret hcloud -o jsonpath='{.data.hcloud}' | base64 -d | wc -c
+```
+
+You should see roughly 64 (the length of an HCloud token). An empty output means the secret name or key is wrong, and the controller will fail to reconcile the cluster.
--- a/docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
+++ b/docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
@@ -129,7 +129,7 @@ Patch the created secret so that it can be automatically moved to the target clu
 kubectl patch secret hcloud -p '{"metadata":{"labels":{"clusterctl.cluster.x-k8s.io/move":""}}}'
 ```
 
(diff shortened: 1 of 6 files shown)


docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
```diff
@@ -118,13 +120,23 @@ In order for the provider integration hetzner to communicate with the Hetzner AP
...
 ```
 
 Patch the created secret so that it can be automatically moved to the target cluster later. The following command helps you do that:
 
 ```shell
-kubectl patch secret hetzner -p '{"metadata":{"labels":{"clusterctl.cluster.x-k8s.io/move":""}}}'
+kubectl patch secret hcloud -p '{"metadata":{"labels":{"clusterctl.cluster.x-k8s.io/move":""}}}'
+```
+
+Why one secret with one key (`hcloud`) is enough: the cluster template's `hetznerSecretRef` is configured to read the token from key `hcloud`. CAPH then auto-syncs this secret into the workload cluster's `kube-system` namespace and, for backward compatibility, also exposes the token under key `token`. That single secret therefore satisfies the Syself CCM (reads file `hcloud`), the upstream HCloud CCM, and the Hetzner CSI driver (both expect key `token`) — without any per-component overrides. You do not need to create a secret in the workload cluster manually.
```
janiskemper: if I'm not mistaken, the Syself CCM reads from secret "hetzner" automatically not "hcloud"

docs/caph/01-getting-started/02-quickstart/03-creating-a-workload-cluster.md
```diff
@@ -56,65 +56,81 @@ To verify the first control plane is up, use the following command:
...
 
-## Deploy the CCM
+## Deploy the Cloud Controller Manager (CCM)
+
+The CCM runs in the workload cluster and integrates Kubernetes with the Hetzner APIs. In practice, it is responsible for tasks such as setting the `ProviderID` on nodes and managing load balancers. You need to install a CCM in this flow so the workload cluster can properly interact with Hetzner infrastructure.
 
-The CCM (Cloud Controller Manager) runs in the workload cluster and integrates Kubernetes with the
-Hetzner APIs. In practice, it is responsible for tasks such as setting the `ProviderID` on nodes
-and managing load balancers.
+CAPH already syncs the `hcloud` secret you created on the previous page into the workload cluster's `kube-system` namespace, so the CCM only needs to be pointed at it (the chart default is `hetzner`, so we override `secret.name`):
```
janiskemper, line 107: From here onward I would put it under the point "Syself CCM". It shouldn't actually be used as "default" here. We want to move slowly to the upstream one

templates/cluster-templates/v1beta1/bases/hcloud-hetznerCluster-network.yaml
```diff
@@ -20,6 +20,6 @@ spec:
   hetznerSecretRef:
     name: hcloud
     key:
-      hcloudToken: token
+      hcloudToken: hcloud
```
janiskemper, line 26: this is necessary to make it work with the syself CCM, right? Because "token" is anyway created for migration purposes?

docs/caph/01-getting-started/02-quickstart/02-management-cluster-setup.md
abdullah599: Yes but we set `--set secret.name=hcloud` so it would read from hcloud

janiskemper: Then I find that sentence misleading. But I believe that the actual discussion is on the level of the keys. I have the other comment for that below.

templates/cluster-templates/v1beta1/bases/hcloud-hetznerCluster-network.yaml
abdullah599: yes

Little context:
`spec.hetznerSecretRef.key.hcloudToken` tells CAPH which key in the management-cluster secret holds the token. So this line decides how the user creates the secret on the management side, not what gets generated downstream.

Why we need hcloud here: the Syself CCM v2 chart hardcodes the token file name to hcloud and only secret.name is configurable, the key is not. Since the quickstart installs the Syself CCM by default, we need the management-side secret (which CAPH then mirrors into the workload cluster) to actually have a key called hcloud

janiskemper: we don't install the Syself CCM by default  ;) I already commented that. However, there is no way to use "token" here in the charts and then use the Syself CCM. Therefore, I agree that it is better, as long as we somehow say that the Syself CCM can be used, to use "hcloud" here instead of "token". 

We should add a comment though, that we put "hcloud" here for backward-compatibility reasons with the Syself CCM - so that the templates can be used for both the fork and the upstream CCM. Once we drop Syself CCM, we should use "token" here. 

You can add that as a comment to the templates.

janiskemper (approved):

janiskemper:
@guettli plz also have a look at these changes. I suggest to merge them soon, to not confuse people. But you should still have a look

guettli (approved):
lgtm

PR #2007, opened 2026-05-06
:seedling: Set hostname to stripped bm- version of desired name

Dhairya-Arora01:
Currently, during deprovisioning the hostname is changed to the name of the HBMM. However, while provisioning the host we check if a stable/constant hostname should be used and explicitly set it accordingly. By switching back to the HBMM name during deprovisioning, we lose this deterministic naming and end up with inconsistent hostnames across the lifecycle of the machine.
Using the stripped "bm-" version preserves the stable hostname format introduced for node recognition (for example by TopoLVM) while still making it clear in the Hetzner UI that the machine is no longer actively provisioned.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


guettli (changes requested):
The comment explains **what** gets done. I would like to see the **why** this is done.

guettli (approved):

PR #2012, opened 2026-05-09
✨ Add remediation cooldown and fix external remediation success path

### LINKED ISSUE #1983: 🐛 Reboot remediation always deletes the Machine, even when the reboot succeeded
## Summary

CAPH's reboot remediation for `HCloudMachine` and `HetznerBareMetalMachine` has a core bug: every remediation ends in Machine deletion, even when the reboot brought the Node back healthy. Fixing that bug safely requires one more change: a cooldown to prevent reboot loops on chronic failures.

This issue covers both: the bug fix and the guard that must land with it.

### 1. Reboot remediation always deletes the Machine, even when the reboot worked

When a Machine fails a health check:

1. CAPI's `MachineHealthCheck` sets `MachineOwnerRemediatedCondition=False` on the Machine and creates an external remediation CR (`HCloudRemediation` or `HetznerBareMetalRemediation`).
2. CAPH reboots the server, sets `Status.LastRemediated`, increments `Status.RetryCount`, and transitions to `Phase=Waiting`.
3. After `retryLimit × timeout` elapses, the controller re-enters `handlePhaseWaiting`.

Based on CAPI's contract for `MachineOwnerRemediatedCondition` ([`vendor/sigs.k8s.io/cluster-api/api/v1beta1/condition_consts.go:167-169`](https://github.com/syself/cluster-api-provider-hetzner/blob/main/vendor/sigs.k8s.io/cluster-api/api/v1beta1/condition_consts.go#L167-L169)):

```go
// MachineOwnerRemediatedCondition is set on machines that have failed a healthcheck
// by the MachineHealthCheck controller.
// MachineOwnerRemediatedCondition is set to False after a health check fails, but
// should be changed to True by the owning controller after remediation succeeds.
MachineOwnerRemediatedCondition ConditionType = "OwnerRemediated"
```

The owning controller (CAPH) should set this to `True` when the reboot brought the Node back. Instead, `handlePhaseWaiting` always takes the failure path:

```go
// pkg/services/hcloud/remediation/remediation.go:162-179
func (s *Service) handlePhaseWaiting(ctx context.Context) (reconcile.Result, error) {
    nextCheck := s.timeUntilNextRemediation(time.Now())
    if nextCheck > 0 {
        return reconcile.Result{RequeueAfter: nextCheck}, nil
    }
    err := s.setOwnerRemediatedConditionToFailed(ctx,
        "exit remediation because because retryLimit is reached and reboot timed out")
    ...
}
```

The bare metal service has the same issue at `pkg/services/baremetal/remediation/remediation.go:172-181`.

Result: every remediation CR ends with `MachineOwnerRemediated=False`. CAPI deletes the Machine. `MachineDeployment` creates a replacement. The reboot path cannot succeed today.

### 2. Chronically unhealthy machines can be rebooted indefinitely

This problem does not exist today because every remediation ends in deletion. It appears once Problem 1 is fixed.

After the fix, if a machine has a chronic issue that still lets it boot (flaky disk, intermittent NIC), the cycle becomes: MHC detects unhealthy, reboot succeeds, MHC detects unhealthy again, reboot succeeds, and so on. Each MHC incident creates a fresh Remediation CR with `RetryCount=0`, so there is no memory across incidents. Nothing stops the loop.

## Proposed Solution

Two coordinated changes. Gated on a `RemediationStrategy` field, so unmanaged users can keep current behavior if they want.

### A. Add the success path to `handlePhaseWaiting`

In both services:

1. Before declaring failure, check whether the Node is healthy again.
   * **HCloud:** Node `bootID` changed from the pre-reboot value and the Node is reachable.
   * **BareMetal:** the equivalent reachability signal (existing checks in `actionProvisioned` / host state).
2. If healthy, set `MachineOwnerRemediatedCondition=True`, remove the `cluster.x-k8s.io/remediate-machine` annotation from the Machine, stamp `LastRemediatedAt` on the provider Machine (see B), and transition the CR to `PhaseSucceeded`.
3. If not healthy when the window expires, take the existing failure path.

### B. Add a cooldown to `RemediationStrategy`

New field on `RemediationStrategy`:

```go
// Cooldown is the minimum time between successive remediations of the same
// machine. If a new remediation starts within this window of the previous
// successful one, the reboot is skipped and the machine is deleted directly.
// Set to 0s to disable the guard.
// +kubebuilder:default="30m"
// +optional
Cooldown *metav1.Duration `json:"cooldown,omitempty"`
```

New field on `HCloudMachineStatus` and `HetznerBareMetalMachineStatus`:

```go
// LastRemediatedAt records when the most recent successful remediation
// completed. Used to prevent reboot loops across successive MHC incidents.
// +optional
LastRemediatedAt *metav1.Time `json:"lastRemediatedAt,omitempty"`
```

Controller logic: at the start of `Reconcile`, if `Machine.Status.LastRemediatedAt` is within `Strategy.Cooldown`, skip the reboot and transition the CR to `PhaseDeleting` with reason `RecentlyRemediated`. Otherwise, run the normal reboot flow. On success, stamp `LastRemediatedAt = now` on the provider Machine before marking the CR succeeded.

The timestamp lives on the provider Machine. It survives Remediation CR deletion, and it resets naturally when the Machine itself is deleted. A new Machine starts fresh.

abdullah599:
Add remediation cooldown and external remediation success path

Fixes #1983

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


pkg/services/baremetal/remediation/remediation.go
```diff
@@ -95,6 +96,24 @@ func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
...
 
+	// Skip remediation when the previous successful remediation is within the cooldown window.
+	// Only evaluated on a fresh CR (Phase is empty) to avoid interrupting in-progress remediations.
+	if s.scope.BareMetalRemediation.Status.Phase == "" {
+		cooldown := s.scope.BareMetalRemediation.Spec.Strategy.EffectiveCooldown()
+		if cooldown > 0 && s.scope.BareMetalMachine.Status.LastRemediatedAt != nil {
+			since := time.Since(s.scope.BareMetalMachine.Status.LastRemediatedAt.Time)
+			if since < cooldown {
+				err := s.markRemediationSkipped(ctx,
+					fmt.Sprintf("previous remediation was %s ago within cooldown window %s", since.Round(time.Second), cooldown))
```
janiskemper: are you sure that this is well readable? Just asking. With the different time formats and the %s

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -178,6 +197,15 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 		return reconcile.Result{RequeueAfter: nextCheck}, nil
 	}
 
+	capiMachine, err := util.GetOwnerMachine(ctx, s.scope.Client, s.scope.BareMetalRemediation.ObjectMeta)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return reconcile.Result{}, fmt.Errorf("failed to get owner machine: %w", err)
+	}
+	if capiMachine != nil && deprecatedv1beta1conditions.IsTrue(capiMachine, clusterv1.MachineNodeHealthyCondition) {
```
janiskemper: you check here with the deprecated package if a v1beta2 condition is set.. Seems wrong

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -248,6 +281,87 @@ func (s *Service) setOwnerRemediatedConditionToFailed(ctx context.Context, msg s
 	return nil
 }
 
+// markRemediationSucceeded clears the remediate-machine annotation (CAPI MHC
```
janiskemper: can you add here on which custom resources these actions are performed

pkg/services/baremetal/remediation/remediation.go
abdullah599: Updated. Both durations are now rounded to the second (so the format is consistent) and the cooldown window is grouped in brackets

janiskemper (approved):

PR #2013, opened 2026-05-09
:seedling: Introduce v1beta2 conditions for HetznerBareMetalHost

### LINKED ISSUE #2011: Introduce v1beta2 conditions for HetznerBareMetalHost
## Summary

Add v1beta2 conditions to `HetznerBareMetalHost` while keeping the existing v1beta1 conditions unchanged.

## What would you like to add

`HetznerBareMetalHost` is a primary infrastructure object that represents a real Hetzner dedicated (Robot) server and drives its provisioning state machine (rescue, image install, OS running, deprovisioning, deleting).

Today the host lifecycle is surfaced through v1beta1 conditions such as:

| Condition | Purpose |
|---|---|
| `CredentialsAvailable` | SSH credentials in the referenced secret are valid |
| `RobotCredentialsAvailable` | Hetzner Robot API credentials are valid and reachable |
| `RootDeviceHintsValidated` | The configured root device hints match the discovered hardware |
| `ProvisionSucceeded` | The host has reached `OperatingSystemRunning` (provisioning finished) |
| `HetznerAPIReachable` | The controller is not blocked by Robot API rate limiting |

These conditions are still needed and must keep their current behavior. At the same time, the CAPI v1beta2 transition requires `HetznerBareMetalHost` to expose v1beta2 conditions in parallel.

Deletion also needs to be visible more clearly. Today deletion is signalled through `Spec.Status.ProvisioningState = StateDeleting` and finalizer removal in `actionDeleting`, and the v1beta1 `HostReadyCondition` is marked False with reason `DeletionInProgress` while the host is being deleted. Outside of deletion, `HostReadyCondition` is only used by the `HetznerBareMetalMachine` controller (which mirrors it onto the machine). We want a dedicated v1beta2 `Deleting` condition for `HetznerBareMetalHost`, since this resource has a finalizer and a delete reconciliation path.

> [!note]
> `HetznerBareMetalHostStatus` is an empty struct today: the controller-managed status (including `Conditions`) lives under `Spec.Status` (`ControllerGeneratedStatus`). The placement of `status.v1beta2.conditions` will follow the same approach used for the existing v1beta1 conditions on this type so that no schema reshuffle is required for this phase.

## The Solution

Follow the same structure used for `HCloudMachine` and `HetznerBareMetalMachine`:

- Add `v1beta2.conditions` to `HetznerBareMetalHost` alongside the existing v1beta1 conditions.
- Keep writing the existing v1beta1 conditions exactly as today.
- Add direct `v1beta2conditions.Set(...)` calls in the same controller and host-service code paths that already set the v1beta1 conditions.
- Add a controller local patch step for `HetznerBareMetalHost` that computes:
  - the legacy v1beta1 summary
  - the v1beta2 `Ready` summary
  - the owned v1beta2 condition set for patching
- Rename `HetznerAPIReachable` to `RobotRateLimitExceeded` in v1beta2 with **negative polarity** (True = rate limited). Set to False on every successful reconcile (happy path setter).
- Add a dedicated v1beta2 `Deleting` condition (negative polarity).
 - Replace the v1beta1 `HostReadyCondition` set-to-False-on-deletion signal with the new dedicated v1beta2 `Deleting` condition, so the deletion state is surfaced through its own condition instead of overloading `HostReadyCondition`. This makes the deletion state transparent and leaves `HostReadyCondition` for its existing use by the `HetznerBareMetalMachine` controller.

This keeps the behavior explicit at each reconcile decision point and avoids a second translation layer that can drift from the real lifecycle logic.

## Final Conditions

The final v1beta2 conditions added for `HetznerBareMetalHost` are:

| Constant | Type | Polarity | Notes |
|---|---|---|---|
| `HetznerBareMetalHostReadyV1Beta2Condition` | `clusterv1.ReadyV1Beta2Condition` | Positive | Summary condition |
| `HetznerBareMetalHostCredentialsAvailableV1Beta2Condition` | `CredentialsAvailable` | Positive | True when the SSH credentials in the referenced secret are valid |
| `HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Condition` | `RobotCredentialsAvailable` | Positive | True when the Hetzner Robot API credentials are valid and reachable |
| `HetznerBareMetalHostRootDeviceHintsValidatedV1Beta2Condition` | `RootDeviceHintsValidated` | Positive | True when the configured root device hints match the discovered hardware |
| `HetznerBareMetalHostProvisionSucceededV1Beta2Condition` | `ProvisionSucceeded` | Positive | True when the host has reached `OperatingSystemRunning` |
| `HetznerBareMetalHostDeletingV1Beta2Condition` | `clusterv1.DeletingV1Beta2Condition` | Negative | Dedicated deletion signal for `HetznerBareMetalHost`; replaces the v1beta1 `HostReadyCondition` set-to-False-on-deletion signal |
| `HetznerBareMetalHostRobotRateLimitExceededV1Beta2Condition` | `RobotRateLimitExceeded` | Negative | True when the Robot API rate limit has been exceeded. Set to False on every successful reconcile (happy path setter). |

abdullah599:
Introduced v1beta2 conditions for HetznerBareMetalHost

Fixes #2011

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,104 @@ const (
...
 )
+
+// v1beta2 conditions for HetznerBareMetalHost.
+
+// InternalErrorV1Beta2Reason is a generic reason for internal errors during summary computation
+// or other unexpected failures that do not have a more specific reason.
+const InternalErrorV1Beta2Reason = "InternalError"
+
+const (
+	// HetznerBareMetalHostReadyV1Beta2Condition is the summary condition for HetznerBareMetalHost.
```
janiskemper: didn't we say that we wanna use the generic shared conditions here?

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,104 @@ const (
...
+	HetznerBareMetalHostHetznerSecretUnreachableV1Beta2Reason = "HetznerSecretUnreachable" // #nosec
+)
+
+const (
+	// HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Condition reports whether Robot API credentials are valid and reachable.
+	HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Condition = "RobotCredentialsAvailable"
+	// HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Reason indicates the Robot credentials are available.
+	HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Reason = "Available"
+	// HetznerBareMetalHostRobotCredentialsInvalidV1Beta2Reason indicates Robot credentials are invalid.
+	HetznerBareMetalHostRobotCredentialsInvalidV1Beta2Reason = "RobotCredentialsInvalid" // #nosec
```
janiskemper: we need only "Invalid" here right? Same with other reasons

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,104 @@ const (
...
+	HetznerBareMetalHostReadyV1Beta2Reason = "Ready"
+	// HetznerBareMetalHostNotReadyV1Beta2Reason indicates that the HetznerBareMetalHost is not ready.
+	HetznerBareMetalHostNotReadyV1Beta2Reason = "NotReady"
+	// HetznerBareMetalHostReadyUnknownV1Beta2Reason indicates that readiness of the HetznerBareMetalHost is unknown.
+	HetznerBareMetalHostReadyUnknownV1Beta2Reason = "ReadyUnknown"
+)
+
+const (
+	// HetznerBareMetalHostCredentialsAvailableV1Beta2Condition reports whether SSH credentials for the host are available.
+	HetznerBareMetalHostCredentialsAvailableV1Beta2Condition = "CredentialsAvailable"
```
janiskemper: should we maybe rename this to "SSHKeysAvailable"? We only use it for that AFAIK

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -455,9 +482,16 @@ func hetznerSecretErrorResult(
 			clusterv1beta1.ConditionSeverityError,
 			infrav1.ErrorMessageMissingHetznerSecret,
 		)
+		v1beta2conditions.Set(bmHost, metav1.Condition{
+			Type:    infrav1.HetznerBareMetalHostCredentialsAvailableV1Beta2Condition,
```
janiskemper: is this maybe the wrong condition? Shouldn't it be set on "RobotCredentialsAvailable"?

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -355,6 +371,60 @@ func (host *HetznerBareMetalHost) SetConditions(conditions clusterv1beta1.Condit
...
+
+// HetznerBareMetalHostV1Beta2SummaryOpts returns the v1beta2 summary options for a HetznerBareMetalHost.
+// It is the single source of truth for which conditions contribute to the Ready summary,
+// used both by the scope's Close() and by early-exit error paths that bypass the scope.
+func HetznerBareMetalHostV1Beta2SummaryOpts() []v1beta2conditions.SummaryOption {
+	return []v1beta2conditions.SummaryOption{
+		v1beta2conditions.ForConditionTypes{
+			HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Condition,
+			HetznerBareMetalHostRobotRateLimitExceededV1Beta2Condition,
+			HetznerBareMetalHostDeletingV1Beta2Condition,
```
janiskemper: I'm not sure if this order is correct. If the HetznerBareMetalHost is deleting, then it might be part of a cluster - so it has to deprovision. 

Therefore, if there are problems with deprovisioning, we need to show these with higher priority than the "deleting". 

Therefore, AFAIK  the credentials available e.g. might have to come first. 

I didn't 100% double check, but I believe this is correct.

You can double check the whole order again.

api/v1beta1/conditions_const.go
janiskemper: I realized that we have currently a HostReadyCondition. We set that on false on deletion. Otherwise we use it only for the hetznerbaremetalcontroller. 

I think that we should add to the issue that we replace the HostReadyCondition set on false on deletion by adding another deletion condition. This makes it transparent.

api/v1beta1/conditions_const.go
```diff
@@ -290,3 +290,104 @@ const (
 	// BootIDEmptyReason indicates that an empty boot ID is present on the node object.
 	BootIDEmptyReason = "BootIDEmpty"
 )
+
+// v1beta2 conditions for HetznerBareMetalHost.
```
janiskemper: this PR is not up to date. There were multiple new conditions / reasons added in the actionProvisioned function.

janiskemper (changes requested):

api/v1beta1/conditions_const.go
```diff
@@ -295,3 +295,113 @@ const (
...
+	// HetznerBareMetalHostValidationFailedV1Beta2Reason indicates the specified root device hints could not be validated.
+	HetznerBareMetalHostValidationFailedV1Beta2Reason = "ValidationFailed"
+)
+
+const (
+	// HetznerBareMetalHostProvisionSucceededV1Beta2Condition reports whether the host has been provisioned.
+	HetznerBareMetalHostProvisionSucceededV1Beta2Condition = "ProvisionSucceeded"
+	// HetznerBareMetalHostProvisionSucceededV1Beta2Reason indicates the host has been provisioned.
+	HetznerBareMetalHostProvisionSucceededV1Beta2Reason = "Provisioned"
+	// HetznerBareMetalHostStillProvisioningV1Beta2Reason indicates the server is still provisioning.
```
janiskemper: we can remove "still"

api/v1beta1/conditions_const.go
```diff
@@ -295,3 +295,113 @@ const (
...
+	// HetznerBareMetalHostServerHasNoIPv4V1Beta2Reason indicates a bare metal server has no IPv4 address.
+	HetznerBareMetalHostServerHasNoIPv4V1Beta2Reason = "ServerHasNoIPv4"
+	// HetznerBareMetalHostLinuxOnOtherDiskFoundV1Beta2Reason indicates the server cannot be provisioned on the given WWN.
+	HetznerBareMetalHostLinuxOnOtherDiskFoundV1Beta2Reason = "LinuxOnOtherDiskFound"
+	// HetznerBareMetalHostWipingDiskFailedV1Beta2Reason indicates erasing the disks before provisioning failed.
+	HetznerBareMetalHostWipingDiskFailedV1Beta2Reason = "WipingDiskFailed"
+	// HetznerBareMetalHostSSHToRescueSystemFailedV1Beta2Reason indicates the rescue system cannot be reached via SSH.
+	HetznerBareMetalHostSSHToRescueSystemFailedV1Beta2Reason = "SSHToRescueSystemFailed"
+	// HetznerBareMetalHostRebootTimedOutV1Beta2Reason indicates the reboot timed out.
+	HetznerBareMetalHostRebootTimedOutV1Beta2Reason = "RebootTimedOutReached"
```
janiskemper: "RebootTimeOutReached" or "RebootTimedOut". Which one is aligned with the rest?

api/v1beta1/conditions_const.go
```diff
@@ -295,3 +295,113 @@ const (
...
+	HetznerBareMetalHostBootIDEmptyV1Beta2Reason = "BootIDEmpty"
+)
+
+const (
+	// HetznerBareMetalHostRebootSucceededV1Beta2Condition reports whether the most recent reboot of the host succeeded.
+	HetznerBareMetalHostRebootSucceededV1Beta2Condition = "RebootSucceeded"
+	// HetznerBareMetalHostRebootSucceededV1Beta2Reason indicates the most recent reboot succeeded.
+	HetznerBareMetalHostRebootSucceededV1Beta2Reason = "Succeeded"
+	// HetznerBareMetalHostWaitingForNodeToBeRebootedV1Beta2Reason indicates the controller is waiting for the node to come back after a reboot.
+	HetznerBareMetalHostWaitingForNodeToBeRebootedV1Beta2Reason = "WaitingForNodeToBeRebooted"
```
janiskemper: Should we just name this "rebooting"? I think we can replace the whole string pretty much

janiskemper (changes requested):

api/v1beta1/conditions_const.go
abdullah599: RebootTimeoutReached. That's what I intended but made a copy paste error :,)

ref: https://github.com/syself/cluster-api-provider-hetzner/pull/1935#discussion_r3274273221

janiskemper (dismissed):

janiskemper (approved):

PR #2021, opened 2026-05-19
:seedling: In the final phase caph did way too many reconciles.

guettli:
In steady state (no reboot requested, boot ID already captured), `actionProvisioned` was calling into the workload cluster on every reconcile and then requeuing after 5 minutes — unnecessary work.

**Fix:** short-circuit early with the new `actionFinished` type, which returns an empty `reconcile.Result{}` and lets watch triggers drive the next reconcile instead of polling.

- Add `actionFinished` to `action_result.go` (no requeue, distinct from `actionStop` which signals a permanent error)
- Fast-path in `actionProvisioned`: if no reboot is desired and `NodeBootID` is already set, return `actionFinished{}` without touching the workload cluster
- Test that the fast path returns `actionFinished`

### DIFF
--- a/pkg/services/baremetal/host/action_result.go
+++ b/pkg/services/baremetal/host/action_result.go
@@ -52,13 +52,23 @@ func (r actionContinue) Result() (reconcile.Result, error) {
 	return reconcile.Result{RequeueAfter: time.Second}, nil
 }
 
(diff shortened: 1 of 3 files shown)


janiskemper (commented):
LGTM

Dhairya-Arora01 (approved):
LGTM

PR #2027, opened 2026-05-22
:seedling: Remove early return when reboot is not requested

Dhairya-Arora01:
Remove the early return when reboot is not requested because a server might be reboot due to some other reason than the presence of reboot annotation. And in that case, the new boot id won't be set on the host object. This creates a mismatch between the actual boot id on the node and the host object.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper (commented):
isn't something off if we never use "actionFinished" anymore? Then we would end up in the loop again no?

janiskemper (approved):

PR #2030, opened 2026-05-25
🌱 Introduce v1beta2 conditions for HCloudMachineTemplate

### LINKED ISSUE #2028: Introduce v1beta2 conditions for HCloudMachineTemplate
## Summary

Add v1beta2 conditions to `HCloudMachineTemplate` while keeping the existing v1beta1 conditions unchanged.

## What would you like to add

`HCloudMachineTemplate` is a CAPI infrastructure template object. Unlike `HCloudMachine`, it does not manage a Hetzner Cloud server. The controller validates the Hetzner API token and queries `ListServerTypes` once to compute and publish `status.capacity`, which is the value the cluster autoscaler reads when it scales a node group from zero.

Today its lifecycle is surfaced through two v1beta1 conditions:

| Condition | Purpose |
| --- | --- |
| `HCloudTokenAvailable` | Hetzner API credentials are valid and reachable |
| `RateLimitExceeded` | The controller is currently throttled by the HCloud API |

Both conditions are still needed and must keep their current behavior. At the same time, the CAPI v1beta2 transition requires `HCloudMachineTemplate` to expose v1beta2 conditions in parallel.

`HCloudMachineTemplate` has no creation or deletion lifecycle on the Hetzner side. The controller does not allocate or release HCloud resources for the template itself, so a dedicated v1beta2 `Deleting` condition is not required for this resource.

The shared rate-limit helper `pkg/services/hcloud/util.HandleRateLimitExceeded` already sets the v1beta2 `HCloudRateLimitExceeded` condition for any object passed to it, so the rate-limit path is already partially wired and just needs to be declared as an owned condition on `HCloudMachineTemplate`. 

## The Solution

We follow the same approach used for `HCloudMachine` in #1934, scoped down to what this resource actually does.

The existing v1beta1 conditions stay as they are. Alongside them, we add a v1beta2 conditions slice to `HCloudMachineTemplate` and set the two v1beta2 conditions (`HCloudTokenAvailable` and `HCloudRateLimitExceeded`) from the same places that already set the v1beta1 ones. The shared rate-limit helper already writes the v1beta2 rate-limit condition for any object passed to it, so most of the actual wiring work here is on the token condition.

The controller has a few branches where it returns early without setting any input condition: when the template is owned by a `ClusterClass`, when the owner cluster is not found, or when the owner cluster has no `infrastructureRef`. To keep these states visible on the v1beta2 status, we add a third v1beta2 condition, `Available`, which the controller sets on every path including the early returns:

| Status | Reason | When |
| --- | --- | --- |
| `True` | `Available` | Normal reconcile completed |
| `True` | `OwnedByClusterClass` | Template is managed by a `ClusterClass` (by design, no action needed) |
| `Unknown` | `WaitingForOwnerCluster` | Standalone template, owner cluster lookup failed (usually transient) |
| `False` | `MissingInfrastructureRef` | Standalone template, owner cluster is missing `infrastructureRef` |

Once these conditions are in place, the controller rolls them up into a v1beta2 `Ready` summary.

abdullah599:
Introduced v1beta2 conditions for HCloudMachineTemplate

Fixes #2028

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


controllers/hcloudmachinetemplate_controller.go
```diff
@@ -88,7 +90,20 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
...
+		if err := scope.SetHCloudMachineTemplateV1Beta2SummaryCondition(machineTemplate); err != nil {
+			log.Error(err, "Failed to set v1beta2 Ready condition")
+			v1beta2conditions.Set(machineTemplate, metav1.Condition{
+				Type:   clusterv1beta1.ReadyV1Beta2Condition,
+				Status: metav1.ConditionUnknown,
+				Reason: infrav1.InternalErrorV1Beta2Reason,
+			})
+		}
+
+		if err := patchHelper.Patch(ctx, machineTemplate, scope.MachineTemplatePatchOpts()...); err != nil {
```
janiskemper, line 106: don't we patch twice? One here one in the Close function?

janiskemper (commented):
From my side it's okay. @Dhairya-Arora01 plz review the details

controllers/hcloudmachinetemplate_controller.go
abdullah599: In the current controller flow, `HCloudMachineTemplateScope.Close(ctx)` is not called, so we are not actually patching twice. The PR adds the controller-level deferred patch because we need to persist v1beta2 conditions even on early-return paths before the scope is created.

That said, the existing scope patch path is unused for `HCloudMachineTemplate` and makes this confusing. Do you want me to include a cleanup in this PR that removes the unused scope patch helper / `Close` / `PatchObject`?

janiskemper: yes sounds good

abdullah599: Decided in chat: We will remove the unused patch helpers `Close` / `PatchObject`.

controllers/hcloudmachinetemplate_controller.go
```diff
@@ -139,27 +193,44 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
...
+		v1beta2conditions.Set(machineTemplate, metav1.Condition{
+			Type:    infrav1.HCloudMachineTemplateAvailableV1Beta2Condition,
+			Status:  metav1.ConditionUnknown,
+			Reason:  infrav1.InternalErrorV1Beta2Reason,
+			Message: err.Error(),
+		})
+		return reconcile.Result{}, err
 	}
 
 	// Always close the scope when exiting this function so we can persist any HCloudMachine changes.
```
Dhairya-Arora01: we can either remove or re-write this comment

Dhairya-Arora01 (dismissed):

Dhairya-Arora01 (approved):

PR #2035, opened 2026-05-26
:seedling: Preserve original reason/message when handling boot state times out

### LINKED ISSUE #2034: Preserve original error reason/message when boot state initialization times out
When an `HCloudMachine` times out during provisioning, the condition's reason and message are overwritten with the timeout information, hiding the actual root cause.

For example, a machine with an invalid server type ends up with:
```yaml
reason: HandleBootStateUnsetTimedOut
message: handleBootStateUnset timed out after 1h3m6s. Deleting machine
```

The original error (e.g. "failed to create server" due to a non-existent server type) is lost entirely.

**Proposal**:
Preserve the original reason, do not overwrite the condition's reason with a timeout-specific value. The pre-timeout reason (e.g. something indicating a server creation failure) should be kept as it is.

But include timeout context in message appended to the original error rather than replacing it. For example:
`message: "<original error message> (boot state initialization timed out after 1h3m6s)"`

Dhairya-Arora01:
When an `HCloudMachine` times out during provisioning, the condition;s reason and message should not be overwritten but instead the original reason and message should be preserved. The information about timeout can be appended to the original message rethar than replacing it.

Fixes #2034

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -225,21 +225,39 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 	durationOfState := time.Since(hm.Status.BootStateSince.Time)
 	if durationOfState > 6*time.Minute {
 		// timeout. Something has failed.
-		msg := fmt.Sprintf("handleBootStateUnset timed out after %s. Deleting machine",
-			durationOfState.Round(time.Second).String())
-		err := s.scope.SetErrorAndRemediate(ctx, msg)
+		timeoutMsg := fmt.Sprintf("boot state unset timed out, in this state since %s", durationOfState.Round(time.Second).String())
+
+		v1beta1Reason := "HandleBootStateUnsetTimedOut"
+		v1beta1Msg := timeoutMsg
+		if existing := v1beta1conditions.Get(hm, infrav1.ServerCreateSucceededCondition); existing != nil {
+			v1beta1Reason = existing.Reason
+			if existing.Message != "" {
+				v1beta1Msg = fmt.Sprintf("%s (%s)", existing.Message, timeoutMsg)
+			}
+		}
+
+		v1beta2Reason := infrav1.HCloudMachineBootStateUnsetTimedOutV1Beta2Reason
+		v1beta2Msg := timeoutMsg
+		if existing := v1beta2conditions.Get(hm, infrav1.HCloudMachineServerCreatedV1Beta2Condition); existing != nil {
+			v1beta2Reason = existing.Reason
+			if existing.Message != "" {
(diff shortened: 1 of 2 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -225,20 +225,34 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
-			durationOfState.Round(time.Second).String())
+		timeoutMsg := fmt.Sprintf("boot state initialization timed out, in this state since %s", durationOfState.Round(time.Second).String())
+
+		v1beta1Reason := "HandleBootStateUnsetTimedOut"
+		if existing := v1beta1conditions.Get(hm, infrav1.ServerCreateSucceededCondition); existing != nil {
+			v1beta1Reason = existing.Reason
+		}
+
+		v1beta2Reason := infrav1.HCloudMachineBootStateInitializingTimedOutV1Beta2Reason
+		msg := timeoutMsg
```
janiskemper: this looks as if it is part of v1beta2. Can you maybe organize the code differently? It seems as if you don't change the message for v1beta1.. Why?

pkg/services/hcloud/server/server.go
```diff
@@ -225,21 +225,39 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 	durationOfState := time.Since(hm.Status.BootStateSince.Time)
 	if durationOfState > 6*time.Minute {
 		// timeout. Something has failed.
-		msg := fmt.Sprintf("handleBootStateUnset timed out after %s. Deleting machine",
-			durationOfState.Round(time.Second).String())
-		err := s.scope.SetErrorAndRemediate(ctx, msg)
+		timeoutMsg := fmt.Sprintf("boot state initialization timed out, in this state since %s", durationOfState.Round(time.Second).String())
```
janiskemper: I think this is the wrong message / state

pkg/services/hcloud/server/server.go
```diff
@@ -225,21 +225,39 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
+		v1beta1Reason := "HandleBootStateUnsetTimedOut"
+		v1beta1Msg := timeoutMsg
+		if existing := v1beta1conditions.Get(hm, infrav1.ServerCreateSucceededCondition); existing != nil {
+			v1beta1Reason = existing.Reason
+			if existing.Message != "" {
+				v1beta1Msg = fmt.Sprintf("%s (%s)", existing.Message, timeoutMsg)
+			}
+		}
+
+		v1beta2Reason := infrav1.HCloudMachineBootStateInitializingTimedOutV1Beta2Reason
```
janiskemper: this also

janiskemper (approved):

PR #2036, opened 2026-05-26
:seedling: fix flaky test via RWMutex

guettli:
### Problem

The test `HetznerBareMetalRemediationReconciler / should not remediate if no annotations is present` was failing intermittently with:

```
assert: mock: I don't know what to return because the method call was unexpected.
    Either do Mock.On("GetHostName").Return(...) first, or remove the GetHostName() call.
```

The panic originated in a **controller-runtime reconcile goroutine** from a previous test. Ginkgo cannot recover panics that escape into goroutines it does not own, so the failure was attributed to the next test's `[BeforeEach]`.

### Root cause

`ResetAndInitNamespace` replaced the `SSHClientFactory` pointer on the reconciler with a brand-new factory containing unconfigured mocks. An in-flight goroutine from the previous test could read the new factory and call `GetHostName` before any `On()` expectations were registered, causing a testify panic.

### Fix

`SSHFactory` (in `pkg/services/baremetal/client/mocks/factory.go`) is now a stable object created once in `NewControllerResetter` and wired into the reconcilers once. It holds its mock clients behind a `sync.RWMutex`. `ResetAndInitNamespace` calls `SetClients` to swap in fresh mocks per test.

Because `SetClients` holds a write lock and `NewClient` holds a read lock, any goroutine that calls `NewClient` sees either the fully-configured old clients or the new ones — never an uninitialised mock. The race window is closed entirely.

Fail: https://github.com/syself/cluster-api-provider-hetzner/actions/runs/26452021030/job/77875240925#step:8:1356


controllers/controllers_suite_test.go
```diff
@@ -157,6 +157,13 @@ func (r *ControllerResetter) ResetAndInitNamespace(namespace string, testEnv *he
 	r.HCloudMachineTemplateReconciler.HCloudClientFactory = hcloudClientFactory
 	r.HCloudMachineTemplateReconciler.Namespace = namespace
 
+	// Register base SSH expectations before wiring the factory into the reconciler.
+	// Any reconcile goroutine from a previous test that is still in-flight may read
+	// the new factory as soon as it is set below. Without expectations already in
+	// place, those goroutines trigger an unexpected-call panic that Ginkgo attributes
+	// to the next test's BeforeEach, causing flaky failures.
+	configureRescueSSHClient(rescueSSHClient)
```
Dhairya-Arora01: Im unsure if we should register the mock client expectations here. function `ResetAndInitNamespace` is used by `testenv.ResetAndCreateNamespace` which is used in every controller test.
Setting the mock expecations here means they will be available in every reconciler test , doesn't matter if we intend to use the mock there or not.

guettli: @Dhairya-Arora01 to be honest, I don't like the PR. I like the result: No more flaky tests here. But it looks complicated. I think about a different way. Maybe we can stop all Reconciles in goroutines which are still running.... I will have a look.

guettli: @Dhairya-Arora01 please review again, I changed the implementation. PR is much easier to read now.

Dhairya-Arora01 (commented):
Que: In ResetAndInitNamespace, SetClients is called with unconfigured mocks. And the individual mock expectations (`On`) are registered later. So for a fraction of period the factory is serving mocks with zero expectations.

Now lets say if test N calls `NewClient` in that gap and then calls mock function, it would get test N+1 unconfigured mock. So I guess the same error can happen. WDYT ?

guettli:
> Que: In ResetAndInitNamespace, SetClients is called with unconfigured mocks. And the individual mock expectations (`On`) are registered later. So for a fraction of period the factory is serving mocks with zero expectations.
> 
> Now lets say if test N calls `NewClient` in that gap and then calls mock function, it would get test N+1 unconfigured mock. So I guess the same error can happen. WDYT ?

@Dhairya-Arora01 

Yes, you are right. I thought about that again.

I think the only solution is a RW-Mutex in Reconcile(): A running Reconcile() holds a read-lock.

Updating the mocks requires a Write-Lock.

This means updating the Mocks will be safe: It will wait until all Reconcile calls are finished.

We need to update production code for that, but we already do that for checking the Namespace.

I found no other solution which really fixes the root-cause. Do you have a better idea?

I will update the PR.

alternatives.md
```diff
@@ -0,0 +1,260 @@
+# Alternative solutions to the mock/reconcile race
```
Dhairya-Arora01: This file should be removed.

plan.md
```diff
@@ -0,0 +1,202 @@
+# Plan: Fix mock/reconcile race with a RW-mutex gate
```
Dhairya-Arora01: This file should be removed as well.

controllers/controllers_suite_test.go
```diff
@@ -78,6 +78,8 @@ func TestControllers(t *testing.T) {
 
 type ControllerResetter struct {
 	debug                                 bool
+	gate                                  sync.RWMutex
```
Dhairya-Arora01: can you add a comment about this gate

Dhairya-Arora01 (commented):
@janiskemper , please take a look

alternatives.md
guettli: done.

controllers/controllers_suite_test.go
guettli: @Dhairya-Arora01 I added a comment, and renamed it. So that both parts use the same variable name.

controllers/controllers_suite_test.go
```diff
@@ -77,7 +77,13 @@ func TestControllers(t *testing.T) {
 }
 
 type ControllerResetter struct {
-	debug                                 bool
+	debug bool
+	// reconcileGate serializes mock-client setup against in-flight reconciles.
```
janiskemper: I don't understand this comment at all. Can you plz explain in different words what reconcileGate does?

controllers/controllers_suite_test.go
```diff
@@ -96,7 +102,15 @@ func NewControllerResetter(
 	hcloudRemediationReconciler *HCloudRemediationReconciler,
 	hetznerBareMetalRemediationReconciler *HetznerBareMetalRemediationReconciler,
 ) *ControllerResetter {
-	return &ControllerResetter{
+	// Create the stable SSH factory once and wire it into the reconcilers that need it.
```
janiskemper: this comment looks like a description for this PR, not a comment that someone understand when looking at the code in 6 months. Please update it in a way that makes it valuable for the code.

controllers/controllers_suite_test.go
```diff
@@ -106,13 +120,26 @@ func NewControllerResetter(
 		HetznerBareMetalRemediationReconciler: hetznerBareMetalRemediationReconciler,
 		debug:                                 os.Getenv("DEBUG") != "",
 	}
+
+	// Wire the reconcile gate so that mock setup (write lock) is serialized with
```
janiskemper: I don't understand this.

controllers/controllers_suite_test.go
```diff
@@ -173,6 +195,14 @@ func (r *ControllerResetter) ResetAndInitNamespace(namespace string, testEnv *he
 	if r.debug {
 		testEnv.GetLogger().Info("Starting test: ===> ===> ===> ===> ===> ===> ===> " + t.Name())
 	}
+
+	var once sync.Once
```
janiskemper: can you add a comment what this does?

controllers/hcloudmachine_controller.go
```diff
@@ -69,6 +70,11 @@ type HCloudMachineReconciler struct {
 
 	// Reconcile only this namespace. Only needed for testing
 	Namespace string
+
+	// ReconcileGate, when non-nil, is held as a read lock for the full duration of each
```
janiskemper: I don't understand the part about read and write lock

janiskemper (changes requested):

controllers/controllers_suite_test.go
guettli: I updated the comment:

```go
	// The reconcileGate ensures that mock setup is serialized with in-flight reconciles. The
	// mock setup waits via a write-lock until all reconciles are finished (by releasing the
	// read-lock).
```

is it now easier to understand?

controllers/controllers_suite_test.go
guettli: it was not needed - removed.

controllers/hcloudmachine_controller.go
guettli: I added:

> Before setting up mocks (in tests only), we want to ensure that all Reconcile() calls are finished.

Better now?

controllers/controllers_suite_test.go
janiskemper: > mock setup is serialized with in-flight reconciles

what does this mean? 

(I'll ask about the read and write lock below)

controllers/hcloudmachine_controller.go
janiskemper: So what does the write lock and what does the read lock do exactly? I understand the general effort with a locking mechanism, but not really the read and write lock details...

controllers/controllers_suite_test.go
guettli: @janiskemper new comment:

```go 
	// reconcileGate serializes mock-client setup against in-flight reconciles.
	// ResetAndInitNamespace holds the write lock while swapping clients/mocks.
	// HetznerBareMetalHostReconciler and HCloudMachineReconciler hold the read lock during each
	// Reconcile call via their ReconcileGate pointer. This ensures no Reconcile is active while
	// clients/mocks are being swapped between tests. Without this, tests can fail randomly
	// because an in-flight Reconcile may observe inconsistent state during the swap.
```

better now?

controllers/controllers_suite_test.go
guettli: @janiskemper I shortened it to:

```go 
// One factory shared across resets so in-flight goroutines always hold a valid pointer.
```

Is that your intention?

controllers/controllers_suite_test.go
guettli: @janiskemper in-flight means here "currently running".

I elaborated the docs for `ReconcileGate`. I could copy+past that text and add it here, too. But somehow this feels strange. What is your suggestion for that?

controllers/hcloudmachine_controller.go
guettli: Does the new docstring for ReconcileGate answer your question?

pkg/services/baremetal/client/mocks/factory.go
```diff
@@ -18,36 +18,60 @@ limitations under the License.
...
-		osClientAfterCloudInit:    osClientAfterCloudInit,
-	}
+) *SSHFactory {
+	f := &SSHFactory{}
+	f.SetClients(rescueClient, osClientAfterInstallImage, osClientAfterCloudInit)
+	return f
 }
 
-var _ = sshclient.Factory(&sshFactory{})
+// SetClients atomically replaces the mock clients returned by NewClient. It is called by
```
janiskemper: where is the function "NewClient" that is mentioned here?

test/helpers/envtest.go
```diff
@@ -285,7 +290,15 @@ func (t *TestEnvironment) Cleanup(ctx context.Context, objs ...client.Object) er
...
 
 // ResetAndCreateNamespace creates a namespace.
-func (t *TestEnvironment) ResetAndCreateNamespace(ctx context.Context, generateName string) (*corev1.Namespace, error) {
+//
+// The second return value is a finish func that the caller must defer immediately:
+//
+//	testNs, finish, err := testEnv.ResetAndCreateNamespace(ctx, "my-reconciler")
+//	defer finish()
+//
+// Deferring finish() ensures that mock clients are activated (SetClients) and
```
janiskemper: can you explain what this finish function does exactly? It's called "ResetAndInitNamespace". 

Can you overall explain this ResetAndCreateNamespace function as well? It creates a namespace, and when that namespace exists already, it calls a function to reset it? 

If this function calls a reset, why do we defer the finish function? Before we called it directly, now we return it and defer it in the function where this one is used. 

This is extremely complicated. Why exactly do we do it this way? Is there no easier way? How can we explain why exactly it has to be done like this?

pkg/services/baremetal/client/mocks/factory.go
```diff
@@ -18,36 +18,60 @@ limitations under the License.
...
+	f.osClientAfterInstallImage = osAfterInstallImage
+	f.osClientAfterCloudInit = osAfterCloudInit
+}
 
-// NewClient implements the NewClient function of the SSHFactory interface.
-func (f *sshFactory) NewClient(in sshclient.Input) sshclient.Client {
-	// return rescueClient when private key rescue-private-key is given. Otherwise give the os private key.
+// NewClient implements sshclient.Factory.
+func (f *SSHFactory) NewClient(in sshclient.Input) sshclient.Client {
+	f.mu.RLock()
```
janiskemper, line 74: I see we call this lock here, but also in the reconcile functions of hcloudmachine and hetznerbaremetalhost. Why?

janiskemper (changes requested):
I still find this really hard to understand. Any idea you have how to make the code easier or how to explain it better, would be great!

controllers/controllers_suite_test.go
janiskemper: does the first line

> reconcileGate serializes mock-client setup against in-flight reconciles.

mean this:

> reconcileGate ensures no Reconcile is being done while mocks are being replaced between tests

?

Because the second one I understand better. If it is not this, then I didn't understand your comment in the first place.

guettli: Yes, that's exactly what it means. I replaced the old phrasing with your clearer version as the opening line of the comment.

pkg/services/baremetal/client/mocks/factory.go
guettli: `NewClient` is the method on the same type (`SSHFactory`) defined a few lines below in the same file. I updated the comment to say "NewClient (defined below)" to make this explicit. I also fixed a stale reference to `ControllerResetter.CommitMockSetup`, which no longer exists — the actual caller is the anonymous `finish()` func returned by `ResetAndInitNamespace`.

test/helpers/envtest.go
guettli: **What does `ResetAndCreateNamespace` do?**  
It always creates a *new* namespace (via `GenerateName`, so it never conflicts) and then calls `ResetAndInitNamespace` on the `Resetter`. The "reset" in the name refers to resetting the mock clients and reconciler state for the next test, not resetting an existing namespace.

**What does `finish()` do?**  
Two things, in order:

1. Calls `SetClients` — installs the mock clients that the test has just configured with `On()` expectations into the shared `SSHFactory`.
2. Calls `reconcileGate.Unlock()` — releases the write lock, allowing new `Reconcile` calls to proceed.

**Why defer it instead of calling directly?**  
`ResetAndInitNamespace` acquires the write lock immediately on entry. The test's `BeforeEach` then registers its `On()` expectations on the mock clients — *while the write lock is still held*, preventing any `Reconcile` from starting with half-configured mocks. `defer finish()` runs at the end of `BeforeEach`, after all expectations are registered. Only then are the mocks installed and the gate released.

If `finish()` were called at the end of `ResetAndCreateNamespace` instead, the gate would be released before the test has registered any expectations, and a `Reconcile` goroutine could sneak in and call `NewClient` before `SetClients` has run.

I updated the doc comment on `ResetAndCreateNamespace` to explain this flow directly.

pkg/services/baremetal/client/mocks/factory.go
guettli: Both locks were introduced by this PR, in separate commits. They solve two different problems:

**`f.mu` in `SSHFactory`** — a low-level data-race guard on the struct fields. `SetClients` writes the three client fields; `NewClient` reads them. Without `f.mu`, two goroutines calling `NewClient` concurrently with a `SetClients` call would be a data race. The read lock lets multiple `NewClient` callers run in parallel; `SetClients` takes the write lock for exclusive field access.

**`ReconcileGate` in the reconcilers** — a higher-level coordination mechanism. It does not protect any data field; it ensures that an entire `Reconcile` call cannot run while a test is setting up `On()` expectations. The test holds the write lock for the whole `BeforeEach` block. Without it, a reconcile goroutine could start at the old mocks, partially execute, then see the new mocks mid-run — causing flaky failures that `f.mu` alone cannot prevent.

In short: `f.mu` prevents a data race on individual field reads/writes; `ReconcileGate` prevents a whole Reconcile from observing inconsistent mock state across a test boundary.

controllers/controllers_suite_test.go
```diff
@@ -98,7 +104,13 @@ func NewControllerResetter(
 	hcloudRemediationReconciler *HCloudRemediationReconciler,
 	hetznerBareMetalRemediationReconciler *HetznerBareMetalRemediationReconciler,
 ) *ControllerResetter {
-	return &ControllerResetter{
+	// One factory shared across resets so in-flight goroutines always hold a valid pointer.
+	f := &mocks.SSHFactory{}
+	hcloudMachineReconciler.SSHClientFactory = f
```
janiskemper: why don't you do this outside of this function? Right now you first define the hcloudMachineReconciler outside of this function and then you overwrite it partially. This might be very confusing

controllers/controllers_suite_test.go
```diff
@@ -108,13 +120,27 @@ func NewControllerResetter(
 		HetznerBareMetalRemediationReconciler: hetznerBareMetalRemediationReconciler,
 		debug:                                 os.Getenv("DEBUG") != "",
 	}
+
+	// Give both reconcilers access to the shared gate. Each Reconcile call holds it as a read lock,
+	// and test setup in ResetAndInitNamespace holds it as a write lock — which blocks until all
+	// in-flight reconciles (read lock holders) finish before mock clients are swapped.
+	hetznerBareMetalHostReconciler.ReconcileGate = &r.reconcileGate
```
janiskemper: where exactly is this mutex defined? I don't see this code.. Am I blind?
In any way, also here, I would prefer us not overwriting the reconcilers from inside this function

test/helpers/envtest.go
```diff
@@ -127,7 +127,12 @@ type Resetter interface {
 	// testEnv: the TestEnvironment which should be resetted.
 	//
 	// t: g.GinkgoT()
-	ResetAndInitNamespace(namespace string, testEnv *TestEnvironment, t g.FullGinkgoTInterface)
+	//
+	// The returned func must be deferred by the caller immediately after
```
janiskemper: And which effect does it have to return it? At which point should it be executed? At the end of a test setup, right before the actual tests run? Why is it crucial to have this function and to defer it?

janiskemper (dismissed):

controllers/controllers_suite_test.go
guettli: I moved it outside. See: f62d71b3a

test/helpers/envtest.go
guettli: @janiskemper I improved the docs: 176ed381888ff7e0f3658335bf454758eea0c450

controllers/controllers_suite_test.go
guettli: I moved that outside.  See: 54a92c80b

janiskemper (commented):
LGTM

janiskemper:
@Dhairya-Arora01 plz check again

Dhairya-Arora01 (approved):

PR #2039, opened 2026-05-27
🐛 Add cluster pause check to HBMH controllers

### LINKED ISSUE #2037: Add cluster pause check to HBMH controllers
While investigating cluster pause behavior in CAPH, we found that the HBMH controller doesn't check if the cluster is paused or the pause annotation is present on the object before reconciling.

Most other CAPH controllers already perform this check using util func from `sigs.k8s.io/cluster-api/util/annotations`
```go
if annotations.IsPaused(cluster, hcloudMachine) {
		log.Info("HCloudMachine or linked Cluster is marked as paused. Won't reconcile")
}
```

- [hcloud_machine_controller](https://github.com/syself/cluster-api-provider-hetzner/blob/e663e5ea4e7aad22f9dda22065660233818d6718/controllers/hcloudmachine_controller.go#L101-L104)
- [hcloud_remediation_controller](https://github.com/syself/cluster-api-provider-hetzner/blob/e663e5ea4e7aad22f9dda22065660233818d6718/controllers/hcloudremediation_controller.go#L110-L113)
- [hetzner_baremetal_machine_controller](https://github.com/syself/cluster-api-provider-hetzner/blob/e663e5ea4e7aad22f9dda22065660233818d6718/controllers/hetznerbaremetalmachine_controller.go#L95-L98)
- [hetzner_baremetal_remediation_controller](https://github.com/syself/cluster-api-provider-hetzner/blob/e663e5ea4e7aad22f9dda22065660233818d6718/controllers/hetznerbaremetalremediation_controller.go#L103-L106)
- [hetzner_cluster_controller](https://github.com/syself/cluster-api-provider-hetzner/blob/e663e5ea4e7aad22f9dda22065660233818d6718/controllers/hetznercluster_controller.go#L120-L123)

The **goal** of this issue is to implement the pause check for HBMH.

The CSR controller is intentionally out of scope. It runs on the workload cluster's manager while the `Cluster` object lives on the management cluster, so honoring pause there would need an extra cluster fetch plus periodic requeue (there is no unpause event to react to). The CSR approval flow is also expected to move out of CAPH into a separate controller in the mid term.

abdullah599:

Apply the standard `annotations.IsPaused(cluster, obj)` pattern to the HBMH controller.

Fixes #2037

None

**TODOs**:
- [ ] squash commits
- [ ] add unit tests


controllers/csr_controller.go
```diff
@@ -97,6 +98,25 @@ func (r *GuestCSRReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
+	// Fetch the Cluster from the management cluster.
+	cluster := &clusterv1.Cluster{}
+	clusterName := types.NamespacedName{
+		Namespace: r.mCluster.Namespace(),
+		Name:      r.clusterName,
+	}
+	if err := r.mCluster.Get(ctx, clusterName, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Cluster is not available yet")
+			return reconcile.Result{}, nil
```
Dhairya-Arora01: I think we either need to add reque in such a case.

controllers/csr_controller_test.go
```diff
@@ -17,11 +17,184 @@ limitations under the License.
...
+			result, err := reconciler.Reconcile(context.Background(), reconcile.Request{
+				NamespacedName: types.NamespacedName{Name: certificateSigningRequest.Name},
+			})
+			if err != nil {
+				t.Fatalf("Reconcile() error = %v, want nil", err)
+			}
+			if result != (reconcile.Result{}) {
+				t.Fatalf("Reconcile() result = %v, want empty result", result)
+			}
+
```
Dhairya-Arora01: can you add maybe a comment here that by verifying the reconcile result and that no conditions are set on CSR, we can be sure that it was not reconciled due to pause annotation.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -248,6 +255,11 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	log = log.WithValues("Cluster", klog.KObj(cluster))
 
+	if annotations.IsPaused(cluster, bmHost) {
+		log.Info("HetznerBareMetalHost or linked Cluster is marked as paused. Won't reconcile")
```
Dhairya-Arora01, line 257: At this point, we are sure that only linked Cluster will be paused as we already checked if HBMH is paused earlier.

Dhairya-Arora01 (dismissed):
LGTM, @janiskemper plz take a look

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -248,6 +255,13 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	log = log.WithValues("Cluster", klog.KObj(cluster))
 
+	// Host-level pause was already handled earlier via annotations.HasPaused(bmHost),
```
janiskemper: the pause would always be on cluster-level no?

abdullah599: The issue (#2037) explicitly asks to honor both cluster.Spec.Paused and the cluster.x-k8s.io/paused annotation on the object. All other CAPH controllers (hcloudmachine, hetznerbaremetalmachine, both remediation controllers, hetznercluster) use annotations.IsPaused(cluster, obj), so dropping object-level support here would be the odd one out imo


cc: @Dhairya-Arora01

janiskemper: wait @abdullah599 I think you understood it the wrong way around. The previous check for the annotation on the hbmh object is the one I don't understand

abdullah599: Ah, got it

The early `annotations.HasPaused(bmHost)` check is there because `reconcileSelectedStates` runs between the two pause checks, and it is not a no-op. For a non-deleted host in `StateNone`, if `NeedsProvisioning()` returns true, it moves the host to `StatePreparing` and sets a v1beta2 condition. Without the early check, a paused host would still get its state changed before the later `annotations.IsPaused(cluster, bmHost)` runs.

We cannot move the cluster lookup above `reconcileSelectedStates`. The `StateDeleting` branch needs to remove the finalizer even when the `HetznerCluster` is already gone, and that branch runs inside `reconcileSelectedStates`.

abdullah599: Update: Removed the early `HasPaused(bmHost)` check (and the test that asserted it). `MaintenanceMode` already covers the "don't touch this host" case, and setting a paused annotation directly on an HBMH is not a real workflow.

controllers/csr_controller.go
```diff
@@ -97,6 +98,25 @@ func (r *GuestCSRReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
+	// Fetch the Cluster from the management cluster.
+	cluster := &clusterv1.Cluster{}
+	clusterName := types.NamespacedName{
+		Namespace: r.mCluster.Namespace(),
+		Name:      r.clusterName,
+	}
+	if err := r.mCluster.Get(ctx, clusterName, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Cluster is not available yet, requeueing")
+			return reconcile.Result{Requeue: true}, nil
```
janiskemper: we should use RequeueAfter everywhere..

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -224,16 +225,17 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 	// Case "Delete" was handled in reconcileSelectedStates. From now we know that the host has no
 	// DeletionTimestamp set. But the hbmm could be in Deprovisioning.
 
+	if bmHost.Spec.Status.HetznerClusterRef == "" {
+		log.Info("bmHost.Spec.Status.HetznerClusterRef is empty. Looks like a stale cache read")
+		return reconcile.Result{Requeue: true}, nil
```
janiskemper: same here

controllers/csr_controller_test.go
janiskemper: can you  actually make this more understandable? You can ideally write above that you test that you test the early return with the paused state by ensuring that no conditions were set?

controllers/csr_controller_test.go
```diff
@@ -17,11 +17,186 @@ limitations under the License.
...
+				t.Fatalf("failed to get CSR: %v", err)
+			}
+			if len(updatedCSR.Status.Conditions) != 0 {
+				t.Fatalf("CSR conditions = %v, want no conditions", updatedCSR.Status.Conditions)
+			}
+		})
+	}
+}
+
+func TestGuestCSRReconciler_ReconcileContinuesWhenCSRAndClusterAreNotPaused(t *testing.T) {
```
janiskemper: here it is also not really clear what you test.. I didn't understand it on first glance

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -47,6 +53,114 @@ const (
...
+type countingRobotFactory struct {
+	calls int
+}
+
+func (f *countingRobotFactory) NewClient(robotclient.Credentials) robotclient.Client {
+	f.calls++
+	return nil
+}
+
+func TestHetznerBareMetalHostReconciler_ReconcileSkipsPausedCluster(t *testing.T) {
```
janiskemper, line 69: also add test descriptions here and below

janiskemper (changes requested):

controllers/csr_controller.go
```diff
@@ -97,6 +98,25 @@ func (r *GuestCSRReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
+	}
+	if err := r.mCluster.Get(ctx, clusterName, cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("Cluster is not available yet, requeueing")
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("failed to get Cluster: %w", err)
+	}
+
+	if annotations.IsPaused(cluster, certificateSigningRequest) {
```
janiskemper: I just thought about one thing: How will this be looking when the cluster comes back and is unpaused? Will this start reconciling again?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -248,6 +250,11 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 
 	log = log.WithValues("Cluster", klog.KObj(cluster))
 
+	if annotations.IsPaused(cluster, bmHost) {
```
janiskemper, line 256: same here, how will this start reconciling again?

controllers/csr_controller.go
abdullah599: You are right, nothing currently triggers reconcile after the cluster unpauses. The same gap exists for the Cluster NotFound case for which we used requeueAfter. If the management Cluster does not exist yet, nothing wakes the controller back up.

For HBMH the cleanest fix is to follow the hcloudmachine pattern and add Watches(Cluster) with the ClusterPausedTransitions predicate. The watch fires on Cluster creation and on pause transitions, so both the NotFound branch and the paused branch can return `reconcile.Result{}, nil` (We can also drop the RequeueAfter from the NotFound case in HBMH, since the watch covers it)

For CSR this pattern does not fit. CSR runs on the workload cluster's manager, and the Cluster object lives on the management cluster, so we cannot watch it from the same manager. RequeueAfter 1 minute on both the NotFound and the paused branches is the practical option there imo

WDYT?

abdullah599: Update: we will not implement pause check for CSR controller

janiskemper (approved):

PR #2044, opened 2026-05-28
:book: add hint about kubernetes version limitations for CAPH `v1.0.x`

skuethe:
We stumbled upon an error when upgrading to the newest patch versions of Kubernetes on versions `v1.33`, `v1.34`, `v1.35` as well as `v1.36` when using CAPH `v1.0.x` and wanted to update the docs on the supported version, so that others don't run into this issue.

There is a hint to this problem in the cluster-api book:  
https://github.com/kubernetes-sigs/cluster-api/blob/6c70a17c7f343442238fc695e66884e6c84cfa01/docs/book/src/user/troubleshooting.md#kubeadm-join-fails-after-upgrading-to-kubernetes-patch-releases

See also hints in the release notes

> Kubeadm: use a dedicated ClusterRole 'system:kubelet-api-admin' for the kube-apiserver kubelet client.

- [v1.33.12](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.33.md#changelog-since-v13311) 
- [v1.34.8](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.34.md#changelog-since-v1347)
- [v1.35.5](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.35.md#changelog-since-v1354)
- [v1.36.1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.36.md#changelog-since-v1360)

**TODOs**:

- [x] squash commits
- [x] include documentation
- [ ] add unit tests

### DIFF
--- a/docs/caph/01-getting-started/01-introduction.md
+++ b/docs/caph/01-getting-started/01-introduction.md
@@ -50,6 +50,8 @@ Related:
 
 Compatibility notes:
 
+- We recommend to use CAPH `v1.1.x` for Kubernetes versions `> 1.33.11`, `> 1.34.7`, `> 1.35.4` and `> 1.36.0`, as they [require a new ClusterRole `system:kubelet-api-admin`](https://github.com/kubernetes-sigs/cluster-api/blob/6c70a17c7f343442238fc695e66884e6c84cfa01/docs/book/src/user/troubleshooting.md#kubeadm-join-fails-after-upgrading-to-kubernetes-patch-releases) which is only created in CAPI versions `>= v1.11.11`, `>= v1.12.8` or `>= v1.13.2`.  
+If you want to use CAPH `v1.0.x` with these versions, you need to create the required `ClusterRoleBinding` yourself. See the link for more details.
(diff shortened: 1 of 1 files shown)


guettli (changes requested):
@skuethe I am unsure. Caph itself does not manage ClusterRoles. Neither in mgt-cluster nor in wl-cluster.

> CAPH `v1.0.x` does not support Kubernetes versions....

Afaik caph could support these versions, if the admin of the wl-cluster would somehow create the ClusterRole.

Maybe rephrase that to:

> We recommend to use CAPH `v1.1.x` for Kubernetes versions .... If you use CAPH `v1.0.x`, you need to  .... (current text).

@skuethe what do you think?

skuethe:
Hm, fair point. But then it probably also makes sense to remove the version limitations from the table and replace it with something else?  
Initially I just wanted to add a MD footnote to these entries, but I am unsure if your docs parser on syself.com/docs support these?  
ref: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes

guettli:
@skuethe I prefer the format of your inital PR. A footnote would make it less visible. Overall I think it is better to update to caph v1.1.x. I would like to merge your PR. Just a little update of the text is needed.
 
 Out of curiosity: Why do you not update to v1.1.x?

skuethe:
@guettli updated. I did remove the version limitations from the compatibility matrix to avoid confusion then.

Regarding our plans to update to `v1.1.x`: current time and resource limitations. We wanted to prioritise the k8s upgrades first and then switch afterwards :)

guettli (approved):

guettli:
@skuethe thank you for the PR!

PR #2046, opened 2026-05-28
:seedling: Break the loop to increment condition's observedGeneration

Dhairya-Arora01:
HetznerBareMetalHost stores its conditions inside Spec (not in a status subresource). The CAPI conditions library always sets ObservedGeneration = metadata.generation when writing a condition. Because the conditions are part of the spec, writing them bumps metadata.generation, which makes ObservedGeneration stale on the very next reconcile, which triggers another write, which bumps generation again, this is a loop that increments metadata.generation on every reconcile loop. To break the loop: if a condition's Status, Reason, and Message are unchanged from what is already stored, keep the existing ObservedGeneration instead of overwriting it with the current (higher) generation. ObservedGeneration will still advance whenever the condition itself actually changes.

Fixes #

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -381,6 +381,30 @@ func (host *HetznerBareMetalHost) GetV1Beta2Conditions() []metav1.Condition {
 
 // SetV1Beta2Conditions sets v1beta2 conditions for a HetznerBareMetalHost API object.
 func (host *HetznerBareMetalHost) SetV1Beta2Conditions(conditions []metav1.Condition) {
+	// HetznerBareMetalHost stores its conditions inside Spec (not in a status subresource).
+	// The CAPI conditions library always sets ObservedGeneration = metadata.generation when
+	// writing a condition. Because the conditions are part of the spec, writing them bumps
+	// metadata.generation, which makes ObservedGeneration stale on the very next reconcile,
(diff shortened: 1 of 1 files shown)


guettli (approved):

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -376,11 +376,41 @@ func (host *HetznerBareMetalHost) GetV1Beta2Conditions() []metav1.Condition {
 	if host.Spec.Status.V1Beta2 == nil {
 		return nil
 	}
-	return host.Spec.Status.V1Beta2.Conditions
+	// Return a deep copy so callers cannot modify the stored slice in-place via a shared
```
janiskemper, line 379: is this function only used on the one place? Because returning a copy might be problematic in other places in theory right?

Dhairya-Arora01: No it is used at other places as well, it is called via the v1beta2conditions.Setter inferface by the CAPI library.

But all the callers follow the same pattern i.e. read the slice and none of them mutate the returned slice and expect that it would persists without calling the setter. They always store by calling the SetV1Beta2Conditions. Therefore, retruning a copy is safe.

janiskemper (approved):

PR #2047, opened 2026-05-28
🌱 Move admission webhooks out of the api package

### LINKED ISSUE #2038: Move admission webhooks out of the api package

Depends on #2018, which creates `api/v1beta2` as a mechanical copy of `api/v1beta1`.

## Summary

Move CAPH's mutating and validating admission webhook code out of `api/` so the API packages contain API types, schema markers, generated code, and conversion methods only.

## The Problem

The current `api/v1beta1` package contains more than API types. It also contains mutating and validating admission webhook code:

* `*_webhook.go` files with `SetupWebhookWithManager`, `Default`, `ValidateCreate`, `ValidateUpdate`, and `ValidateDelete`.
* `webhooks.go`, which builds Kubernetes API validation errors for webhook responses.
* Some validation helpers that are only called from admission webhooks.

This mixes two concerns in one package. The API package should be safe to import when a consumer only needs CAPH types. Today it also pulls in controller runtime webhook code and manager setup code.

CAPI keeps this boundary cleaner. Its v1beta2 API package contains the API types and conversion code, while admission webhooks live outside the API package and register themselves against the API types. We decided we will follow CAPI's approach here.

## The Solution

Move the admission webhook implementation into an internal webhook package and keep the API packages focused on the API contract.

After this change:

* `api/v1beta1` and `api/v1beta2` do not contain admission webhook files.
* webhook structs and admission interfaces live outside `api/`.
* `main.go` registers webhooks through the new webhook package instead of calling methods on API objects.
* `controller-gen` still sees the kubebuilder webhook markers and generates the same webhook manifests.

Keep this as a move, not a behavior change. The existing defaulting and validation behavior should stay the same.

abdullah599:
Move admission webhooks out of the api package

Fixes #2038

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper (commented):
LGTM from high-level

Dhairya-Arora01 (dismissed):

Dhairya-Arora01 (approved):

PR #2051, opened 2026-05-29
🌱 HBMH - v1beta2: simplify ConsumerRef

### LINKED ISSUE #2024: HBMH - v1beta2: simplify ConsumerRef
Part of #2017 (the CAPH v1beta2 migration umbrella). Read the umbrella first for the full migration plan and how this issue fits into the larger sequence; this issue covers the one internal cross CRD reference in CAPH's API package.

## Summary

On the v1beta2 `HetznerBareMetalHost`, replace `Spec.ConsumerRef` with a smaller reference shape that stores only the fields CAPH needs.

## The Problem

`HetznerBareMetalHost.Spec.ConsumerRef` links a host to the `HetznerBareMetalMachine` that consumes it. This is host ownership state inside CAPH.

The field uses `corev1.ObjectReference` today. That gives us more data than we need. The controllers use the machine name and namespace. The namespace always matches the host namespace because the host and the consuming machine live together. The current `UID`, `ResourceVersion`, and `FieldPath` values are not used.

The current shape also stores `apiVersion`. That is fragile during the v1beta2 migration because the referenced machine may be served as v1beta1 or v1beta2 depending on the client and the CRD conversion setup. CAPH does not need to persist that version on the host.

CAPI already changed this pattern in its v1beta2 API contract. For cross CRD references, CAPI uses `ContractVersionedObjectReference`, which stores kind, name, and API group, but not the concrete API version. The version can be inferred from the target CRD's contract label when a version is needed.

**`ConsumerRef` itself is CAPH internal state. CAPI does not own it or reconcile it.**

## The Solution

After this issue, `ConsumerRef` should still mean the same thing: this host is consumed by this `HetznerBareMetalMachine`. The visible change is only the shape of the v1beta2 field. It should store the target kind, name, and API group. It should not store a namespace, concrete API version, UID, resource version, or field path.

That shape matches how CAPH uses the field. The machine name identifies the consumer. The namespace is already known from the host namespace. The API group and kind identify the target resource type. The concrete API version is not part of the identity we need to persist on the host.

To get this behavior, define a local v1beta2 reference type with `kind`, `name`, and `apiGroup`, and use it for `HetznerBareMetalHost.Spec.ConsumerRef` in v1beta2. Keep the v1beta1 field as `corev1.ObjectReference` so existing clients keep working during the compatibility window.

Conversion should make the API change invisible to old clients. When converting v1beta1 to v1beta2, copy `Kind` and `Name`, and derive `APIGroup` from the old reference. Drop `Namespace`, `UID`, `ResourceVersion`, `FieldPath`, and the concrete version. When converting v1beta2 back to v1beta1, restore `Namespace` from the host namespace and rebuild the old reference shape for v1beta1 clients.

## Scope

This issue changes the v1beta2 API shape and the conversion logic for `ConsumerRef`.

It does not switch controllers to v1beta2 types. The controller switch is part of the later controller migration work tracked under #2017. Until then, controllers keep reading the v1beta1 shape. When that switch happens, the code paths that currently read `ConsumerRef.Namespace` should use the host namespace instead.

## References

* Parent umbrella issue: #2017.

abdullah599:
Simplifies v1beta2 `HetznerBareMetalHost.spec.consumerRef` to persist only `kind`, `name`, and `apiGroup`. v1beta1 keeps the old `ObjectReference` shape, with conversion preserving compatibility.

Fixes #2024

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper (commented):
LGTM

Dhairya-Arora01 (approved):

Issue #2017, opened 2026-05-17
CAPH v1beta2 migration


abdullah599:
## Summary

Track the full Cluster API v1beta2 contract migration for CAPH. This is an umbrella issue. Actual work lives in 16 sub issues plus the in flight `metav1.Conditions` PRs. Progress is tracked by the sub issue checkboxes below.

## Background

CAPI v1beta2 is the new provider contract. It was released in CAPI v1.11.0 on 2025-08-19. v1beta1 contract compatibility is scheduled for removal in April 2027 (per the CAPI contract pages). CAPH must complete the migration before that date. CAPH today is still on the v1beta1 contract. 

Each sub issue below carries the full description of its change. The parent issue keeps a short summary only.

## Timeline pressure

| Date | Event |
|---|---|
| 2025-08-19 | CAPI v1.11.0 released. v1beta2 contract is the new normal. |
| ~Apr 2027 | v1beta1 contract compatibility removal tentative. CAPH must finish before this. |

## Migration plan

The work is staged in five phases.

* **Phase 1: v1beta2 conditions on v1beta1 types** (in flight). Add `metav1.Condition` constants and a v1beta2 conditions slice on each v1beta1 type, set alongside the existing v1beta1 conditions during reconciliation. See "Existing condition work" below.
* **Phase 2: Prerequisites**. Six sub issues; see "Prerequisites" below. Establishes the v1beta2 API package and the groundwork that every resource issue depends on.
* **Phase 3: Per resource migration**. Six sub issues; see "Per resource" below. Each CRD's full v1beta2 cleanup. Includes wiring v1beta2 conditions on the v1beta2 type's `status.conditions`, which is separate from the Phase 1 work that put them on the v1beta1 types.
* **Phase 4: Switch over**. Five sub issues; see "Phase 4 switch over" below. Controllers and webhooks switch to v1beta2 types, storage version flips, templates and e2e migrate.
* **Phase 5: Deprecation cleanup** (post April 2027). Remove the v1beta1 API package and drop spoke conversion code. Recorded here, sub issues filed closer to the date.

This issue tracks Phases 1 through 4. Phase 5 is filed as separate work later.

## Existing condition work (Phase 1, in flight)

[...]

PR #2053, opened 2026-05-29
✨ v1beta2 migration: HetznerCluster

### LINKED ISSUE #2052: v1beta2 migration: HetznerCluster

## Summary

Move `HetznerCluster` from the copied v1beta1 API shape to the final v1beta2 InfraCluster contract shape.

## What would you like to add

This is part of the ongoing #2017 migration. The current `api/v1beta2` package started as a mechanical copy of `api/v1beta1`, so each resource still needs a pass that turns the copied shape into the final v1beta2 shape.

This issue is the `HetznerCluster` resource pass. It covers the `HetznerCluster` v1beta2 API shape, the related `HetznerClusterTemplate` metadata shape, and the conversion needed for both. It does not reopen the condition design.

PR #1964 already staged the final HetznerCluster v1beta2 condition set under v1beta1 `status.v1beta2.conditions`. This issue should promote those conditions to v1beta2 `status.conditions` and move the old v1beta1 condition slice to `status.deprecated.v1beta1.conditions`.

The final changes are:

| Area | v1beta2 change |
|---|---|
| Conditions | Promote v1beta1 `status.v1beta2.conditions` to v1beta2 `status.conditions`. Move old v1beta1 `status.conditions` to `status.deprecated.v1beta1.conditions`. Keep the #1964 condition set and summary behavior. |
| Initialization | Replace `status.ready` with `status.initialization.provisioned`. Treat this as a one time initialization signal, not as the `Ready` condition. |
| Failure domains | Change `status.failureDomains` from the v1beta1 map shape to `[]clusterv1.FailureDomain` with map list markers. |
| Endpoint | Change `spec.controlPlaneEndpoint` from `*clusterv1beta1.APIEndpoint` to a local non pointer `APIEndpoint` with CAPI v1beta2 validation. |
| Template metadata | Change `HetznerClusterTemplate.spec.template.metadata` from `clusterv1beta1.ObjectMeta` to `clusterv1.ObjectMeta` with `omitzero`. |
| SSA markers | Add explicit list markers to all lists used by `HetznerCluster`, including spec lists, status lists, conditions, and failure domains. |
| Print columns | Use default columns `Cluster`, `Ready`, `Age`; move `Endpoint`, `Regions`, `Network enabled`, `Reason`, and `Message` to wide output. `Ready` should read from `status.conditions[Ready]`. |
| SSH keys | Rename `spec.sshKeys.robotRescueSecretRef` to `spec.sshKeys.rescueSecretRef` and convert it both ways. |
| Conversion | Add manual conversion for all moved fields, including conditions, initialization, failure domains, endpoint, template metadata, and the renamed rescue SSH secret field. |

References:

* Parent issue: #2017.
* HetznerCluster v1beta2 conditions PR: #1964.
* CAPI InfrastructureCluster contract: https://cluster-api.sigs.k8s.io/developer/providers/contracts/infra-cluster.

abdullah599:

This is the first per-resource pass that takes a CAPH resource (HetznerCluster) from the mechanical v1beta2 copy to its final v1beta2 API shape, following the CAPI v1beta2 InfraCluster contract. 

Controllers, scope and services stay on v1beta1. Storage version stays v1beta1. Switching the controller to v1beta2 is tracked under #2017.

Fixes #2052

Audited the change against the CAPI v1beta2 InfraCluster contract (verified against CAPI's own conformant reference, CAPD `DockerCluster`).

| # | InfraCluster contract requirement | Change in this PR | Compliant |
|---|---|---|---|
| 1 | `spec.controlPlaneEndpoint` value type, `omitempty,omitzero`, `MinProperties=1`, Host/Port optional with min/max | Converted pointer -> value `APIEndpoint`; markers added; webhook nil-check dropped | ✅ Yes |
| 2 | `status.initialization.provisioned *bool`, set once and never updated after provisioning | Added `HetznerClusterInitializationStatus.Provisioned *bool`; comments mirror CAPD | ✅ Yes |
| 3 | `status.conditions []metav1.Condition`, `+listType=map`, `+listMapKey=type`, `MaxItems=32` | Retyped `Conditions`; `GetConditions`/`SetConditions` updated; `GetV1Beta1Conditions` added | ✅ Yes |
| 4 | `status.failureDomains []clusterv1.FailureDomain`, `+listType=map`, `+listMapKey=name`, `MinItems=1`, `MaxItems=100` | Retyped from map to slice; sorted conversion helpers both directions | ✅ Yes |
| 5 | `failureReason` / `failureMessage` deprecated and moved under `status.deprecated.v1beta1` | N/A — HetznerCluster never had these fields | ✅ N/A |
| 6 | `status.deprecated.v1beta1.conditions` pointer-wrapper shape | Added `Deprecated *...` -> `V1Beta1 *...` -> `Conditions`, matches CAPD | ✅ Yes |

Storage version stays v1beta1 (`storage: true` on v1beta1, `false` on v1beta2), which is correct because the controller still writes v1beta1.

Round-trip note: the `MarshalData` annotation only needs to restore `initialization.provisioned`, the one mapping where nil-vs-false is contractually meaningful. The other empty-vs-nil edges (`FailureDomain.ControlPlane` nil/false, empty condition slices) are semantically equivalent and are normalized in the fuzz round-trip test.

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta1/conversion.go
```diff
@@ -218,7 +221,207 @@ func Convert_v1beta1_HetznerBareMetalMachineStatus_To_v1beta2_HetznerBareMetalMa
...
+
+	out.FailureDomains = convertFailureDomainsToV1Beta1(in.FailureDomains)
+
+	// status.initialization.provisioned maps back to status.ready during the compatibility window.
+	out.Ready = ptr.Deref(in.Initialization.Provisioned, false)
+
+	return nil
+}
+
+func convertFailureDomainsToV1Beta2(in clusterv1beta1.FailureDomains) []clusterv1.FailureDomain {
```
janiskemper, line 401: can you explain shortly what you do in this function? I find it a bit hard to understand

api/v1beta1/conversion_test.go
```diff
@@ -109,5 +339,38 @@ func spokeV1Beta2StatusFuzzFuncs(_ runtimeserializer.CodecFactory) []interface{}
...
+		func(in *HetznerClusterStatus, c randfill.Continue) {
+			c.FillNoCustom(in)
+			if len(in.Conditions) == 0 {
+				in.Conditions = nil
+			}
+			if in.V1Beta2 != nil && len(in.V1Beta2.Conditions) == 0 {
+				in.V1Beta2 = nil
+			}
+		},
+		// HetznerCluster v1beta1 spec: a non-nil pointer to a zero controlPlaneEndpoint would
```
janiskemper: I'm not entirely understanding why this wouldn't be a problem in reality, if this round trip doesn't work

api/v1beta2/conditions_const.go
```diff
@@ -295,3 +295,103 @@ const (
 	// BootIDEmptyReason indicates that an empty boot ID is present on the node object.
 	BootIDEmptyReason = "BootIDEmpty"
 )
+
+// v1beta2 conditions.
+
+// common conditions used across resource types.
+
+const (
+	// HCloudRateLimitExceededV1Beta2Condition reports on whether the HCloud API rate limit has been exceeded.
```
janiskemper: I don't think that the names should still contain v1beta2

api/v1beta2/hetznercluster_types.go
```diff
@@ -81,31 +86,102 @@ type HetznerClusterSpec struct {
...
+// APIEndpoint represents a reachable Kubernetes API endpoint.
+// +kubebuilder:validation:MinProperties=1
+type APIEndpoint struct {
+	// Host is the hostname on which the API server is serving.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=512
+	Host string `json:"host,omitempty"`
+
+	// Port is the port on which the API server is serving.
```
janiskemper, line 97: I think all comments should start with lower case. Can you adapt that in this PR, or do you wanna align it later?

IMO we should only do this for v1beta2

api/v1beta2/hetznercluster_types.go
```diff
@@ -81,31 +86,102 @@ type HetznerClusterSpec struct {
...
 
 	// +optional
 	Network *NetworkStatus `json:"networkStatus,omitempty"`
 
 	ControlPlaneLoadBalancer *LoadBalancerStatus `json:"controlPlaneLoadBalancer,omitempty"`
 	// +optional
-	HCloudPlacementGroups []HCloudPlacementGroupStatus  `json:"hcloudPlacementGroups,omitempty"`
-	FailureDomains        clusterv1beta1.FailureDomains `json:"failureDomains,omitempty"`
-	Conditions            clusterv1beta1.Conditions     `json:"conditions,omitempty"`
+	// +listType=map
```
janiskemper, line 124: can you leave an empty line after ControlPlaneLoadbalancer?

api/v1beta1/conversion_test.go
abdullah599: The control plane endpoint is a pointer in v1beta1 and a value in v1beta2. The fuzz test does a full round trip and checks that the result still matches.

The only value that does not survive that round trip is a pointer to an empty endpoint, `&{}`. When we convert from v1beta1 to v1beta2 it becomes the zero value, and when we convert back it becomes `nil`. So `&{}` turns into `nil`. The fuzzer creates random `&{}` pointers, so without this line the test fails. That is the only reason for it.

In production this makes no difference, because `&{}` and `nil` mean the same thing: no endpoint is set. The controller either writes a real endpoint with a host and a port, or it leaves the field unset. It never stores `&{}` to mean something else. The v1beta2 type also uses `omitzero` and `MinProperties=1`, so an empty endpoint is already treated as unset by the API.

api/v1beta2/hetznercluster_types.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2061

api/v1beta2/conditions_const.go
```diff
@@ -19,279 +19,379 @@ package v1beta2
...
+	HCloudRateLimitExceededCondition = "HCloudRateLimitExceeded"
+	// HCloudRateLimitExceededReason indicates that the HCloud API rate limit has been exceeded.
+	HCloudRateLimitExceededReason = "Exceeded"
+)
+
+const (
+	// HCloudTokenAvailableCondition reports on whether the HCloud Token is available.
+	HCloudTokenAvailableCondition = "HCloudTokenAvailable"
+	// HCloudTokenAvailableReason indicates that the HCloudToken is available.
+	HCloudTokenAvailableReason = clusterv1beta1.AvailableV1Beta2Reason
```
janiskemper: this doesn't look right with clusterv1beta1 import

api/v1beta1/conversion.go
```diff
@@ -47,13 +39,34 @@ import (
...
 
 // ConvertFrom converts the Hub version (v1beta2) to this HetznerCluster.
 func (dst *HetznerCluster) ConvertFrom(srcRaw conversion.Hub) error {
 	src := srcRaw.(*infrav1.HetznerCluster)
-	return Convert_v1beta2_HetznerCluster_To_v1beta1_HetznerCluster(src, dst, nil)
+	if err := Convert_v1beta2_HetznerCluster_To_v1beta1_HetznerCluster(src, dst, nil); err != nil {
+		return err
+	}
+
+	// Preserve the hub object so the next up-conversion can restore hub-only intent (see ConvertTo).
```
Dhairya-Arora01: can you add here that it preserves the hub object in the annotation

api/v1beta1/conversion.go
```diff
@@ -292,7 +293,212 @@ func Convert_v1beta1_HetznerBareMetalMachineStatus_To_v1beta2_HetznerBareMetalMa
...
+// FailureDomains map, dereferencing each ControlPlane *bool (a nil pointer becomes false).
+func convertFailureDomainsToV1Beta1(in []clusterv1.FailureDomain) clusterv1beta1.FailureDomains {
+	if in == nil {
+		return nil
+	}
+
+	out := make(clusterv1beta1.FailureDomains, len(in))
+	for _, failureDomain := range in {
+		out[failureDomain.Name] = clusterv1beta1.FailureDomainSpec{
+			ControlPlane: ptr.Deref(failureDomain.ControlPlane, false),
```
Dhairya-Arora01, line 435: I have a question - why didn't we do this conversion based on the data present in the annotation? The nil state here for the ControlPlane here is lost in the round trip.

api/v1beta1/conversion_test.go
```diff
@@ -18,6 +18,8 @@ package v1beta1
```
Dhairya-Arora01, line 18: Do we have a testcase with status.ready: false but no annotation so it should be converted to provisioned=nil in v1beta2.
I think we should test that, as this path would be hit during storage migration. WDYT ?

api/v1beta1/conversion.go
abdullah599: We map it the same way CAPI core does. Their ClusterStatus conversion uses `ptr.To(fd.ControlPlane)` and `ptr.Deref(fd.ControlPlane, false)` for failure domains and does not read it back from the annotation. They keep the annotation only for the contract field `provisioned`, same as us 🙂

It is also safe here because `status.failureDomains` is owned by the controller. We rebuild it on every reconcile in `SetStatusFailureDomain` and always set `ControlPlane` to true, so a nil is never stored. So nil vs false is not a real state for this field.


Reference: https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.1/api/core/v1beta1/conversion.go#L1314

Dhairya-Arora01: Okay, 
Can you also tell me - why do we do it for `provisioned? How are provisioned=nil and provisioned=false are different from each other ?

api/v1beta1/conversion.go
```diff
@@ -65,7 +65,7 @@ func (dst *HetznerCluster) ConvertFrom(srcRaw conversion.Hub) error {
 		return err
 	}
 
-	// Preserve the hub object so the next up-conversion can restore hub-only intent (see ConvertTo).
+	// Preserve the hub object in a data annotation so the next up-conversion can restore hub-only intent (see ConvertTo).
```
Dhairya-Arora01: btw what do we mean by **hub-only intent** ?

api/v1beta1/conversion.go
abdullah599: For orchestration nil and false both mean "not provisioned yet", so the real difference is intent: nil is "no signal yet", false is "an explicit negative signal".

Converting v1beta2 to v1beta1 collapses both nil and false into `ready=false`, and converting back up the CAPI helper defaults to nil. So without the stored hub an explicit `false` would silently turn into nil on every round trip, which breaks the lossless round trip that conversion has to guarantee. MarshalData stores the original hub object so the exact `*bool` is restored.

This is the same pattern CAPI core uses. In `Cluster.ConvertTo` they UnmarshalData the hub and then map `status.infrastructureReady` and `status.controlPlaneReady` into `status.initialization.*` with the same helper. Our HetznerCluster `ready` to `initialization.provisioned` mapping mirrors their `infrastructureReady` to `initialization.infrastructureProvisioned`.

Failure domains have no such intent to preserve (the controller always sets ControlPlane), so we skip the annotation there.

References:
- CAPI core doing it for initialization: https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.1/api/core/v1beta1/conversion.go#L70-L86
- the helper rule (see comment regarding nil vs false): https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.1/api/core/v1beta2/conversion.go#L60

api/v1beta1/conversion.go
abdullah599: In conversion terms the hub is the stored version (v1beta2) and the spoke is the old one (v1beta1). "hub-only intent" just means information that only the hub can express and the spoke cannot, so it would be lost when we convert down to v1beta1.

i agree it is confusing i will update this :)

api/v1beta1/conversion.go
Dhairya-Arora01: Yes I understand why we do it - but I think we do this without a loss only for `provisioned`. But as you said that we align with CAPI (for converting failureDomains etc. as well). So I think overall it is fine. My concern earlier was bit about consistency.

Dhairya-Arora01 (commented):
LGTM, cc @janiskemper

janiskemper (approved):

PR #2072, opened 2026-06-04
:seedling: k8s deps in go.mod: Renovate should not update Go packages which reflect the Kubernetes Version

guettli:
Renovate should not update Go packages which reflect the Kubernetes Version.

Reason: PR #2069 tried to bump one dependency, which failed:

> go: github.com/syself/cluster-api-provider-hetzner/controllers tested by
	github.com/syself/cluster-api-provider-hetzner/controllers.test imports
	k8s.io/kubectl/pkg/scheme imports
	k8s.io/api/scheduling/v1alpha1: cannot find module providing package k8s.io/api/scheduling/v1alpha1

The list of packages to ignore was created by:

```console 
❯ grep k8s.io go.mod | grep -v indirect | grep  0.35 
    k8s.io/api v0.35.5
    k8s.io/apiextensions-apiserver v0.35.5
    k8s.io/apimachinery v0.35.5
    k8s.io/apiserver v0.35.5
    k8s.io/client-go v0.35.5
    k8s.io/kubectl v0.35.4
```

---

We want to update these by hand.

### DIFF
--- a/.github/renovate/golang.json5
+++ b/.github/renovate/golang.json5
@@ -31,7 +31,7 @@
     {
       description: "Disable update k8s packages",
       matchManagers: ["gomod"],
-      matchDepNames: ["k8s.io/api", "k8s.io/apiextensions-apiserver", "k8s.io/apimachinery", "k8s.io/apiserver", "k8s.io/client-go", "k8s.io/kubectl"],
+      matchDepNames: ["k8s.io/api", "k8s.io/apimachinery", "k8s.io/apiserver", "k8s.io/client-go", "k8s.io/kubectl", "k8s.io/code-generator"],
       enabled: false,
     },
     {
--- a/.github/renovate/golang.json5
+++ b/.github/renovate/golang.json5
@@ -31,13 +31,13 @@
     {
       description: "Disable update k8s packages",
       matchManagers: ["gomod"],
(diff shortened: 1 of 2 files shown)


janiskemper (commented):
we have excluded almost all k8s dependencies from renovatebot. I assume the ones you found there (also kind) was missed. I think we can just exclude it.

If we want to make it well, then we should group all k8s dependencies and allow patch version upgrades. 

Same with kind and other related dependencies.

guettli:
> we have excluded almost all k8s dependencies from renovatebot. I assume the ones you found there (also kind) was missed. I think we can just exclude it.
> 
> If we want to make it well, then we should group all k8s dependencies and allow patch version upgrades.
> 
> Same with kind and other related dependencies.

OK, I update the glob list by looking at:


```console 
❯ grep k8s.io go.mod | grep -v indirect | grep  0.35 
    k8s.io/api v0.35.5
    k8s.io/apiextensions-apiserver v0.35.5
    k8s.io/apimachinery v0.35.5
    k8s.io/apiserver v0.35.5
    k8s.io/client-go v0.35.5
    k8s.io/kubectl v0.35.4
```

I guess we want all these to be in sync. Here kubectl is different, but I guess this does not matter.

janiskemper (commented):
I think you should rename the title and change the description to reflect the code change

guettli:
> I think you should rename the title and change the description to reflect the code change

@janiskemper done.

janiskemper (approved):

PR #2075, opened 2026-06-05
✨ v1beta2 migration: HBMH

### LINKED ISSUE #2063: v1beta2 migration: HetznerBareMetalHost

## Summary

Move `HetznerBareMetalHost` from the copied v1beta1 API shape to its final v1beta2 shape. 

## What would you like to add

This is part of the ongoing #2017 migration. The current `api/v1beta2` package started as a mechanical copy of `api/v1beta1`, so each resource still needs a pass that turns the copied shape into the final v1beta2 shape.

This issue is the `HetznerBareMetalHost` resource pass. It covers the v1beta2 API shape and the conversion for that shape. `HetznerBareMetalHost` is a provider internal resource, not a CAPI contract resource, so this work follows CAPI status best practices and the agreed team cleanups, not a CAPI machine or cluster contract.

The condition set is already staged. PR #2013 writes the final `HetznerBareMetalHost` v1beta2 conditions into the v1beta1 resource under `spec.status.v1beta2.conditions`. This issue promotes those conditions to the v1beta2 `status.conditions` and moves the old v1beta1 condition slice to `status.deprecated.v1beta1.conditions`. It does not redesign the condition set except the one newly introduced v1beta2 condition (`ActionCompleted`).

This issue changes the API shape only. The controller keeps using v1beta1 types. The behavior changes that these cleanups enable belong to the later controller switch issue:

* Read `installImage`, `sshSpec`, `userData`, and the cluster live from the `HetznerBareMetalMachine` and the cluster, instead of copying them onto the host.
* Replace the `installImage` based deprovision trigger. Today the host deprovisions because the machine clears the host's `installImage`. With no copy, the host reacts to the machine being deleted, which means adding a watch on `HetznerBareMetalMachine`.
* Replace the `errorCount` backoff with static requeue tiers.
* Add a new `ActionCompleted` condition that takes over the `errorMessage` text. `SetError` marks it `False` with the error as the message, and `ClearError` removes it.

The final changes are:

| Area | v1beta2 change |
|---|---|
| Status location | Move the remaining fields from `spec.status` (`ControllerGeneratedStatus`) into the real `status` subresource, and remove the `spec.status` field. The dropped fields below do not move. |
| Conditions | Promote the staged `spec.status.v1beta2.conditions` to v1beta2 `status.conditions`. Move old v1beta1 `spec.status.conditions` to `status.deprecated.v1beta1.conditions`. Keep the #2013 condition set and summary behavior. |
| Drop copied machine data | Drop `userData`, `hetznerClusterRef`, `installImage`, and `sshSpec` from status. The host keeps only `consumerRef`, the binding to its machine. The controller reads the four dropped values live from the machine and the cluster (see above). |
| Drop backoff and errorMessage fields | Drop `errorCount` and `errorMessage` from status. Keep `errorType`, which the state machine reads for the reboot ladder. The `errorMessage` text is not lost: it moves to a new `ActionCompleted` condition |
| Status field hygiene | Drop the deprecated `lastUpdated`. Change `rebootTriggeredAt` from `*metav1.Time` to `metav1.Time`. |
| Drop CPU flags | Drop `status.hardwareDetails.cpu.flags`, which is captured but never read. |
| SSA markers | Add explicit list markers to `status.conditions`, `status.rebootTypes`, and the hardware detail lists. |
| Print columns | Default columns `Phase`, `HetznerBareMetalMachine`, `Maintenance`, `Ready`, `Age`. Move `IPv4`, `IPv6`, `CPU`, `RAM`, `Reason`, `Message` to wide output, and add `SERVER-ID` from `.spec.serverID` in wide output. `Ready` reads from `status.conditions[Ready]`. |
| Conversion | Add manual conversion for the `spec.status` to `status` move, the condition split, the dropped fields, and the `rebootTriggeredAt` pointer change. Preserve dropped fields through the conversion data annotation so the round trip stays lossless. |

References:

* Parent issue: #2017.
* HetznerBareMetalHost ConsumerRef issue: #2024.
* HetznerBareMetalHost v1beta2 conditions PR: #2013.

abdullah599:
This PR updates the HetznerBaremetalHost implementation and generated manifests for the new v1beta2 status shape.

Fixes #2063

None

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [x] add unit tests


api/v1beta1/conversion.go
```diff
@@ -205,13 +237,18 @@ func (dst *HetznerBareMetalRemediationTemplate) ConvertFrom(srcRaw conversion.Hu
 // host object for old clients.
 
 // Convert_v1beta1_HetznerBareMetalHost_To_v1beta2_HetznerBareMetalHost converts a v1beta1
-// HetznerBareMetalHost to v1beta2, including the v1beta2 ConsumerRef shape.
+// HetznerBareMetalHost to v1beta2. v1beta1 keeps the controller-generated status in spec.status,
```
janiskemper, line 266: why don't we move spec.status -> status and vice versa?

abdullah599: We do, in the hand-written object converters: `ConvertTo` moves `spec.status` to `status`, and `ConvertFrom` moves it back. It's by hand because conversion-gen can't map across the `spec.status`/`status` boundary, and the two shapes aren't identical (conditions are split, and a few v1beta1-only fields are kept in the conversion-data annotation for a lossless round trip).

janiskemper (commented):
LGTM from high-level

api/v1beta1/conversion.go
```diff
@@ -137,13 +139,43 @@ func (dst *HCloudRemediationTemplate) ConvertFrom(srcRaw conversion.Hub) error {
...
 	dst := dstRaw.(*infrav1.HetznerBareMetalHost)
-	return Convert_v1beta1_HetznerBareMetalHost_To_v1beta2_HetznerBareMetalHost(src, dst, nil)
+	if err := Convert_v1beta1_HetznerBareMetalHost_To_v1beta2_HetznerBareMetalHost(src, dst, nil); err != nil {
+		return err
+	}
+
+	// The spec.status fields collected by extractV1Beta1OnlyStatus have no equivalent in the v1beta2 shape.
+	// Stash them in the conversion data annotation on the hub so the matching ConvertFrom can restore
+	// them, keeping the round trip lossless.
+	if extracted := extractV1Beta1OnlyStatus(&src.Spec.Status); extracted != nil {
```
Dhairya-Arora01, line 149: why did we stored the complete object as annotation and not just the lossy fields in the conversion of some other types like `HetznerCluster` ?

api/v1beta1/conversion_test.go
```diff
@@ -211,6 +211,202 @@ func TestConvertHetznerBareMetalHostConsumerRefInvalidAPIVersion(t *testing.T) {
...
+	if status.ErrorCount != 3 || status.ErrorMessage != "boom" {
+		t.Fatalf("error fields not restored: count=%d message=%q", status.ErrorCount, status.ErrorMessage)
+	}
+	if status.LastUpdated == nil || !status.LastUpdated.Equal(&lastUpdated) {
+		t.Fatalf("lastUpdated not restored: %#v", status.LastUpdated)
+	}
+	if status.HardwareDetails == nil || !reflect.DeepEqual(status.HardwareDetails.CPU.Flags, []string{"sse", "avx"}) {
+		t.Fatalf("hardwareDetails.cpu.flags not restored: %#v", status.HardwareDetails)
+	}
+}
```
Dhairya-Arora01, line 408: Add a test for - when rebootTriggeredAt is nil pointer in v1beta1 -> so it should convert to zero value in v1beta2 and then nil pointer again in v1beta1

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -491,28 +497,31 @@ type NIC struct {
...
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status",description="HetznerBareMetalHost is ready"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Time duration since creation of BaremetalHost"
-// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".spec.status.conditions[?(@.type=='Ready')].reason"
-// +kubebuilder:printcolumn:name="Message",type="string",JSONPath=".spec.status.conditions[?(@.type=='Ready')].message"
+// +kubebuilder:printcolumn:name="IPv4",type="string",JSONPath=".status.ipv4",description="IPv4 of the host",priority=1
+// +kubebuilder:printcolumn:name="IPv6",type="string",JSONPath=".status.ipv6",description="IPv6 of the host",priority=1
+// +kubebuilder:printcolumn:name="CPU",type="string",JSONPath=".status.hardwareDetails.cpu.threads",description="CPU threads",priority=1
+// +kubebuilder:printcolumn:name="RAM",type="string",JSONPath=".status.hardwareDetails.ramGB",description="RAM in GB",priority=1
+// +kubebuilder:printcolumn:name="SERVER-ID",type="integer",JSONPath=".spec.serverID",description="Server ID of the host",priority=1
```
Dhairya-Arora01: should this be upper-case ?

api/v1beta1/conversion.go
abdullah599: we store only the lossy fields here, not the whole object. extractV1Beta1OnlyStatus builds a sparse HetznerBareMetalHost holding just the v1beta1-only fields (hetznerClusterRef, userData, installImage, sshSpec, errorCount, errorMessage, lastUpdated, cpu.flags) and marshals that 

does this answer your question or did i misunderstand something? 😅

Dhairya-Arora01: Exactly why did we store the complete object in case of conversion for HetznerCluster, why not store just the lossy fields in annotation ?

abdullah599: Both HetznerCluster and HBMH keep the v1beta1 to v1beta2 round trip lossless by stashing the fields one version cannot represent in the CAPI conversion data annotation, but in opposite directions.

HetznerCluster's one such field, provisioned, lives on the v1beta2 hub side, so it stashes the whole hub with MarshalData in ConvertFrom. That is the direction the annotation is built for, and with a single field it can stash unconditionally and read that one field back out, so copying the whole object is the simplest choice. This pattern is also followed by CAPI and CAPV

HBMH's dropped fields (installImage, sshSpec, errorCount, and a few more) live on the v1beta1 spoke side, so it stashes them in ConvertTo, the reverse direction. This case is not present in CAPI or CAPV AFAIK.
Here it writes the annotation on the hub, and in the `hub to spoke to hub` round trip that write is the last step, so nothing removes it afterwards. A needless annotation would then stay on the final object, the round trip would no longer return an identical object, the conversion would not be lossless, and the test would catch it and fail. So HBMH checks those fields, writes the annotation only when one is set, and restores exactly them on the way back. It stashes just those fields rather than the whole object since it already have those fields now.

HetznerCluster does not have this problem because its annotation lands on the spoke, where the round trip removes it: converting back to the hub reads and deletes it, and the CAPI test utility strips any leftover on the spoke before comparing.

abdullah599: We can store the whole object but I don't see any benefit. If you think otherwise we can store the whole object as well

Dhairya-Arora01: IMO we store the whole v1beta2 HetznerCluster object on v1beta1 spoke because in future we can have other fields in v1beta2  that have no representation in v1beta1.

Dhairya-Arora01 (approved):

api/v1beta1/conversion.go
abdullah599: Yes, but this is the other way around. HBMH v1beta1 has these fields and v1beta2 does not. Since we will not update v1beta1, this should be fine, no?

PR #2076, opened 2026-06-05
:seedling: Show permanent error via `ActionCompleted` condition

### LINKED ISSUE #2043: Show permanent error in conditions
**Context**
Currently, when a permanent error occurs during bare metal host provisioning (e.g. hardware reboot to rescue mode times out), the information about this error is stored in status.ErrorMessage but is not surfaced through conditions in a way that is easily visible to users. The `ProvisionSucceeded` condition was being used to preserve this information, but that leads to confusing state after deprovisioning (see issue #2041 ).

Also we plan to remove the field `status.ErrorMessage` in v1beta2. And introduce a generic condition `ActionCompleted` which should marked as false with reason as `errorType` and message as `errorMessage`.

We plan to implement this condition `ActionCompleted` to mirror `errorType` and message as `errorMessage` in a separate issue. However, as a part of this issue we can introduce the condition `ActionCompleted` with the sole reason `PermanentError` as a first step in the direction.

**Proposed Solution**
Introduce a new condition to surface errors clearly:
- Condition: `ActionCompleted`
- Reason: `PermanentError` (other reasons can be added later)

The priority order for setting the summary condition should incorporate `ActionCompleted` as the condition with second highest priority. After `RobotCredentialsAvailable` condition which is with the highest priority. Reason being during deprovisioning, host will be stuck in state deprovisioning if `RobotCredentialsAvailable` condition is false, even if the host has the permanent error condition.

Dhairya-Arora01:
In order to make the permanent errors more visible, we introduce the `ActionCompleted` condition with reason `PermanentError`. This condition is proposed to mirror the `errorType` and `errorMessage` but that will be implemented as part of a different issue.

Fixes #2043 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -713,3 +713,12 @@ const (
 	// HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason indicates the Robot API rate limit has been exceeded.
 	HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason = "Exceeded"
 )
+
+const (
+	// HetznerBareMetalHostActionCompletedV1Beta2Condition reports whether the last action on the host completed
+	// successfully. Set to False when a fatal, non-recoverable error is detected that requires manual intervention.
+	// Absent on healthy hosts; only present when a permanent error has been recorded.
+	HetznerBareMetalHostActionCompletedV1Beta2Condition = "ActionCompleted"
+	// HetznerBareMetalHostActionCompletedPermanentErrorV1Beta2Reason represents a fatal, non-recoverable error that persists on the host.
+	HetznerBareMetalHostActionCompletedPermanentErrorV1Beta2Reason = "PermanentError"
+)
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -189,6 +189,15 @@ const (
 )
 
 const (
+	// ActionCompletedCondition reports whether the last action on the host completed successfully.
+	// Set to False when a fatal, non-recoverable error is detected that requires manual intervention.
+	// Absent on healthy hosts; only present when a permanent error has been recorded.
+	ActionCompletedCondition clusterv1beta1.ConditionType = "ActionCompleted"
+	// ActionCompletedPermanentErrorReason represents a fatal, non-recoverable error that persists on the host.
+	ActionCompletedPermanentErrorReason = "PermanentError"
+)
+
+const (
 	// ProvisionSucceededCondition indicates that a host has been provisioned.
 	ProvisionSucceededCondition clusterv1beta1.ConditionType = "ProvisionSucceeded"
 	// StillProvisioningReason indicates that the server is still provisioning.
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -25,7 +25,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	clusterv1beta1 "sigs.k8s.io/cluster-api/api/core/v1beta1"
+	v1beta1conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"
 	v1beta2conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions/v1beta2"
 	"sigs.k8s.io/cluster-api/util/record"
 )
@@ -735,11 +737,18 @@ func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string)
 		if host.Annotations == nil {
 			host.Annotations = make(map[string]string, 1)
 		}
+
 		host.Annotations[PermanentErrorAnnotation] = time.Now().Format(time.RFC3339)
-		record.Warnf(host, "PermanentErrorSet", "Remove annotation %q, if you want the controller to use the hbmh again.",
-			PermanentErrorAnnotation)
+
+		// Put the host in maintenance mode.
+		host.Spec.MaintenanceMode = ptr.To(true)
+
+		record.Warnf(host, "PermanentErrorSet", "Turn off maintenance mode, if you want the controller to use the hbmh again.")
 
 		// set the ActionCompleted condition to false with reason PermanentError.
+		v1beta1conditions.MarkFalse(host, ActionCompletedCondition,
+			ActionCompletedPermanentErrorReason, clusterv1beta1.ConditionSeverityError,
+			"%s", errMessage)
 		v1beta2conditions.Set(host, metav1.Condition{
 			Type:    HetznerBareMetalHostActionCompletedV1Beta2Condition,
 			Status:  metav1.ConditionFalse,
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -426,13 +426,14 @@ func (host *HetznerBareMetalHost) SetV1Beta2Conditions(conditions []metav1.Condi
 // order (highest-priority first). Credentials and provisioning problems must outrank
 // Deleting, since deletion may itself need credentials to succeed.
 //  1. RobotCredentialsAvailable - invalid Robot credentials block every Robot API call.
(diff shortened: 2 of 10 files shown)


controllers/hetznerbaremetalhost_controller.go
```diff
@@ -609,6 +609,7 @@ func removePermanentErrorIfAnnotationIsGone(bmHost *infrav1.HetznerBareMetalHost
 	bmHost.Spec.Status.ErrorType = ""
 	bmHost.Spec.Status.ErrorMessage = ""
 	bmHost.Spec.Status.ErrorCount = 0
+	v1beta2conditions.Delete(bmHost, infrav1.HetznerBareMetalHostActionCompletedV1Beta2Condition)
```
abdullah599, line 616: **Bug: the `ActionCompleted` condition can get stuck, which keeps the host at `Ready=False` forever.**

This `Delete` is the only place that removes the condition, and it runs only while `ErrorType` is still `PermanentError` (the guard on line 599). The problem is that `ClearError()` sets `ErrorType` back to empty but does not touch this condition. Once that happens, the guard here is never true again, so the condition stays `False` and nothing can remove it.

**Example:**

1. A reboot times out during registration. `recordActionFailure(PermanentError)` runs, which calls `SetError` and sets `ActionCompleted=False`. The host stays in `StateRegistering`.
2. A few seconds later the server finishes booting into rescue. The next reconcile runs `actionRegistering` again, it succeeds, and it calls `ClearError()` (host.go:802). Now `ErrorType` is empty, but `ActionCompleted` is still `False`.
3. From now on the guard on line 599 is false, so this `Delete` never runs. The host stays at `Ready=False` at priority 2, and the user cannot clear it. Removing the annotation does not help either, because that path also checks `ErrorType == PermanentError`.

**Fix:** also delete the condition inside `ClearError()`:

```go
v1beta2conditions.Delete(host, HetznerBareMetalHostActionCompletedV1Beta2Condition)
```

This keeps the condition in sync with `ErrorType`. It is safe, because the one place that must keep a permanent error (deprovisioning, host.go:2578) already calls `ClearError()` only when `ErrorType != PermanentError`.

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -440,6 +440,7 @@ func HetznerBareMetalHostV1Beta2SummaryOpts() []v1beta2conditions.SummaryOption
 		// surfaces them in this order, so the most important issue is listed first.
 		v1beta2conditions.ForConditionTypes{
 			HetznerBareMetalHostRobotCredentialsAvailableV1Beta2Condition,
+			HetznerBareMetalHostActionCompletedV1Beta2Condition,
```
abdullah599, line 445: nit: The comment above this function (around lines 424 to 435) describes the priority order as 8 numbered conditions. This change adds `ActionCompleted` as the second priority, so that list no longer matches the code. Please add `ActionCompleted` as number 2 and shift the others down, so RobotRateLimitExceeded becomes 3 and Deleting becomes 9.

api/v1beta1/conditions_const.go
```diff
@@ -698,3 +698,12 @@ const (
 	// HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason indicates the Robot API rate limit has been exceeded.
 	HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason = "Exceeded"
 )
+
+const (
+	// HetznerBareMetalHostActionCompletedV1Beta2Condition reports that the host is currently in a non-terminal
```
abdullah599: The comment says the host is in a state that is not terminal. But the only reason the condition uses today is `PermanentError`, which the comment just below calls a fatal error that cannot be recovered

pkg/services/baremetal/host/host_test.go
```diff
@@ -959,6 +960,10 @@ var _ = Describe("actionPreparing", func() {
 		Expect(host.Spec.Status.ErrorMessage).To(ContainSubstring("no IPv4"))
 		Expect(host.Spec.Status.IPv4).To(BeEmpty())
 		Expect(host.Annotations).To(HaveKey(infrav1.PermanentErrorAnnotation))
+		actionCompletedCondition := v1beta2conditions.Get(host, infrav1.HetznerBareMetalHostActionCompletedV1Beta2Condition)
```
abdullah599, line 1050: **Suggestion: add a summary test.** These tests check the condition on its own, but none of them check the main goal of #2043, which is that `ActionCompleted=False` makes the Ready summary become `False` with reason `NotReady`, shown at priority 2. Consider a test in the scope package that sets the condition, calls `SetHetznerBareMetalHostV1Beta2ReadySummary`, and checks the resulting Ready condition. You can follow the pattern in `pkg/services/hcloud/server/server_test.go`.

api/v1beta1/hetznerbaremetalhost_types_test.go
```diff
@@ -479,8 +480,13 @@ func TestHetznerBareMetalHost_SetError(t *testing.T) {
 	}
 	host.SetError(PermanentError, "some error")
 	require.Equal(t, []string{PermanentErrorAnnotation, "other-annotation"}, mapKeys(host.Annotations))
+	actionCompletedCondition := v1beta2conditions.Get(&host, HetznerBareMetalHostActionCompletedV1Beta2Condition)
```
abdullah599, line 491: **Nit:** `SetError(PermanentError, ...)` runs on every reconcile while the error is present. Consider calling it twice here and checking that the condition stays `False` with the same reason and message, and that `ErrorCount` goes up. This covers the case where the condition is set again and again, which runs through the custom `GetV1Beta2Conditions` and `SetV1Beta2Conditions` logic.

abdullah599:
One more thing: The condition should be added to `WithOwnedV1Beta2Conditions`.

abdullah599 (commented):
LGTM :)
btw commit and PR description has a minor typo `PermanentErrror` ;)

janiskemper (commented):
LGTM from high-level

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -747,6 +768,7 @@ func (host *HetznerBareMetalHost) ClearError() {
 		host.Spec.Status.ErrorMessage = ""
 	}
 	host.Spec.Status.ErrorCount = 0
+	v1beta2conditions.Delete(host, HetznerBareMetalHostActionCompletedV1Beta2Condition)
```
abdullah599, line 771: ClearError is not symmetric with SetError. SetError sets both the v1beta1 and v1beta2 ActionCompleted conditions for a permanent error, but ClearError removes only the v1beta2 one, so the v1beta1 condition is left behind on the normal clear path.

api/v1beta1/hetznerbaremetalremediation_types.go
```diff
@@ -25,8 +25,9 @@ const (
 	RebootAnnotation = "capi.syself.com/reboot"
 
 	// PermanentErrorAnnotation indicates that the bare metal host has an error which needs to be resolved manually.
-	// After the permanent error the annotation got removed (usually by a human), the controller removes
-	// ErrorType, ErrorCount and ErrorMessages, so that the hbmh will be usable again.
+	// It is set together with Spec.MaintenanceMode=true. After the maintenance mode is
+	// turned off (usually by a human), the controller removes ErrorType, ErrorCount and ErrorMessages, as well as
```
abdullah599: ErrorMessage*

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -728,12 +733,28 @@ func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string)
 	host.Spec.Status.ErrorType = errType
```
abdullah599, line 732: SetError overwrites ErrorType with no check on the value already stored, so it never protects an error that is already permanent. A later call with a lower type, like the routine reboot bookkeeping at host.go:303, silently replaces PermanentError. Its permanent error branch does not run, so the annotation, MaintenanceMode and both conditions are left behind. Only removePermanentErrorIfResolved removes them, and it returns early once ErrorType is not PermanentError, so nothing can ever clean them up again

abdullah599 (commented):
should we update docs/caph/03-reference/08-annotations.md? it still tells users to remove the annotation to recover and lists Auto-Remove as Disabled. After this PR recovery is turning off maintenance mode and the controller removes the annotation itself

abdullah599 (approved):
LGTM

PR #2079, opened 2026-06-06
✨ v1beta2 migration: switch the HetznerCluster controller to v1beta2

### LINKED ISSUE #2077: v1beta2 migration: switch the HetznerCluster controller to v1beta2

## Summary

Move the `HetznerCluster` controller and its `ClusterScope` from the `api/v1beta1` types to the `api/v1beta2` types that landed in #2053, and make the shared rate limit and token helpers work for both API versions.

## What would you like to add

This is part of the ongoing #2017 migration. #2053 changed the `HetznerCluster` v1beta2 API to its final shape, but the controllers were left untouched, so every controller still imports `api/v1beta1` and writes conditions through the old CAPI condition packages.

This issue is the `HetznerCluster` controller pass. It switches the controller and `ClusterScope` to the v1beta2 types, moves condition handling to the CAPI v1.13 condition packages, and wires the v1beta2 status fields from #2053.

The other controllers stay on v1beta1 for now and switch in their own issues. The rate limit and token helpers live in the `HetznerCluster` controller but are shared by all six controllers, so this issue also splits them per version. After this change, a controller on either API version calls the variant for its version.

This issue does not redesign conditions. The condition set is #1964 and the API shape is #2053.

The final changes are:

| Area | Change |
|---|---|
| Imports | Point the `HetznerCluster` controller and `pkg/scope.ClusterScope` at `api/v1beta2` and reconcile the v1beta2 type. Stop embedding `ClusterScope` in `MachineScope` so `MachineScope` keeps its own v1beta1 fields while `ClusterScope` moves to v1beta2. |
| Condition packages | Move to `util/conditions` for v1beta2 conditions and `util/conditions/deprecated/v1beta1` for the deprecated conditions. |
| Condition constants | Use the renamed v1beta2 constants from #2053 (no `V1Beta2` suffix) and the `V1Beta1` suffixed deprecated set. |
| Dual write | Keep writing both the v1beta2 conditions and the deprecated v1beta1 conditions during the compatibility window, like CAPD's `DevCluster` controller. |
| Ready summary | Recreate the Ready summary inside the controller with `util/conditions`. #2053 removed `ClusterV1Beta2SummaryOpts` from the API package. |
| Shared helpers | Split `reconcileRateLimit`, `hcloudTokenErrorResult`, `getAndValidateHCloudToken`, `HandleRateLimitExceeded`, and `workloadClientConfigFromKubeconfigSecret` per version, with no transitional methods added to the v1beta2 API type. The v1beta2 variant keeps the original name and takes the concrete `*v1beta2.HetznerCluster`. The v1beta1 variant gets a `V1Beta1` suffix, and the v1beta1 condition helpers move to a new `controllers/conditions_v1beta1.go`. |
| Initialization | Set `status.initialization.provisioned` as a one time signal and stop reading `status.ready`. |
| Endpoint | Read and write `spec.controlPlaneEndpoint` as the non pointer `APIEndpoint`. Drop the load balancer `ErrControlPlaneEndpointNotSet` guard, which otherwise deadlocks the managed load balancer once the endpoint is a non pointer value. |
| Failure domains | Read and write `status.failureDomains` as `[]clusterv1.FailureDomain`. |
| SSH rescue secret | Use the renamed `spec.sshKeys.rescueSecretRef` field. |
| Conversion and tests | Enable the conversion webhook in the envtest setup, registering the v1beta2 scheme and setting `Scheme` on the `envtest.Environment`, because the CRD storage version stays v1beta1. Test both rate limit helpers. |

References:

* Parent issue: #2017.
* HetznerCluster v1beta2 API change: issue #2052, PR #2053.
* HetznerCluster v1beta2 conditions: PR #1964.
* CAPI InfrastructureCluster contract: https://cluster-api.sigs.k8s.io/developer/providers/contracts/infra-cluster.

abdullah599:
Switched the HetznerCluster controller to use v1beta2 HetznerCluster API instead of v1Beta1

Fixes #2077

None.

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


controllers/util.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	return false
+}
+
+// getAndValidateHCloudToken acquires the Hetzner secret referenced by the cluster and returns the
+// HCloud token from it. It returns a *ResolveSecretRefError if the secret is missing and a
+// *HCloudTokenValidationError if the token is empty, which the hcloudTokenErrorResult helpers map to
+// the right conditions.
+func getAndValidateHCloudToken(ctx context.Context, namespace string, hetznerCluster *infrav2.HetznerCluster, secretManager *secretutil.SecretManager) (string, *corev1.Secret, error) {
```
janiskemper, line 68: this has nothing to do with "conditions" IMO. what do you suggest? We could also name it "util.go" or shift this to another file?

controllers/conditions.go
```diff
@@ -0,0 +1,183 @@
...
+		return "", nil, &secretutil.HCloudTokenValidationError{}
+	}
+
+	return hcloudToken, hetznerSecret, nil
+}
+
+// hcloudTokenErrorResult handles errors from getAndValidateHCloudToken for controllers reconciling
+// v1beta2 resources. It sets the HCloudTokenAvailable condition (both surfaces), then computes the
+// Ready summary (the deprecated v1beta1 summary always, the v1beta2 summary from the passed opts)
+// before writing the status once. It mirrors hcloudTokenErrorResultV1Beta1.
```
janiskemper: don't say that because this is not interesting for people in the future.

controllers/conditions_v1beta1.go
```diff
@@ -0,0 +1,192 @@
...
+
+// reconcileRateLimitV1Beta1 is the still-v1beta1 counterpart of reconcileRateLimit. It detects the
+// rate limit through the deprecated v1beta1 HetznerAPIReachable condition and, when the wait is over,
+// marks it reachable again and removes the staged v1beta2 rate-limit condition if the object has one.
+func reconcileRateLimitV1Beta1(setter v1beta1conditions.Setter, rateLimitWaitTime time.Duration) bool {
+	condition := v1beta1conditions.Get(setter, infrav1.HetznerAPIReachableCondition)
+	if condition != nil && condition.Status == corev1.ConditionFalse {
+		if time.Now().Before(condition.LastTransitionTime.Add(rateLimitWaitTime)) {
+			// Not yet timed out, reconcile again after timeout.
+			// Don't give a more precise requeueAfter value to not reconcile too many
```
janiskemper: I think we set requeueAfter.. I think this comment is wrong (see also v1beta2 function)

controllers/controllers_suite_test.go
```diff
@@ -470,6 +473,18 @@ func isPresentAndFalseWithReasonV2(key types.NamespacedName, obj client.Object,
 	return c != nil && c.Status == corev1.ConditionFalse && c.Reason == reason
 }
 
+func isPresentAndTrueV2(key types.NamespacedName, obj client.Object, condition clusterv1.ConditionType) bool {
+	if err := testEnv.Get(ctx, key, obj); err != nil {
+		return false
+	}
+	getter, ok := obj.(deprecatedv1beta1conditions.Getter)
```
janiskemper, line 492: why do we check for v1beta1 here? Is there a plan when we want to migrate our tests to v1beta2 conditions?

controllers/controllers_suite_test.go
```diff
@@ -488,6 +503,11 @@ func isV1Beta2ConditionWithStatusAndReason(key types.NamespacedName, getter clie
 		return false
 	}
 
+	if nativeGetter, ok := getter.(conditions.Getter); ok {
```
janiskemper, line 532: why do you do this? How does it work, with the two getters? Why isn't this v1beta1 getter always satisfied? 

Why is this imported as "conditions" and the v1beta2 conditions as "v1beta2conditions"? Shouldn't we migrate?

controllers/controllers_suite_test.go
```diff
@@ -510,6 +530,10 @@ func isAbsentV1Beta2(key types.NamespacedName, getter client.Object, condition s
...
```
janiskemper, line 559: this function is named "v1beta2". It seems as if we should also migrate things here.

controllers/hetznercluster_controller.go
```diff
@@ -59,7 +58,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta1"
+	infrav2 "github.com/syself/cluster-api-provider-hetzner/api/v1beta2"
```
janiskemper, line 61: should we import this as "v2"? I think CAPI imports things as "v1" (e.g. clusterv1) even if it is v1beta2.

controllers/hetznercluster_controller.go
```diff
@@ -186,7 +190,7 @@ func (r *HetznerClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}()
 
 	// delete the deprecated condition from existing cluster objects
-	v1beta1conditions.Delete(hetznerCluster, infrav1.DeprecatedRateLimitExceededCondition)
+	deprecatedv1beta1conditions.Delete(hetznerCluster, infrav2.DeprecatedRateLimitExceededV1Beta1Condition)
```
janiskemper: I think we can remove all these "deprecated" conditions now

pkg/scope/cluster.go
```diff
@@ -119,34 +120,30 @@ func (s *ClusterScope) HetznerSecret() *corev1.Secret {
 
 // Close closes the current scope persisting the cluster configuration and status.
 func (s *ClusterScope) Close(ctx context.Context) error {
-	// set summary for v1beta1 conditions.
-	v1beta1conditions.SetSummary(s.HetznerCluster)
+	// set summary for deprecated v1beta1 conditions.
+	deprecatedv1beta1conditions.SetSummary(s.HetznerCluster)
 
 	// set summary for v1beta2 conditions.
```
janiskemper: remove "v1beta2" from here and from all places pretty much (this applies to the whole PR). This is the new "normal" so we don't really wanna talk about it

pkg/scope/machine.go
```diff
@@ -148,6 +204,11 @@ func (m *MachineScope) Namespace() string {
 	return m.HCloudMachine.Namespace
 }
 
+// HetznerSecret returns the hetzner secret.
```
janiskemper, line 207: add a todo that this should be removed later. It should, right?

janiskemper (changes requested):

controllers/controllers_suite_test.go
abdullah599: isPresentAndTrueV2 reads the deprecated v1beta1 conditions off the v1beta2
HetznerCluster (status.deprecated.v1beta1.conditions). The tests assert both surfaces on purpose:
each of these checks sits next to a native isPresentAndTrueWithReasonV1Beta2 check (e.g. lines
779-780), so we verify the controller dual-writes the legacy and the native conditions. We keep the
deprecated checks because the v1beta1 API still serves those conditions to old clients via the
conversion layer. Migrating the assertions to native-only is a clean follow-up once we decide the
deprecated surface no longer needs coverage, so I would defer it to keep this PR a mechanical switch.

controllers/controllers_suite_test.go
abdullah599: They are two different interfaces. conditions.Getter is the native util/conditions (reads status.conditions), which the v1beta2 HetznerCluster implements. v1beta2conditions.Getter is the deprecated staging package (reads status.v1beta2.conditions), which the still-v1beta1 objects implement. 

The v1beta1 getter is not satisfied because a native v2 object has GetConditions(), not GetV1Beta2Conditions(), so it does not satisfy the staged getter. We need both branches because this helper is called with the native HetznerCluster and with still-v1beta1 objects (HCloudMachine, HCloudMachineTemplate). 

On the names: conditions is the current package and v1beta2conditions is the deprecated staging package; the aliases follow our importas config, the name is unfortunate but repo-wide. When every resource is native v2 the staged branch and alias go away. I would defer that since it needs all resources migrated first.

controllers/controllers_suite_test.go
abdullah599: Same shape as the helper above: native branch for the v1beta2 HetznerCluster, staged
branch for the still-v1beta1 objects. When all resources are native v2
this collapses to conditions.Get(...) == nil and we can rename and merge these helpers. Deferring
to the same cleanup, once every resource is migrated. I think we can add TODOs here. WDYT?

controllers/hetznercluster_controller.go
abdullah599: yea. I think we can rename this to infrav1 and where we are importing v1beta1 now, we can call it `infrav1beta1` (this won't be present after the migration)

WDYT?

controllers/controllers_suite_test.go
janiskemper: I think you can add this to the code to explain the logic

controllers/controllers_suite_test.go
janiskemper: so you say that you plan on changing this in one of the later PRs of the current v1beta2 effort? In that case it makes sense to add a todo

controllers/hetznercluster_controller.go
janiskemper: Yes. I think that follows CAPI as well right?

abdullah599: yep

controllers/controllers_suite_test.go
janiskemper: so in the end we would have to rename this function (and the v1beta2 equivalent), to reflect that v1beta2 is the "new normal", right? Or am I misunderstanding something? (not sure in which file the lines 779-780 are)

controllers/hetznercluster_controller.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2092

Created a separate issue for this

controllers/controllers_suite_test.go
abdullah599: > so in the end we would have to rename this function (and the v1beta2 equivalent), to reflect that v1beta2 is the "new normal", right? 

Yes, I have done this in this PR

controllers/controllers_suite_test.go
abdullah599: > so you say that you plan on changing this in one of the later PRs of the current v1beta2 effort?

Yea, but I think it is better that we do it now. This is done

controllers/controllers_suite_test.go
```diff
@@ -470,7 +473,19 @@ func isPresentAndFalseWithReasonV2(key types.NamespacedName, obj client.Object,
...
+	}
+	getter, ok := obj.(deprecatedv1beta1conditions.Getter)
+	if !ok {
+		return false
+	}
+	c := deprecatedv1beta1conditions.Get(getter, condition)
+	return c != nil && c.Status == corev1.ConditionTrue
+}
+
+func isPresentAndTrueV1Beta1(key types.NamespacedName, getter v1beta1conditions.Getter, condition clusterv1beta1.ConditionType) bool {
```
janiskemper, line 508: do I undrestand this correctly that this one would be removed after our series of PRs is done? And that we keep only the other two because they check status.conditions and status.deprecated.v1beta1

abdullah599: yes
isPresentAndTrueV1Beta1 (and its ...FalseWithReasonV1Beta1 sibling) use the legacy v1beta1conditions getter, which only still-v1beta1 objects satisfy

janiskemper: can you add a todo then to this? Currently it's quite confusing that we have one for "deprecated v1beta1" and one for v1beta1

api/v1beta1/conversion.go
```diff
@@ -38,19 +38,19 @@ import (
 	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 
-	infrav1 "github.com/syself/cluster-api-provider-hetzner/api/v1beta2"
+	infrav2 "github.com/syself/cluster-api-provider-hetzner/api/v1beta2"
```
Dhairya-Arora01, line 41: lets do infrav1beta2 instead and infrav1 as infrav1beta1

controllers/hetznercluster_controller_test.go
```diff
@@ -1562,14 +1655,14 @@ var _ = Describe("reconcileRateLimit", func() {
...
-		Expect(reconcileRateLimit(hcloudMachine, testEnv.RateLimitWaitTime)).To(BeFalse())
+		Expect(reconcileRateLimitV1Beta1(hcloudMachine, testEnv.RateLimitWaitTime)).To(BeFalse())
 		// Condition must be deleted (not just set to False) so the next API call
 		// determines the real rate-limit status instead of assuming it is gone.
 		Expect(v1beta2conditions.Has(hcloudMachine, infrav1.HCloudRateLimitExceededV1Beta2Condition)).To(BeFalse())
 	})
 
 	It("returns wait==false if rate limit condition is set to true", func() {
-		v1beta1conditions.MarkTrue(hetznerCluster, infrav1.HetznerAPIReachableCondition)
+		deprecatedv1beta1conditions.MarkTrue(hetznerCluster, infrav2.HetznerAPIReachableV1Beta1Condition)
```
Dhairya-Arora01: I don't think this returns false for the intended reason.

In reconcileRateLimit we check if the native RateLimitExceeded condition is present. In this case its not present that;s why it returns false - which is not what we wanted to test here

api/v1beta1/conversion.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2092 :)

Dhairya-Arora01 (commented):
LGTM

janiskemper (approved):

PR #2089, opened 2026-06-09
:seedling: test/e2e: fail fast when no bare-metal host is available

guettli:
## Summary

- When a `HetznerBareMetalMachine` has `HostAssociateSucceeded=False` with reason `NoAvailableHost`, the e2e test previously waited the full 20-minute timeout before failing
- Root cause: in CI run [#27152609574](https://github.com/syself/cluster-api-provider-hetzner/actions/runs/27152609574/job/80147524840), all 5 bare-metal hosts were unavailable (2 in maintenance mode, 3 didn't match the label selector) — the test ran for 26 minutes before timing out
- The existing `logStatusContinuously` goroutine already had a fast-fail path for permanent host errors (`errPermanentHBMH`); this PR extends it to also detect the `NoAvailableHost` condition on `HetznerBareMetalMachine` objects and fail within ~30 seconds

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -86,9 +86,10 @@ var (
 
 // Test suite global vars.
 var (
-	ctx              = ctrl.SetupSignalHandler()
-	suiteStartTime   = time.Now()
-	errPermanentHBMH = errors.New("permanent HetznerBareMetalHost error")
+	ctx                = ctrl.SetupSignalHandler()
+	suiteStartTime     = time.Now()
+	errPermanentHBMH   = errors.New("permanent HetznerBareMetalHost error")
+	errNoAvailableHost = errors.New("no available bare-metal host")
 
 	// e2eConfig to be used for this test, read from configPath.
 	e2eConfig *clusterctl.E2EConfig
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -983,6 +988,23 @@ func logBareMetalHostStatus(ctx context.Context, c client.Client) error {
...
+	machineList := &infrav1.HetznerBareMetalMachineList{}
+	if err := c.List(ctx, machineList); err != nil {
+		return fmt.Errorf("failed to list HetznerBareMetalMachines: %w", err)
+	}
+	for _, machine := range machineList.Items {
+		if machine.DeletionTimestamp != nil {
+			continue
+		}
+		cond := v1beta1conditions.Get(&machine, infrav1.HostAssociateSucceededCondition)
+		if cond != nil && cond.Status == corev1.ConditionFalse && cond.Reason == string(infrav1.NoAvailableHostReason) {
```
abdullah599, line 1001: The suite now fails fast when a HetznerBareMetalMachine reports NoAvailableHost. Correct me if i am wrong but this condition is not always terminal. When no host can be associated, the controller marks NoAvailableHost and then requeues after 7 minutes. So it expects the state can recover on its own.

The causes are mixed. Some are permanent for the run, like a host in maintenance mode or a host that does not match the label selector. That is what the CI incident hit. Others are only temporary, like every host being in use while a previous cluster is still deprovisioning. Since logStatus polls every 30 seconds and fails on the first sighting, a temporary NoAvailableHost would abort the whole suite. 

Could we fail only after the condition stays stuck for a few minutes? This still fails fast instead of waiting the full 20 minute timeout, and it gives the controller room to recover. What do you think?

guettli: @abdullah599 the e2e tests creates an own mgt-cluster. Only one test runs at a time. I see no reason to wait. If there is no hbmh, then this won't change during the next minutes.

abdullah599: cool 👍

abdullah599 (approved):

PR #2091, opened 2026-06-09
:seedling: Refactor action EnsureProvisioned

### LINKED ISSUE #2073: Refactor `ensureProvisioned` to reliably detect failed rescue mode reboot
**Context**
We currently handle post installimage provisioning by triggering a reboot and then verifying that the host transitions from the rescue system to the desired OS. 
However, we don't reliably distinguish between whether the host is still stuck in rescue mode or an ongoing reboot.

The goal of this issue is to make `ensureProvisioned` robust against failed or stuck reboots by explicitly detecting whether the host is still in rescue mode and waiting for reboot completion before validating if the host is in the desired OS.

**Proposal**:

We propose to refactor `actionEnsureProvisioned` into two steps:

Before entering either step, check if `ErrorTypeHardwareRebootTriggered` is set on host. If set go directly to step-2 (i.e. hw reboot is already triggered). Otherwise go to step-1.

**Step-1**: Check if the host is still in rescue mode
SSH on port 22 and check `hostname`.

If hostname is `rescue`:
- Check if the host was just rebooted (within 15s period)
  - If yes, reque with delay.
- Ensure rescue is not set for the next reboot.
  - If rescue is set for next boot, unset it via robot API.
- Based on the current `ErrorType`, decide what to do:
  - *Empty*: This means no prior reboot triggered. Use [`rebootAndErrorTypeAfterTimeout()`](https://github.com/syself/cluster-api-provider-hetzner/blob/f4d4a805c7aa3eee1296f0b191bd9c6891a3d311/pkg/services/baremetal/host/host.go#L504) to pick software reboot (if the host supports it) otherwise hardware reboot. Trigger the chosen reboot via Robot API. Set the corresponding `ErrorType` (`ErrorTypeSoftwareRebootTriggered` or `ErrorTypeHardwareRebootTriggered`) if it was hardware reboot move to step-2 otherwise reque.
  - *`ErrorTypeSoftwareRebootTriggered` and not timed out yet (within `softwareResetTimeout` of 10 min)*: The software reboot is still in progress, so reque with delay.
  - *`ErrorTypeSoftwareRebootTriggered` and timed out*: Software reboot failed to bring the host out of rescue. Escalate to hardware reboot. Trigger hardware reboot via Robot API, set `ErrorTypeHardwareRebootTriggered` and move to Step-2.

If SSH error or hostname is not rescue, then also move to step-2

**Step-2**: 
If `SSHAfterInstallImage` is disabled, mark the action as completed, clear the `ErrorType` and move to the next phase `Provisioned`.
Otherwise, SSH via configured OS port and check hostname.
- If SSH error (connection reset or timed out): reque with delay.
- If hostname matches the desired hostname: mark the action as completed, clear the errorType and move to the next phase.
- If the hostname doesn't match:
  - If `ErrorTypeHardwareRebootTriggered` and not yet timed out (within `hardwareResetTimeout` of 10 min): requeue with delay  as the hardware reboot is still in progress.
  - Otherwise set a permanent error.

**NOTE:** We should reuse the existing reboot and timeout related utility functions where possible.

Dhairya-Arora01:
Refactor `actionEnsureProvisioned` to reliably detect failed rescue mode reboot.

actionEnsureProvisioned is refactored into 2 steps:
 **Step-1**: Check if the host is still in rescue mode SSH on port 22 and check `hostname`.

If hostname is `rescue`:
- Check if the host was just rebooted (within 15s period)
  - If yes, reque with delay.
- Ensure rescue is not set for the next reboot.
  - If rescue is set for next boot, unset it via robot API.
- Based on the current `ErrorType`, decide what to do:
  - *Empty*: This means no prior reboot triggered. Use trigger software reboot (if the host supports it) otherwise hardware reboot.
  - *`ErrorTypeSoftwareRebootTriggered` and not timed out yet (within `softwareResetTimeout` of 10 min)*: The software reboot is still in progress, so reque with delay.
  - *`ErrorTypeSoftwareRebootTriggered` and timed out*: Software reboot failed to bring the host out of rescue. Escalate to hardware reboot. Trigger hardware reboot via Robot API, set `ErrorTypeHardwareRebootTriggered` and move to Step-2.

If SSH error or hostname is not rescue, then also move to step-2

**Step-2**:
If `SSHAfterInstallImage` is disabled, mark the action as completed, clear the `ErrorType` and move to the next phase `Provisioned`. Otherwise, SSH via configured OS port and check hostname.
- If hostname matches the desired hostname: mark the action as completed, clear the errorType and move to the next phase.
- If the hostname doesn't match:
  - If `ErrorTypeHardwareRebootTriggered` and not yet timed out (within `hardwareResetTimeout` of 10 min): requeue with delay as the hardware reboot is still in progress.
  - Otherwise set a fatal error.

Fixes #2073 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
 func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult) {
 	markProvisionPending(s.scope.HetznerBareMetalHost, infrav1.StateEnsureProvisioned)
 
+	// If hardware reboot was triggered, then move to Step-2 i.e. ensuring host is booted into the desired OS.
+	// Otherwise, move to Step-1 i.e. check if the host is still in rescue mode.
+	if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType != infrav1.ErrorTypeHardwareRebootTriggered {
+		if ar := s.checkRescueAndTriggerReboot(ctx); ar != nil {
+			return ar
+		}
+	}
+
+	// Step-2
+	return s.verifyProvisionedOS(ctx)
+}
+
+// checkRescueAndTriggerReboot checks if the host is still in rescue mode, if it is still in rescue mode
+// it triggers a reboot. First a software reboot and if it fails then a hardware reboot.
+func (s *Service) checkRescueAndTriggerReboot(ctx context.Context) actionResult {
+	// If software reboot has timed out, escalate to hardware reboot immediately. No need to
+	// confirm rescue reachability, the host may be in a state where neither the rescue mode nor
+	// the desired OS might be reachable.
+	if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType == infrav1.ErrorTypeSoftwareRebootTriggered &&
+		hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt, softwareResetTimeout) {
+		// Ensure that rescue mode is disabled for the next reboot.
+		if err := s.unsetRescueIfActive(); err != nil {
+			return actionError{err: fmt.Errorf("failed to ensure that rescue mode is disabled for the next boot: %w", err)}
+		}
+
+		// trigger a hardware reboot.
+		if _, err := s.scope.RobotClient.RebootBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID, infrav1.RebootTypeHardware); err != nil {
+			s.handleRobotRateLimitExceeded(err, rebootServerStr)
+			return actionError{err: fmt.Errorf(errMsgFailedReboot, err)}
+		}
+
+		s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
+		msg := createRebootEvent(ctx, s.scope.HetznerBareMetalHost, infrav1.RebootTypeHardware, "Software reboot timed out, triggered a hardware reboot.")
+		s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeHardwareRebootTriggered, msg)
+
+		// move to step-2
+		return nil
+	}
+
+	creds := sshclient.CredentialsFromSecret(s.scope.RescueSSHSecret, s.scope.HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef)
+	in := sshclient.Input{
+		PrivateKey: creds.PrivateKey,
+		Port:       rescuePort,
+		IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
+	}
+	sshClient := s.scope.SSHClientFactory.NewClient(in)
+
+	// Check hostname with sshClient.
+	out := sshClient.GetHostName(ctx)
+	if out.Err != nil || out.StdErr != "" {
+		// Rescue system is not reachable, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
+		return nil
+	}
+
+	hostname := trimLineBreak(out.StdOut)
+	if hostname != rescue {
+		// Host is not in rescue mode, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
+		//
+		// NOTE: We are not analyzing the SSH errors because in this step we only need to detect if the host is in rescue mode.
+		// Any SSH error would mean that we cannot confirm if the host is in rescue mode so we fall through to step-2.
+		// connection-refused/timeout: host is between boots or already in the OS (step-2 will verify).
+		// auth failed: host rebooted into OS, so rescue key is no longer valied (step-2 will verify).
+		// non-rescue hostname: host booted into OS (step-2 will verify).
+
+		return nil
+	}
+
+	// Host is in rescue mode.
+
+	// If it was not rebooted in the last 15s then we should trigger a reboot.
+	if s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt != nil &&
+		!hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt, rebootWaitTime) {
+		// reboot was recently triggered (possibly from image-installing), give it some time.
+		return actionContinue{delay: 5 * time.Second}
+	}
+
+	// Ensure that rescue mode is disabled for the next reboot.
+	if err := s.unsetRescueIfActive(); err != nil {
+		return actionError{err: fmt.Errorf("failed to ensure that rescue mode is disabled for the next boot: %w", err)}
+	}
+
+	// Check which reboot was triggered previously, if a software reboot was triggered in the previous run
+	// and it has timed out now then trigger a hardware reboot.
+	var emptyErrorType infrav1.ErrorType
+	switch s.scope.HetznerBareMetalHost.Spec.Status.ErrorType {
+	case emptyErrorType:
+		// pick the available reboot type (software or hardware) and trigger it (preferred order software > hardware).
+		rebootType, errorType := rebootAndErrorTypeAfterTimeout(s.scope.HetznerBareMetalHost)
+		if _, err := s.scope.RobotClient.RebootBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID, rebootType); err != nil {
+			s.handleRobotRateLimitExceeded(err, rebootServerStr)
+			return actionError{err: fmt.Errorf(errMsgFailedReboot, err)}
+		}
+
+		s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
+		msg := createRebootEvent(ctx, s.scope.HetznerBareMetalHost, rebootType, "Rebooting out of rescue mode into the provisioned OS.")
+		s.scope.HetznerBareMetalHost.SetError(errorType, msg)
+
+		// if the hardware reboot was triggered then move to step-2, otherwise reque with delay.
+		if rebootType == infrav1.RebootTypeHardware {
+			return nil
+		}
+
+		return actionContinue{delay: 5 * time.Second}
+
+	case infrav1.ErrorTypeSoftwareRebootTriggered:
+		// If the software reboot timed out, then trigger a hardware reboot, otherwise reque.
+		if !hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt, softwareResetTimeout) {
+			markProvisionPendingWithInfo(s.scope.HetznerBareMetalHost, infrav1.StateEnsureProvisioned, "waiting for software reboot to complete")
+			return actionContinue{delay: 5 * time.Second}
+		}
+
+		// trigger a hardware reboot.
+		if _, err := s.scope.RobotClient.RebootBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID, infrav1.RebootTypeHardware); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
 func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult) {
 	markProvisionPending(s.scope.HetznerBareMetalHost, infrav1.StateEnsureProvisioned)
 
+	// If hardware reboot was triggered, then move to Step-2 i.e. ensuring host is booted into the desired OS.
```
janiskemper: explain why you do this check... 

IMO the explanation is: in case the hardware reboot is triggered already, we have reached the last phase of escalation already.

janiskemper: can you explain the two steps a bit better here in this function? It's not really undrestandable

pkg/services/baremetal/host/host.go
```diff
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+		PrivateKey: creds.PrivateKey,
+		Port:       rescuePort,
+		IP:         s.scope.HetznerBareMetalHost.Spec.Status.GetIPAddress(),
+	}
+	sshClient := s.scope.SSHClientFactory.NewClient(in)
+
+	// Check hostname with sshClient.
+	out := sshClient.GetHostName(ctx)
+	if out.Err != nil || out.StdErr != "" {
+		// Rescue system is not reachable, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
```
janiskemper: I think we can write this a bit differently: "We failed to reach the server via ssh and assume that it rebooting, or has rebooted already. Therefore, we can go to step 2 to verify that it is provisions successfully.

pkg/services/baremetal/host/host.go
```diff
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+		return nil
+	}
+
+	hostname := trimLineBreak(out.StdOut)
+	if hostname != rescue {
+		// Host is not in rescue mode, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
+		//
+		// NOTE: We are not analyzing the SSH errors because in this step we only need to detect if the host is in rescue mode.
```
janiskemper: didn't we analyze SSH errors before? I don't understand this.. Do you mean that we don't check for specific errors and differentiate them? In this case, you should probably add this to the if statement above

pkg/services/baremetal/host/host.go
```diff
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+
+		// if the hardware reboot was triggered then move to step-2, otherwise reque with delay.
+		if rebootType == infrav1.RebootTypeHardware {
+			return nil
+		}
+
+		return actionContinue{delay: 5 * time.Second}
+
+	case infrav1.ErrorTypeSoftwareRebootTriggered:
+		// If the software reboot timed out, then trigger a hardware reboot, otherwise reque.
```
janiskemper: typo

pkg/services/baremetal/host/host.go
```diff
@@ -1907,9 +1907,156 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+		if !hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt, softwareResetTimeout) {
+			markProvisionPendingWithInfo(s.scope.HetznerBareMetalHost, infrav1.StateEnsureProvisioned, "waiting for software reboot to complete")
+			return actionContinue{delay: 5 * time.Second}
```
janiskemper: should we requeue every 5 seconds? Seems very quickly. How do we do it in other places? We should keep that aligned probably.

pkg/services/baremetal/host/host.go
```diff
@@ -1927,54 +2074,37 @@ func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult)
...
+	// Check hostname with sshClient.
+	desiredHostName := s.scope.Hostname()
 	out := sshClient.GetHostName(ctx)
 	hostname := trimLineBreak(out.StdOut)
-	if hostname != wantHostName {
-		// give the reboot some time until it takes effect
+
+	if hostname != desiredHostName {
+		// if the host was just rebooted, give it some time.
 		if s.hasJustRebooted() {
```
janiskemper, line 2197: this function is something you don't re-use above - but you have the same case, don't you?

pkg/services/baremetal/host/host.go
```diff
@@ -1927,54 +2074,37 @@ func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult)
...
+		if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType == infrav1.ErrorTypeHardwareRebootTriggered &&
+			hasTimedOut(s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt, hardwareResetTimeout) {
+			msg := fmt.Sprintf("hostname %q does not match expected %q after reboot", hostname, desiredHostName)
 
-		failed, err := s.handleIncompleteBoot(ctx, false, isTimeout, isSSHConnectionRefusedError)
-		if failed {
-			msg := "reboot handling failed"
-			if err != nil {
-				msg = err.Error()
+			if sshclient.IsConnectionRefusedError(out.Err) || os.IsTimeout(out.Err) || sshclient.IsTimeoutError(out.Err) {
```
janiskemper: I don't think that this makes a lot of sense. If it is another error why we cannot reach the server, then we would say "host name doesn't match" even though we didn't even manage to ssh onto the server.

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
Dhairya-Arora01: Okay I will use 10s which is aligned to the general fallback case that we previously had.

pkg/services/baremetal/host/host.go
Dhairya-Arora01: IMO here it is a different case, because [`hasJustRebooted()`](https://github.com/syself/cluster-api-provider-hetzner/blob/6f29b9e9790ac39b2f6f32bd03d4dace94e1cc80/pkg/services/baremetal/host/host.go#L2765-L2774) requires the `ErrorType` to be one of software or hardware reboot triggered in our case.  But in the function above i.e. in step-1 , the host has `ErrorType` cleared  in the image-installing phase.

That's why we have the `RebootTriggered != nil` check here to have the same behaviour https://github.com/syself/cluster-api-provider-hetzner/blob/6f29b9e9790ac39b2f6f32bd03d4dace94e1cc80/pkg/services/baremetal/host/host.go#L1996-L2000

pkg/services/baremetal/host/host.go
```diff
@@ -2001,9 +2002,174 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+	// after a timeout then trigger a hardware reboot.
+	//
+	// Step-2: Verify if the host is booted into the desired OS.
+
+	// If hardware reboot was triggered, then it means that we have reached the
+	// last phase of escalation of getting out of rescue mode already.
+	// Therefore move to Step-2 i.e. ensuring host is booted into the desired OS.
+	// Otherwise, move to Step-1 i.e. check if the host is still in rescue mode.
+	if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType != infrav1.ErrorTypeHardwareRebootTriggered {
+		if ar := s.checkRescueAndTriggerReboot(ctx); ar != nil {
```
janiskemper, line 2019: ar?

pkg/services/baremetal/host/host.go
```diff
@@ -2001,9 +2002,174 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+
+// unsetRescueIfActive checks whether rescue boot is currently set for the next boot and unsets it if so.
+func (s *Service) unsetRescueIfActive() error {
+	rescue, err := s.scope.RobotClient.GetBootRescue(s.scope.HetznerBareMetalHost.Spec.ServerID)
+	if err != nil {
+		s.handleRobotRateLimitExceeded(err, "GetBootRescue")
+		return fmt.Errorf("failed to get boot rescue: %w", err)
+	}
+	if rescue.Active {
+		if _, err := s.scope.RobotClient.DeleteBootRescue(s.scope.HetznerBareMetalHost.Spec.ServerID); err != nil {
```
janiskemper, line 2159: in theory this is asynchronous right? I suppose we should block this operation here though for a max of 5 seconds to ensure that rescue is not active anymore once we reboot. We can use the action returned from Hetzner for this.

pkg/services/baremetal/host/host.go
```diff
@@ -2001,9 +2002,174 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+
+		// trigger a hardware reboot.
+		if _, err := s.scope.RobotClient.RebootBMServer(s.scope.HetznerBareMetalHost.Spec.ServerID, infrav1.RebootTypeHardware); err != nil {
+			s.handleRobotRateLimitExceeded(err, rebootServerStr)
+			return actionError{err: fmt.Errorf(errMsgFailedReboot, err)}
+		}
+
+		s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
+		msg := createRebootEvent(ctx, s.scope.HetznerBareMetalHost, infrav1.RebootTypeHardware, "Software reboot timed out, triggered a hardware reboot.")
+		s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeHardwareRebootTriggered, msg)
```
janiskemper, line 2049: not entirely sure if we stopped doing that in the new v1beta2. Can you check that PR?

pkg/services/baremetal/host/host.go
```diff
@@ -2001,9 +2002,174 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+
+	hostname := trimLineBreak(out.StdOut)
+	if hostname != rescue {
+		// Host is not in rescue mode, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
+		return nil
+	}
+
+	// Host is in rescue mode.
+
+	// If it was not rebooted in the last 15s then we should trigger a reboot.
```
janiskemper, line 2089: I don't understand this comment here. E.g. when we have already done a software reset, we wait until it times out before we trigger a hardware reset.

pkg/services/baremetal/host/host.go
```diff
@@ -2001,9 +2002,174 @@ func verifyConnectionRefused(ctx context.Context, sshClient sshclient.Client, po
...
+		// connection-refused/timeout error: host is between boots or already in the
+		// OS (step-2 will verify).
+		// auth failed: host rebooted into OS, so rescue key is no longer valid
+		// (step-2 will verify).
+		return nil
+	}
+
+	hostname := trimLineBreak(out.StdOut)
+	if hostname != rescue {
+		// Host is not in rescue mode, we can move to step-2 i.e. verifying if host is booted into provisioned OS.
```
janiskemper, line 2083: this confused me a bit but we do it only because step 2 actually completes immediately if ssh access is not activated. I think we should mention it here because it looks totally strange to go into the verification of the OS if we just rebooted and know we cannot access it.

janiskemper (commented):
plz update it to the new v1beta2 controller.

Issue #2095, opened 2026-06-11
ValidateKubeletCSR rejects vSwitch IPs because the allow-list only knows inspection-time NICs


ashish1099:
## Summary

`pkg/csr/csr.go:ValidateKubeletCSR` builds its allowed-IP set from `HetznerBareMetalMachine.Status.Addresses`, which is populated by `pkg/services/baremetal/baremetal/baremetal.go:nodeAddresses()` from `HetznerBareMetalHost.Spec.Status.HardwareDetails.NIC`. The NIC list is a one-shot snapshot taken during HBO's rescue-OS inspection — **before** the operator's `preKubeadmCommands` get a chance to bring up vSwitch sub-interfaces. Result: every kubelet-serving CSR whose SAN includes a vSwitch IP is denied with:

```
Validation by cluster-api-provider-hetzner failed: the IP address "10.0.1.x" is not allowed
```

## Concrete failure

Pure bare-metal cluster using a Hetzner Robot vSwitch as the in-cluster private network:

1. Bootstrap's `preKubeadmCommands` runs `netplan apply` to create `enp35s0.<vlanID>` with `10.0.1.x`, then configures kubelet with `--node-ip=10.0.1.x` (and `nodeIP:` in `KubeletConfiguration`).
2. kubelet starts (with `serverTLSBootstrap: true`), generates a serving CSR including `10.0.1.x` in `subjectAltName`.
3. CAPH's CSR controller calls `ValidateKubeletCSR`. The allow-list was derived at inspection time, when `enp35s0` had only its public IP and `enp35s0.<vlanID>` didn't exist yet. `10.0.1.x` is not in the allow-list. CSR denied.
4. kubelet keeps regenerating the CSR (same SANs, same denial).
5. Without a properly-signed serving cert, every apiserver→kubelet streaming subresource — `kubectl port-forward`, `exec`, `logs` — fails with `unable to upgrade connection: Unauthorized`. (The 401 is how the apiserver's streaming proxy surfaces an upstream TLS SAN mismatch.)

Public IPs work today only because HBO sees them at inspection time. vSwitch IPs are structurally invisible at that point because they're created by `preKubeadm`, hours after inspection.

## Proposed fix

[...]

PR #2096, opened 2026-06-11
:seedling: re-read hardwareDetails

### LINKED ISSUE #1789: Always set HetznerBareMetalHost.Spec.Status.HardwareDetails
/kind feature

After the IP of a bm machine was change, I want reprovisioning to work. I don't want to modify the hbmh by hand/script.

Currently, the hardwareDetails of hbmh are only set, when they do not exist yet:

in ActionRegisteringing():

```go
...
	if s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails == nil {
		hardwareDetails, err := getHardwareDetails(sshClient)
		if err != nil {
			return actionError{err: fmt.Errorf("failed to get hardware details: %w", err)}
		}
		s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails = &hardwareDetails
	}
...
```

I suggest that we always do that (always overwrite the struct).

Use-case of a customer: They ordered a better NIC and thus the machine got a new IP.

Now the customer needs to update or recreate the hbmh. That is not convenient.

Currently I see no reason to not always update hardwareDetails.

guettli:
Always refresh `HardwareDetails` in `actionRegistering` instead of only setting it when nil. This means hardware changes (e.g. NIC replacement → new IP, disk exchange) are picked up automatically on re-registration without manual intervention.

When the refreshed details differ from the previously stored ones, the controller logs a `"HardwareDetails changed"` message with the diff and emits a `HardwareDetails changed` Kubernetes event on the `HetznerBareMetalHost` object, so the change is visible on the object itself, not just in the logs.

Even if a hardware change makes the configured `RootDeviceHints` invalid (e.g. the disk referenced by WWN was replaced), the freshly read `HardwareDetails` is still persisted before that validation runs, so the object's status always reflects the current hardware.

Fixes #1789 

Related: #2095


janiskemper:
can you explain how this fixes the issues?

guettli:
> can you explain how this fixes the issues?

Before the HardwareDetails got only set, when they were nil:

```go
if s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails == nil
```

Now they get set always. This means, if the HardwareDetails were outdated, they get updated.

janiskemper:
how is that related to the vswitch issue?

guettli:
@janiskemper 

> how is that related to the vswitch issue?

I suggest to update HardwareDetails to include the vSwitch IP. Otherwise the security check is useless: When we allow the sender to include data to  a list, which gets used by the receiver to check the data of the sender, then this is loop. Caph must read the data from somewhere else, not from CSR/Node.

pkg/services/baremetal/host/host.go
```diff
@@ -698,13 +699,14 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
...
-		if err != nil {
-			return actionError{err: fmt.Errorf("failed to get hardware details: %w", err)}
-		}
-		s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails = &hardwareDetails
+	hardwareDetails, err := getHardwareDetails(ctx, sshClient)
+	if err != nil {
+		return actionError{err: fmt.Errorf("failed to get hardware details: %w", err)}
+	}
+	if existing := s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails; existing != nil && !reflect.DeepEqual(*existing, hardwareDetails) {
+		s.scope.Info("HardwareDetails changed", "old", existing, "new", hardwareDetails)
```
janiskemper: use an event here as well, otherwise you wouldn't be able to see this on the object-level

ashish1099:
we can either query vswitch api to get the details before populating the hardwaredetails (with internal ips) <- I havent did this but I think it should be doable.

or we simply reconcile to whatever node.status.address reports

guettli:
> or we simply reconcile to whatever node.status.address reports

@ashish1099 

> or we simply reconcile to whatever node.status.address reports

The task of the CSR is to check if the node can be trusted. When the node provides the IP list, which CSR will use to validate the node, then we do no need to make this check at all. I am open to discuss if we should drop the IP check completely. 

But checking input, which uses a list given by the input, makes no sense. Or I am missing something, then please correct me!

guettli:
> we can either query vswitch api to get the details before populating the hardwaredetails (with internal ips) <- I havent did this but I think it should be doable.

@ashish1099  Can you provide a PR?

janiskemper (dismissed):

ashish1099:
> > we can either query vswitch api to get the details before populating the hardwaredetails (with internal ips) <- I havent did this but I think it should be doable.
> 
> @ashish1099 Can you provide a PR?

https://github.com/Obmondo/cluster-api-provider-hetzner/pull/4/changes

I did a quick patch here.

the problem is on hetzner you can only setup the vswitch in the post/prekubeadmin stage and by then hardware details is given already, so we would never get those internal IP (if you are on vswithc)

I was tempted to use the vswitch api query but then appending the node spec just make it easy if there are more IPs

guettli:
> > > we can either query vswitch api to get the details before populating the hardwaredetails (with internal ips) <- I havent did this but I think it should be doable.
> > 
> > 
> > @ashish1099 Can you provide a PR?
> 
> https://github.com/Obmondo/cluster-api-provider-hetzner/pull/4/changes
> 
> I did a quick patch here.
> 
> the problem is on hetzner you can only setup the vswitch in the post/prekubeadmin stage and by then hardware details is given already, so we would never get those internal IP (if you are on vswithc)
> 


> I was tempted to use the vswitch api query but then appending the node spec just make it easy if there are more IPs

@ashish1099 have you understood why I don't want to read the node object, to validate the node object? If not, please try to undestand that first. Of course, this is easier. But if we would do that, then checking the IP makes no sense.

ashish1099:
> @ashish1099 have you understood why I don't want to read the node object, to validate the node object? If not, please try to undestand that first. Of course, this is easier. But if we would do that, then checking the IP makes no sense.

That make sense, so vswitch api query seems to be a correct way to me

pkg/services/baremetal/host/host.go
guettli: @janiskemper done in 12182623b

pkg/services/baremetal/host/host.go
```diff
@@ -698,13 +699,19 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
...
-	if s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails == nil {
-		hardwareDetails, err := getHardwareDetails(ctx, sshClient)
-		if err != nil {
-			return actionError{err: fmt.Errorf("failed to get hardware details: %w", err)}
+	hardwareDetails, err := getHardwareDetails(ctx, sshClient)
+	if err != nil {
+		return actionError{err: fmt.Errorf("failed to get hardware details: %w", err)}
+	}
+
+	if s.scope.HetznerBareMetalHost.Spec.Status.HardwareDetails != nil {
```
janiskemper, line 707: can we add a note here on the handling of the wwns? 

> In case of a change in the disks the wwns got updated. This might lead to an outdated "rootDeviceHint" which the user sets to decide which disk to use for provisioning, even if the server was already provisioned before. This will get caught by "validateRootDeviceHints" below.

pkg/services/baremetal/host/host_test.go
```diff
@@ -1665,6 +1665,80 @@ var _ = Describe("actionRegistering check RAID", func() {
...
+		service.actionRegistering(ctx)
+
+		var events []string
+		for len(testEventRecorder.Events) > 0 {
+			events = append(events, <-testEventRecorder.Events)
+		}
+		Expect(events).To(ContainElement(ContainSubstring("HardwareDetails Changed")))
+	})
+
+	It("still updates HardwareDetails when the hardware change makes RootDeviceHints invalid", func() {
```
janiskemper:
```suggestion
	It("invalidates RootDeviceHints in cases where a hardware change leads to different wwns", func() {
```

pkg/services/baremetal/host/host.go
guettli: Added a comment above the `HardwareDetails` assignment in `pkg/services/baremetal/host/host.go` explaining this.

done in 19d7f4b87

pkg/services/baremetal/host/host_test.go
guettli: done in 19d7f4b87

janiskemper (approved):

ashish1099:
@guettli thank you for getting this into master, waiting for the release cut :)

PR #2074, opened 2026-06-05
✨ v1beta2 migration: HCloudMachine

abdullah599:
**What this PR does / why we need it**:
This PR updates the HCloudMachine implementation and generated manifests for the new v1beta2 status shape.

**Which issue(s) this PR fixes:**
Fixes #2059

**Special notes for your reviewer**:
None

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [x] add unit tests

PR #2098, opened 2026-06-11
:bug: Rate limit in findServer marks ready machine as not existing

### LINKED ISSUE #2097: Rate limit error in `findServer` causes a ready machine to be incorrectly marked as no longer exists
When a fully-provisioned (`Ready=true`) HCloud machine encounters a rate limit error during reconciliation, the controller incorrectly concludes that the server no longer exists. This sets a `NoHCloudServerFound` machine error, marks `BootState=ProvisioningFailed`, and triggers remediation.

The root cause behind this is:
In `findServer` (https://github.com/syself/cluster-api-provider-hetzner/blob/7c53975feaae25a70310848326991e1a3b642053/pkg/services/hcloud/server/server.go#L2067), when `GetServer` returns a rate limit error, the code delegates to `handleRateLimit`:

```go
return nil, handleRateLimit(s.scope.HCloudMachine, err, "GetServer", errMsg)
```

`handleRateLimit` intentionally returns `nil` for ready machines without a deletion timestamp, to avoid noisy rate-limit conditions on healthy servers:

https://github.com/syself/cluster-api-provider-hetzner/blob/7c53975feaae25a70310848326991e1a3b642053/pkg/services/hcloud/server/server.go#L1266-L1267

This causes `findServer` to return `(nil, nil)` which is identical to the "server not found" return value. The reconcile loop then interprets this as the server having been deleted and triggers remediation.

**Proposed Fix**:
In `findServer`, return the error directly instead of delegating to `handleRateLimit` for the `GetServer` call. The rate limit is already handled correctly at the call site https://github.com/syself/cluster-api-provider-hetzner/blob/2c039c09d40e894acaa22c77209f22f293a5558c/pkg/services/hcloud/server/server.go#L151-L157
which sets the appropriate condition for non-ready machines and no-ops for ready machines

Dhairya-Arora01:
When GetServer returned a rate-limit error for a ready machine, handleRateLimit silently returned nil, causing findServer to return (nil, nil). The reconcile loop interpreted this as the server having been deleted and called SetErrorAndRemediate, setting BootState to ProvisioningFailed and annotating the CAPI machine for remediation.

Fix findServer to return the error directly instead of delegating to handleRateLimit, which was designed for provisioning-path calls. The rate-limit is already handled correctly at the reconcile-loop level, which no-ops for ready machines without setting any error conditions.

Fixes #2097

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -2064,7 +2064,7 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 			}
 
 			errMsg := fmt.Sprintf("failed to get server %d", serverID)
-			return nil, handleRateLimit(s.scope.HCloudMachine, err, "GetServer", errMsg)
+			return nil, fmt.Errorf("%s: %w", errMsg, err)
 		}
 
 		v1beta1conditions.MarkTrue(s.scope.HCloudMachine, infrav1.HCloudTokenAvailableCondition)
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -2063,8 +2063,7 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 				return nil, err
 			}
 
-			errMsg := fmt.Sprintf("failed to get server %d", serverID)
-			return nil, fmt.Errorf("%s: %w", errMsg, err)
+			return nil, fmt.Errorf("failed to get server %d: %w", serverID, err)
 		}
 
 		v1beta1conditions.MarkTrue(s.scope.HCloudMachine, infrav1.HCloudTokenAvailableCondition)
--- a/pkg/services/hcloud/server/server_test.go
+++ b/pkg/services/hcloud/server/server_test.go
@@ -428,6 +428,35 @@ var _ = Describe("findServer", func() {
 		Expect(hcloudClient.AssertExpectations(GinkgoT())).To(BeTrue())
 	})
 
+	It("returns rate-limit error for a ready machine", func() {
+		hcloudMachine := &infrav1.HCloudMachine{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "my-machine",
+				Namespace: "default",
+			},
+			Status: infrav1.HCloudMachineStatus{
+				Ready: true,
+			},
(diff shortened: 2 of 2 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -2064,7 +2064,7 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 			}
 
 			errMsg := fmt.Sprintf("failed to get server %d", serverID)
-			return nil, handleRateLimit(s.scope.HCloudMachine, err, "GetServer", errMsg)
+			return nil, fmt.Errorf("%s: %w", errMsg, err)
```
janiskemper: Can you add the previous line to this one? I think that would make it a bit easier to read.

pkg/services/hcloud/server/server_test.go
```diff
@@ -428,6 +428,35 @@ var _ = Describe("findServer", func() {
 		Expect(hcloudClient.AssertExpectations(GinkgoT())).To(BeTrue())
 	})
 
+	It("returns rate-limit error for a ready machine", func() {
```
janiskemper: can you explain why you test this?

You can also add a test that you can call reconcile on a ready server with rate limit and then that there is no error in the status and everything is good. 

This test would prevent the bug from happening in the future

Dhairya-Arora01:
```quote
You can also add a test that you can call reconcile on a ready server with rate limit and then that there is no error in the status and everything is good.

This test would prevent the bug from happening in the future
```

I already added that test (2nd test I added): https://github.com/syself/cluster-api-provider-hetzner/blob/c97825ea639ab390cf2ce68069610bbfeccecf46/pkg/services/hcloud/server/server_test.go#L1464-L1499

```quote
can you explain why you test this?

```

This test covers the specific fix — it ensures findServer always returns an error for rate limits on ready machines, not (nil, nil)

janiskemper: can you in both cases add a comment and explain the background of this test? (That's what I meant also before with regards to the first test)

janiskemper (approved):

janiskemper:
@abdullah599 FYI, this will conflict with #2074

PR #2101, opened 2026-06-12
:seedling: Emit events when HBMH provisioning fails

### LINKED ISSUE #2100: Emit events whenever HetznerBareMetalHost provisioning fails
When a `HetznerBareMetalHost` fails during provisioning , for example because the OCI image download is too slow then the reconciler logs an error and sets a condition, but no event with error is emitted. This makes it very hard to notice or diagnose the failure without looking into the controller logs as sometimes these errors are visible in the conditions for the slightest of second and then get reset / host starts reprovisioning.

The host then silently goes back to the `registering` phase with no user-visible signal.

**Proposal**:
Any failure that causes provisioning to abort should emit a warning event on the `HetznerBareMetalHost` object containing the actual error message.

Dhairya-Arora01:
Ensure that we emit events whenever HBMH provisioning fails.

Fixes #2100 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -720,6 +720,7 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
 			Reason:  infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason,
 			Message: infrav1.ErrorMessageMissingRootDeviceHints,
 		})
+		record.Warn(s.scope.HetznerBareMetalHost, "RootDeviceHintsValidationFailed", infrav1.ErrorMessageMissingRootDeviceHints)
 		return s.recordActionFailure(infrav1.RegistrationError, infrav1.ErrorMessageMissingRootDeviceHints)
 	}
 	errMsg := s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.IsValidWithMessage()
@@ -738,6 +739,7 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
 			Reason:  infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason,
 			Message: errMsg,
 		})
+		record.Warn(s.scope.HetznerBareMetalHost, "RootDeviceHintsValidationFailed", errMsg)
 		return s.recordActionFailure(infrav1.RegistrationError, errMsg)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -720,6 +720,7 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
 			Reason:  infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason,
 			Message: infrav1.ErrorMessageMissingRootDeviceHints,
 		})
+		record.Warn(s.scope.HetznerBareMetalHost, "RootDeviceHintsValidationFailed", infrav1.ErrorMessageMissingRootDeviceHints)
```
janiskemper: you changed the reason if I'm not mistaken here and below. The actual reason is "host validation failed"

pkg/services/baremetal/host/host.go
```diff
@@ -1764,6 +1778,7 @@ func (s *Service) createAutoSetupInput(ctx context.Context, sshClient sshclient.
 			Reason:  infrav1.HetznerBareMetalHostImageSpecInvalidV1Beta2Reason,
 			Message: errorMessage,
 		})
+		record.Warn(s.scope.HetznerBareMetalHost, "infrav1.HetznerBareMetalHostImageSpecInvalidV1Beta2Reason", errorMessage)
```
janiskemper: I think that this is not how you intended it

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
```diff
@@ -720,7 +720,7 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
 			Reason:  infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason,
 			Message: infrav1.ErrorMessageMissingRootDeviceHints,
 		})
-		record.Warn(s.scope.HetznerBareMetalHost, "RootDeviceHintsValidationFailed", infrav1.ErrorMessageMissingRootDeviceHints)
+		record.Warn(s.scope.HetznerBareMetalHost, infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason, infrav1.ErrorMessageMissingRootDeviceHints)
```
janiskemper, line 723: I think you are missing one in line 1981

Dhairya-Arora01: That is when `handleIncompleteBoot` fails and returns failed as true.
But `handleIncompleteBoot` itself already emits an error 
https://github.com/syself/cluster-api-provider-hetzner/blob/11836cc9a178d478daa9ab61f404e93105cf6da3/pkg/services/baremetal/host/host.go#L424-L425

and `failure = true` return only happens after that event is already emitted.

janiskemper (approved):

PR #2107, opened 2026-06-12
:bug: Fix stale HetznerCluster.Status causing cp targets to be skipped on LB attach

### LINKED ISSUE #2094: Fix stale LB target state in HetznerCluster.Status during rollout
**Context**:
When a controlplane machine (`HetznerBareMetalMachine` or `HCloudMachine`) is provisioned,
its [`reconcileLoadBalancerAttachment`](https://github.com/syself/cluster-api-provider-hetzner/blob/29317545cba5c199fff77ccae131a5ba99769ccb/pkg/services/baremetal/baremetal/baremetal.go#L618) checks `HetznerCluster.Status.ControlPlaneLoadBalancer.Target`
to decide whether the machine needs to be added as a target to the load balancer. This status field is only
updated when the `HetznerCluster` controller reconciles, which happens **independently** of the machine
controllers.

This causes problems during rollout:
Example: The target with ip x was removed from the LB and then again  the updated target with same ip x was being added but the stale HetznerCluster.Status shows that the target was already added so it was skipped, this leads to control-plane targets being not added to the load-balancer.

**Proposed Solution**:
In both `reconcileLoadBalancerAttachment` implementations, check whether `ServerAvailableCondition`
is already `True` on the machine:

- False/absent (first time): fetch the live LB state directly from the Hetzner API. This
  guarantees correctness regardless of stale status. Add the target if missing. When the function
  returns nil, `ServerAvailableCondition` gets set to `True` which signals that the target was
  confirmed in the real LB at least once.
- True (steady state): use `HetznerCluster.Status.ControlPlaneLoadBalancer.Target` as the
  cache. No extra API call is needed. If the target has been removed, `HetznerCluster.Status` will
  eventually reflect this (see below), and the add logic will re-run.

This ensures exactly one Hetzner API call for lb per machine lifecycle.

Also, add watches on `HetznerBareMetalMachine` and `HCloudMachine` in the `HetznerCluster` controller's
`SetupWithManager`. Each watch triggers a `HetznerCluster` reconcile when: 
- `ServerAvailableCondition` transitions to `True` (machine provisioned, target just added)
- The machine is deleted (target needs to be reflected as removed in status)

This ensures that the `HetznerCluster.Status` is not stale.

Dhairya-Arora01:
reconcileLoadBalancerAttachment in both HCloudMachine and HetznerBareMetalMachine services checks HetznerCluster.Status to decide whether a machine is already a load-balancer target. That status is only updated when the HetznerCluster controller reconciles, independently of the machine controllers. During a rollout this can produce a false positive: a target is removed from the LB and then re-added with the same IP/server-ID, but the stale status still shows it as present, so the attachment step is skipped and the control-plane node is never actually added back to the LB.

fixes done:

In both `reconcileLoadBalancerAttachment` implementations, check whether `ServerAvailableCondition` is already `True` on the machine:

- False/absent (first time): fetch the live LB state directly from the Hetzner API. This guarantees correctness regardless of stale status. Add the target if missing. When the function returns nil, `ServerAvailableCondition` gets set to `True` which signals that the target was confirmed in the real LB at least once.
- True (steady state): use `HetznerCluster.Status.ControlPlaneLoadBalancer.Target` as the cache. No extra API call is needed. If the target has been removed, `HetznerCluster.Status` will eventually reflect this (see below), and the add logic will re-run.

This ensures exactly one Hetzner API call for lb per machine lifecycle.

Also, add watches on `HetznerBareMetalMachine` and `HCloudMachine` in the `HetznerCluster` controller's `SetupWithManager`. Each watch triggers a `HetznerCluster` reconcile when:
- `ServerAvailableCondition` transitions to `True` (machine provisioned, target just added)
- The machine is deleted (target needs to be reflected as removed in status)

This ensures that the `HetznerCluster.Status` is not stale.

Fixes #2094 

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -620,13 +620,61 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
 		return nil
 	}
 
-	// check whether IPs of host have been added as load balancer targets already
+	// check whether IPs of host have been added as load balancer targets already.
 	var foundIPv4 bool
 	var foundIPv6 bool
 
-	for _, target := range s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.Target {
-		if target.Type == infrav1.LoadBalancerTargetTypeIP {
-			switch target.IP {
+	if v1beta1conditions.IsTrue(s.scope.BareMetalMachine, infrav1.ServerAvailableCondition) {
+		// If ServerAvailableCondition is set to true, then it means that this machine's ip was
+		// already successfully added as a target in the load balancer during a prior reconcile, this condition
+		// is only set to True after reconcileLoadBalancerAttachment returns nil.
+		// It is safe to use HetznerCluster.Status as a cache here because the HetznerCluster controller
+		// watches HetznerBareMetalMachine objects and triggers a reconcile on every ServerAvailableCondition
+		// marked as true transition and on machine deletion, keeping the target list up to date.
+		for _, target := range s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.Target {
+			if target.Type == infrav1.LoadBalancerTargetTypeIP {
+				switch target.IP {
+				case host.Spec.Status.IPv4:
+					foundIPv4 = true
+				case host.Spec.Status.IPv6:
+					foundIPv6 = true
+				}
+			}
+		}
+	} else {
+		// use the HCloud API.
+		clusterTagKey := s.scope.HetznerCluster.ClusterTagKey()
+		opts := hcloud.LoadBalancerListOpts{
+			ListOpts: hcloud.ListOpts{
+				LabelSelector: utils.LabelsToLabelSelector(map[string]string{
+					clusterTagKey: string(infrav1.ResourceLifecycleOwned),
+				}),
+			},
+		}
+
+		loadBalancers, err := s.scope.HCloudClient.ListLoadBalancers(ctx, opts)
+		if err != nil {
+			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "ListLoadBalancers")
+			return fmt.Errorf("failed to list load balancers: %w", err)
+		}
+
+		if len(loadBalancers) != 1 {
+			return fmt.Errorf("found %v loadbalancers in HCloud", len(loadBalancers))
+		}
+
+		lb := loadBalancers[0]
+
+		// This should never be the case: the label selector is cluster-scoped,
+		// so the only LB it can return is the one we own.
+		if lb.ID != s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ID {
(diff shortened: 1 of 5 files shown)


pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -620,13 +620,61 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
...
 
-	for _, target := range s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.Target {
-		if target.Type == infrav1.LoadBalancerTargetTypeIP {
-			switch target.IP {
+	if v1beta1conditions.IsTrue(s.scope.BareMetalMachine, infrav1.ServerAvailableCondition) {
+		// If ServerAvailableCondition is set to true, then it means that this machine's ip was
+		// already successfully added as a target in the load balancer during a prior reconcile, this condition
+		// is only set to True after reconcileLoadBalancerAttachment returns nil.
+		// It is safe to use HetznerCluster.Status as a cache here because the HetznerCluster controller
+		// watches HetznerBareMetalMachine objects and triggers a reconcile on every ServerAvailableCondition
```
janiskemper: I think this reasoning is not so great.. In the end, you want a reasoning why you do not call the API.

So I would actually write: "It is safe to use the status, even though it might be (slightly) outdated, because this is only a safeguard in case an unexpected change happened, e.g. by the user manually deleting a load balancer target. In 99% of the cases there is nothing to do anyway, so we can skip the extra API call to fetch the "true" load balancer targets."

pkg/services/baremetal/baremetal/baremetal_test.go
```diff
@@ -963,6 +981,46 @@ var _ = Describe("reconcileLoadBalancerAttachment", func() {
...
+
+		hetznerCluster := &infrav1.HetznerCluster{
+			Status: infrav1.HetznerClusterStatus{
+				ControlPlaneLoadBalancer: &infrav1.LoadBalancerStatus{
+					ID: 123,
+				},
+			},
+		}
+
+		// It should fallback to adding the load balancer target based on HetznerCluster status.
```
janiskemper: comment here that it should add the target to the load balancer via Hetzner API, but it should not call the endpoint to list load balancers and instead use the HetznerCluster status. 

I was confused about the fact that you add the mock here until I realized that it is for another function.

pkg/services/baremetal/baremetal/baremetal_test.go
```diff
@@ -1003,7 +1003,9 @@ var _ = Describe("reconcileLoadBalancerAttachment", func() {
 			},
 		}
 
-		// It should fallback to adding the load balancer target based on HetznerCluster status.
+		// It should fallback to adding the load balancer target via Hetzner API.
```
janiskemper: fallback is not a right expression here, is it?

janiskemper (approved):

PR #2108, opened 2026-06-13
✨ Add v1beta2 ServerAvailable condition to HetznerBareMetalMachine

### LINKED ISSUE #2110: Add v1beta2 `ServerAvailable` condition to HetznerBareMetalMachine
## Summary

HetznerBareMetalMachine sets a v1beta1 `ServerAvailable` condition but has no v1beta2 counterpart; add the v1beta2 `ServerAvailable` condition so it is set wherever the v1beta1 condition is set and is included in the v1beta2 `Ready` summary.

## The Problem

CAPH is migrating its conditions to the Cluster API v1beta2 format, so each resource maintains both v1beta1 conditions and their v1beta2 equivalents.

HetznerBareMetalMachine sets a v1beta1 `ServerAvailable` condition to mark a machine as available: `True` once a worker is provisioned, `True` for a control plane once it is attached to the load balancer, and `False` with reason `WaitingForAPIServer` while that attachment waits for a healthy API server. Unlike HCloudMachine, which has `ServerAvailable` in both versions, HetznerBareMetalMachine has no v1beta2 counterpart. 

As a result, the v1beta2 status is incomplete and its `Ready` summary ignores server availability. For control planes this lets v1beta2 `Ready` report ready before the load balancer gate clears, which is inconsistent with v1beta1 `Ready`.

> [!note]
> The v1beta1 `ServerAvailable` condition was added to HetznerBareMetalMachine in #1987 (addressing #1986), which gated bare metal control plane load balancer attachment on API server health. It reused the shared `ServerAvailable` constant directly in the controller code instead of being declared as a HetznerBareMetalMachine condition in `conditions_const.go`. Because it was never declared the standard way, it had no v1beta2 counterpart and was not picked up when the resource's v1beta2 conditions were set.

## The Solution

Add a v1beta2 `ServerAvailable` condition for HetznerBareMetalMachine that mirrors the v1beta1 condition.

abdullah599:
Added v1beta2 ServerAvailable condition to HetznerBareMetalMachine

Fixes #2110

### DIFF
--- a/pkg/services/baremetal/baremetal/baremetal.go
+++ b/pkg/services/baremetal/baremetal/baremetal.go
@@ -671,13 +671,13 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
 			infrav1.ServerAvailableCondition,
 			"WaitingForAPIServer",
 			clusterv1beta1.ConditionSeverityInfo,
-			"reconcile LoadBalancer: apiserver pod not healthy yet.",
+			"Waiting for API server pod to become healthy before attaching to load balancer",
 		)
 		v1beta2conditions.Set(s.scope.BareMetalMachine, metav1.Condition{
 			Type:    infrav1.HetznerBareMetalMachineServerAvailableV1Beta2Condition,
 			Status:  metav1.ConditionFalse,
 			Reason:  infrav1.HetznerBareMetalMachineWaitingForAPIServerV1Beta2Reason,
-			Message: "reconcile LoadBalancer: apiserver pod not healthy yet.",
+			Message: "Waiting for API server pod to become healthy before attaching to load balancer",
 		})
 		return &scope.RequeueAfterError{RequeueAfter: requeueAfter}
 	}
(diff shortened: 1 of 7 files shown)


janiskemper (commented):
do we not have an issue for this?

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -663,6 +673,12 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
 			clusterv1beta1.ConditionSeverityInfo,
 			"reconcile LoadBalancer: apiserver pod not healthy yet.",
 		)
+		v1beta2conditions.Set(s.scope.BareMetalMachine, metav1.Condition{
+			Type:    infrav1.HetznerBareMetalMachineServerAvailableV1Beta2Condition,
+			Status:  metav1.ConditionFalse,
+			Reason:  infrav1.HetznerBareMetalMachineWaitingForAPIServerV1Beta2Reason,
+			Message: "reconcile LoadBalancer: apiserver pod not healthy yet.",
```
Dhairya-Arora01: Lets make the message more user friendly. 
I would prefer "Waiting for API server pod to become healthy before attaching to load balancer"

abdullah599:
> do we not have an issue for this?

Created :)

Dhairya-Arora01 (approved):
LGTM

janiskemper (approved):

Issue #2092, opened 2026-06-10
v1beta2 migration: adopt CAPI-style import aliases for the CAPH API packages


abdullah599:
Part of #2017

## What would you like to be added?

Adopt the Cluster API import-alias convention for the two CAPH API packages, so the latest API version keeps a short, stable alias across future version bumps. Cluster API always aliases its latest core version as `clusterv1`, regardless of the actual beta suffix.

- alias `api/v1beta2` as `infrav1` (the current, latest version)
- alias `api/v1beta1` as `infrav1beta1` (the deprecated version, removed once v1beta1 support is dropped)

Apply this to every hand-written import of these two packages across the repo, and add an `importas` rule to `.golangci.yaml` so the convention is enforced and does not drift back. Generated files (for example conversion and mocks) keep the aliases their generators produce and stay excluded, since the linter already skips generated code.

PR #2109, opened 2026-06-13
✨ v1beta2 migration: switch the HetznerBareMetalHost controller to v1beta2

### LINKED ISSUE #2103: v1beta2 migration: switch the HetznerBareMetalHost controller to v1beta2

## Summary

Move the HetznerBareMetalHost controller, its host state machine and service, the host scope, and the machine side host logic to reconcile the v1beta2 type, and implement the behavior the new shape requires.

## What would you like to add

The API shape (#2063) is done. The v1beta2 host keeps its data in the real `status` subresource and drops the fields the host used to copy from its machine (`installImage`, `sshSpec`, `userData`, `hetznerClusterRef`) along with `errorCount`, `errorMessage`, `lastUpdated`, and `cpu.flags`. The controller still runs on v1beta1.

This issue switches the controller and the host logic to v1beta2. The CRD storage version stays v1beta1, so the apiserver converts. Flipping the storage version is a separate step.

PR #2242 splits the machine `installImage` into `installImage` and `customProvisioner`. The host reads the machine `installImage` live, so coordinate this switch with #2242: read `customProvisioner` live as well and pick the provisioning path by which field is set (see the Read inputs live row).

The work:

| Area | Change |
|---|---|
| Type switch | Reconcile the v1beta2 host across the controller, the host state machine and service, the scope, and the machine side host logic. `Spec.Status.X` becomes `Status.X` for the kept fields. Read `host.Namespace` instead of the dropped `consumerRef.Namespace`. |
| Read inputs live | Read `installImage` and `sshSpec` from the HetznerBareMetalMachine, `userData` from the CAPI Machine bootstrap, and the cluster from the cluster name label. These are no longer on the host. Safe because the machine webhook makes `installImage` and `sshSpec` immutable. With PR #2242 the machine splits provisioning into `installImage` and `customProvisioner`, so read `customProvisioner` live too, and pick installimage vs custom command by which of the two is set. |
| Provision and deprovision triggers | `installImage` on the host signaled both: present meant provision, absent meant deprovision. Both move to the machine state, no new field needed. The host provisions when `consumerRef` is set by a live machine (the machine only claims after its bootstrap ready gate) and deprovisions when that machine is being deleted. A new watch on HetznerBareMetalMachine drives both. |
| Requeue | Replace the `errorCount` backoff with static requeue tiers (30s reboot waits, 60s mid install, 5min user fix errors, stop on fatal and permanent). Drop the backoff helpers. |
| Error text | Add an `ActionCompleted` condition that carries the error text. `SetError` marks it false with the message, `ClearError` removes it. Migrate the `errorMessage` readers to `errorType` and the condition. |
| Conditions | Conditions become primary on `status.conditions`. Keep writing `status.deprecated.v1beta1.conditions` for the window. Move to the current patch and conditions helpers. Compute the `Ready` summary in the scope. |
| Tests | Enable the conversion webhook in the controller envtest, otherwise v1beta2 status writes are pruned under the v1beta1 schema. Assert both the native and the deprecated conditions. |

References:

* Parent issue: #2017.
* HetznerBareMetalHost v1beta2 API shape: #2063.
* HetznerBareMetalMachine installImage/customProvisioner split (affects the live read): PR #2242.

abdullah599:

This switches the HetznerBareMetalHost (HBMH) controller from the v1beta1 to the v1beta2.

Fixes #2103

*Package aliases.* We import v1beta1 as `infrav1` and v1beta2 as `infrav2`. Issue #2092 will flip these repo-wide to match CAPI (`infrav1beta1` for v1beta1, `infrav1` for v1beta2), so this PR keeps the current aliases and does not rename.

*Deprovision trigger.* `installImage` on the host signalled provision and deprovision before. Both come from the machine now, driven by a new watch on HetznerBareMetalMachine. The host deprovisions when the HetznerBareMetalMachine is being deleted, not when the CAPI Machine is being deleted. Reason being that CAPI marks the Machine for deletion, drains the node, and deletes the HetznerBareMetalMachine afterwards. Deprovisioning resets kubeadm and must not start while the node is still draining.

*Behaviour changes.* `errorCount` and `errorMessage` are gone from the shape. Recoverable errors requeue on static tiers in place of the exponential backoff, fatal and permanent errors stops the reconcile, and `SetError` records the message on the `ActionCompleted` condition. The dropped `spec.status` fields still round trip through the conversion annotation, but nothing writes them after this. 

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -599,6 +676,83 @@ func (host *HetznerBareMetalHost) HasHardwareReboot() bool {
...
+		Type:    HetznerBareMetalHostActionCompletedCondition,
+		Status:  status,
+		Reason:  reason,
+		Message: errorMessage,
+	})
+
+	// Clients that read the host through the v1beta1 API see the same condition in status.conditions.
+	// A v1beta1 condition carries a reason and a message only when it is False.
+	if status == metav1.ConditionTrue {
+		deprecatedv1beta1conditions.MarkTrue(host, ActionCompletedV1Beta1Condition)
```
janiskemper: until now we removed this condition and didn't set it to true

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -599,6 +676,83 @@ func (host *HetznerBareMetalHost) HasHardwareReboot() bool {
 	return false
 }
 
+// SetError sets the error type on the status and records it on the ActionCompleted condition. The
```
janiskemper: You currently remove the information from Status.ErrorMessage and Status.ErrorCount, don't you? Wouldn't we have to keep this information? If not, the conversion cannot work AFAIK.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
 	removed := removePermanentErrorIfAnnotationIsGone(bmHost)
 	if removed {
-		// The permanent error was removed from Spec.Status.
+		// The permanent error was removed from the status.
 		// Save the changes, and then reconcile again.
 		return reconcile.Result{Requeue: true}, nil
 	}
 
+	// Fetch the consuming HetznerBareMetalMachine and its CAPI Machine. The host both starts
+	// provisioning and deprovisions based on the machine, and reads its provisioning inputs
```
janiskemper: replace "machine" with the actual word (I think you mean HetznerBareMetalMachine)

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+	// (installImage, sshSpec, bootstrap data) from it.
+	var hetznerBareMetalMachine *infrav2.HetznerBareMetalMachine
+	var machine *clusterv1.Machine
+
+	if bmHost.Spec.ConsumerRef != nil {
+		// The consuming machine always lives in the namespace of the host.
```
janiskemper: also here let's be precise and not write "machine"

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+		if err := r.Get(ctx, name, hbmm); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return reconcile.Result{}, err
+			}
+			// The HetznerBareMetalMachine was force deleted. The host scope gets a nil
+			// HetznerBareMetalMachine and the host deprovisions with the robot-side cleanup only.
+		} else {
+			hetznerBareMetalMachine = hbmm
+
+			// If the owner Machine was force deleted while the hbmm lingers, keep reconciling with a
```
janiskemper: I don't fully understand this comment.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -233,54 +269,48 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 	// Case "Delete" was handled in reconcileSelectedStates. From now we know that the host has no
 	// DeletionTimestamp set. But the hbmm could be in Deprovisioning.
 
-	if bmHost.Spec.Status.HetznerClusterRef == "" {
-		log.Info("bmHost.Spec.Status.HetznerClusterRef is empty. Looks like a stale cache read")
+	if bmHost.Labels[clusterv1.ClusterNameLabel] == "" {
+		log.Info("HetznerBareMetalHost has no cluster-name label. Looks like a stale cache read")
```
janiskemper, line 273: AFAIK we prevent stale cache reads.

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -233,54 +269,48 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
 		return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
-	hetznerCluster := &infrav1.HetznerCluster{}
+	// Fetch the Cluster through the cluster-name label. The machine side sets the label when a
```
janiskemper: What does the second sentence mean?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -233,54 +269,48 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+	// machine takes the host.
+	cluster, err := util.GetClusterFromMetadata(ctx, r, bmHost.ObjectMeta)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
```
janiskemper, line 282: I believe that we actively deprovision the host when the Machine is not found. In case the Cluster or HetznerCluster is not found, we stop reconciling. 

This sounds inconsistent. Why exactly do we continue if the Machine is nil? I guess the main aspect would be that we manage to deprovision, but then this is something we could already handle for Cluster and HetznerCluster in a similar way...

However, in case the Cluster or HetznerCluster is already gone, we would end up not doing anything and return early.

I'm not sure what to do and we shouldn't spend a lot of time on this right now, but I just point this out and we can maybe think about our options.

janiskemper (changes requested):

api/v1beta2/hetznerbaremetalhost_types.go
abdullah599: You're right that we stop writing them. The conversion doesn't depend on that though. ConvertTo puts every spec.status field that v1beta2 doesn't have into the conversion data annotation and ConvertFrom reads it back, so errorCount and errorMessage survive the round trip

janiskemper: so you suggest to keep it in this way and it's on purpose, right?

abdullah599: yep

controllers/hetznerbaremetalhost_controller.go
abdullah599: We do, but only for what this controller writes itself. The cluster-name label comes from the HetznerBareMetalMachine controller

controllers/hetznerbaremetalhost_controller.go
abdullah599: host controller needs the HetznerCluster for the robot credentials. It builds the Robot client from hetznerCluster.Spec.HetznerSecret so we return early as host can not do anything. From the HetznerBareMetalMachine we don't need anything to deprovision so a missing one isnt a reason to stop

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -383,6 +387,79 @@ func (host *HetznerBareMetalHost) SetV1Beta1Conditions(conditions clusterv1.Cond
...
 
+// HetznerBareMetalHostSummaryOpts returns the summary options for a HetznerBareMetalHost.
+//
+// The order of conditions in ForConditionTypes defines the priority for the Ready summary:
+// when multiple conditions are unhealthy, the summary lists all of them in priority
+// order (highest-priority first). Credentials and provisioning problems must outrank
+// Deleting, since deletion may itself need credentials to succeed.
+//  1. RobotCredentialsAvailable - invalid Robot credentials block every Robot API call.
+//  2. ActionCompleted           - the action the host is running did not complete.
+//  3. RobotRateLimitExceeded    - rate-limit issues (negative polarity).
```
janiskemper, line 398: are you sure about ActionCompleted coming before RobotRateLimitExceeded? I think a rate limit would block other operations, so my guess would be that it should come first. That's just a guess though, you can check.

similar about ssh keys available or rootDeviceHints validated.

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -383,6 +387,79 @@ func (host *HetznerBareMetalHost) SetV1Beta1Conditions(conditions clusterv1.Cond
...
+			HetznerBareMetalHostRebootSucceededCondition,
+			HetznerBareMetalHostNodeBootIDRetrievedCondition,
+			HetznerBareMetalHostDeletingCondition,
+		},
+		// IgnoreTypesIfMissing tells the summary not to treat the absence of a
+		// listed condition as Unknown. Several reconcile paths exit before every
+		// condition has been set (for example, before Robot credentials are
+		// checked or before the host has been provisioned), and we don't want
+		// those early exits to flip Ready to Unknown.
+		conditions.IgnoreTypesIfMissing{
```
janiskemper, line 426: can you order this list same as above and same as the order of the conditions?

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -599,6 +676,84 @@ func (host *HetznerBareMetalHost) HasHardwareReboot() bool {
 	return false
 }
 
+// SetError sets the error type on the status and puts errorMessage on the ActionCompleted condition.
+// For a permanent error the message also names the annotation that an operator has to remove.
+// handleIncompleteBoot reads the error type back to pick the next reboot method.
```
janiskemper, line 681: let's remove the reference to another function that is in another package and not exported.

api/v1beta2/hetznerbaremetalhost_types.go
```diff
@@ -599,6 +676,84 @@ func (host *HetznerBareMetalHost) HasHardwareReboot() bool {
...
+	message := errorMessage
+	if errorType == PermanentError {
+		// A permanent error stays on the host until someone removes the annotation. The condition has
+		// to name the annotation.
+		message = fmt.Sprintf("%s. Remove annotation %q, if you want the controller to use the hbmh again.",
+			errorMessage, PermanentErrorAnnotation)
+	}
+
+	reason, v1beta1Reason := actionCompletedFor(errorType)
+	conditions.Set(host, metav1.Condition{
```
janiskemper, line 694: can you leave an empty line before this?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+		name := client.ObjectKey{
+			Namespace: bmHost.Namespace,
+			Name:      bmHost.Spec.ConsumerRef.Name,
+		}
+		if err := r.Get(ctx, name, hbmm); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return reconcile.Result{}, err
+			}
+			// The HetznerBareMetalMachine named by the consumerRef is already gone. The host scope
+			// gets a nil HetznerBareMetalMachine and the host deprovisions.
```
janiskemper: is that the case? What if I forcefully delete hbmm?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+		} else {
+			hetznerBareMetalMachine = hbmm
+
+			// The Machine is nil until CAPI sets the ownerRef on the HetznerBareMetalMachine, and
+			// if the Machine is force deleted. The host still has to deprovision and
+			// deprovisioning does not read the Machine, so it keeps reconciling with a nil Machine.
+			machine, err = util.GetOwnerMachine(ctx, r, hbmm.ObjectMeta)
+			if apierrors.IsNotFound(err) {
```
janiskemper, line 251: I think you can just negate this statement, as you do before. I believe the if else is not necessary

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -210,21 +210,57 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
...
+				machine = nil
+			} else if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	log = log.WithValues("HetznerBareMetalMachine", klog.KObj(hetznerBareMetalMachine))
```
janiskemper, line 259: this you can also do with the CAPI Machine. I'm not sure about how it handles nil. Can you ensure it works?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -482,51 +527,51 @@ func getAndValidateRobotCredentials(
...
-			Reason:  infrav1.HetznerBareMetalHostSecretUnreachableV1Beta2Reason,
-			Message: infrav1.ErrorMessageMissingHetznerSecret,
+			Reason:  infrav2.HetznerBareMetalHostSecretUnreachableReason,
+			Message: infrav2.ErrorMessageMissingHetznerSecret,
 		})
 
-		record.Warnf(bmHost, infrav1.HetznerSecretUnreachableReason, fmt.Sprintf("%s: %s", infrav1.ErrorMessageMissingHetznerSecret, err.Error()))
-		v1beta1conditions.SetSummary(bmHost)
-		scope.SetHetznerBareMetalHostV1Beta2ReadySummary(bmHost)
+		record.Warnf(bmHost, infrav2.HetznerSecretUnreachableV1Beta1Reason, fmt.Sprintf("%s: %s", infrav2.ErrorMessageMissingHetznerSecret, err.Error()))
```
janiskemper, line 551: I assume we don't wanna use the v1beta1 here when it is about the event. You can check this also in other places because we have that pattern elsewhere

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -598,27 +638,80 @@ func (r *HetznerBareMetalHostReconciler) SetupWithManager(ctx context.Context, m
...
+			newMachine, newOK := e.ObjectNew.(*infrav2.HetznerBareMetalMachine)
+			if !oldOK || !newOK {
+				return true
+			}
+
+			if oldMachine.DeletionTimestamp.IsZero() != newMachine.DeletionTimestamp.IsZero() {
+				return true
+			}
+
+			// The host annotation says which host this HetznerBareMetalMachine claimed. A change
```
janiskemper, line 657: not sure about any actual use cases for this. We would have to check exactly how hosts are being taken over. For example, we should check whether this is possible in the first place, and if so, whether we should reconcile both old and new host in this case.

pkg/scope/baremetalhost.go
```diff
@@ -164,72 +170,78 @@ func (s *BareMetalHostScope) GetRawBootstrapData(ctx context.Context) ([]byte, e
...
 
 func (s *BareMetalHostScope) hasConstantHostname() bool {
-	return s.Cluster.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true" ||
-		s.HetznerBareMetalMachine != nil && s.HetznerBareMetalMachine.GetAnnotations()[infrav1.ConstantBareMetalHostnameAnnotation] == "true"
+	return s.Cluster.GetAnnotations()[infrav2.ConstantBareMetalHostnameAnnotation] == "true" ||
+		s.HetznerBareMetalMachine != nil && s.HetznerBareMetalMachine.GetAnnotations()[infrav2.ConstantBareMetalHostnameAnnotation] == "true"
 }
 
 // SSHAfterInstallImageEnabled returns the effective SSH-after-installimage setting for the host.
+// When the consuming machine no longer exists, there is no SSH spec to connect with, so SSH access
```
janiskemper, line 187: let's never use "machine"

pkg/scope/baremetalhost.go
```diff
@@ -164,72 +170,78 @@ func (s *BareMetalHostScope) GetRawBootstrapData(ctx context.Context) ([]byte, e
...
+// is treated as disabled.
 func (s *BareMetalHostScope) SSHAfterInstallImageEnabled() bool {
```
janiskemper, line 189: just a note: IMO we shouldn't ever reach this place when the hbmm is nil, because this already means we are continuing to provision the host.

pkg/scope/baremetalhost.go
```diff
@@ -164,72 +170,78 @@ func (s *BareMetalHostScope) GetRawBootstrapData(ctx context.Context) ([]byte, e
...
-	return []v1beta1patch.Option{
-		v1beta1patch.WithOwnedConditions{Conditions: []clusterv1beta1.ConditionType{
-			clusterv1beta1.ReadyCondition,
-			infrav1.CredentialsAvailableCondition,
-			infrav1.RobotCredentialsAvailableCondition,
-			infrav1.RootDeviceHintsValidatedCondition,
-			infrav1.ProvisionSucceededCondition,
-			infrav1.HetznerAPIReachableCondition,
-			infrav1.ActionCompletedCondition,
+// BareMetalHostPatchOpts returns the patch options declaring both the deprecated v1beta1 and the
```
janiskemper, line 218: afaik this is only used in controllers right? Why don't we define it there? Maybe, similar to my other comment here, it is some refactoring effort we can write an issue for.

pkg/services/baremetal/host/state_machine.go
```diff
@@ -161,93 +162,99 @@ func (hsm *hostStateMachine) updateSSHKey() actionResult {
...
-		case infrav1.StateEnsureProvisioned:
+		case infrav2.StateEnsureProvisioned:
 			// Go back to StateImageInstalling as we need to provision again
-			hsm.nextState = infrav1.StateImageInstalling
-		case infrav1.StateProvisioned:
+			hsm.nextState = infrav2.StateImageInstalling
+		case infrav2.StateProvisioned:
 			errMessage := "secret has been modified although a provisioned machine uses it"
 			record.Event(hsm.host, "SSHSecretUnexpectedlyModified", errMessage)
-			return hsm.reconciler.recordActionFailure(infrav1.RegistrationError, errMessage)
```
janiskemper: why do we change the behavior here?

janiskemper, line 179: don't we drop the log that we had previously?

pkg/services/baremetal/host/host.go
```diff
@@ -713,110 +707,122 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
...
-			Reason:  infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason,
-			Message: infrav1.ErrorMessageMissingRootDeviceHints,
+			Reason:  infrav2.HetznerBareMetalHostValidationFailedReason,
+			Message: infrav2.ErrorMessageMissingRootDeviceHints,
 		})
-		record.Warn(s.scope.HetznerBareMetalHost, infrav1.HetznerBareMetalHostValidationFailedV1Beta2Reason, infrav1.ErrorMessageMissingRootDeviceHints)
-		return s.recordActionFailure(infrav1.RegistrationError, infrav1.ErrorMessageMissingRootDeviceHints)
+		record.Warn(s.scope.HetznerBareMetalHost, infrav2.HetznerBareMetalHostValidationFailedReason, infrav2.ErrorMessageMissingRootDeviceHints)
+		s.scope.HetznerBareMetalHost.SetError(infrav2.RegistrationError, infrav2.ErrorMessageMissingRootDeviceHints)
+		// The user has to specify root device hints. Check again in five minutes.
```
janiskemper, line 728: if the root device hints are missing, don't we enqueue anyway after the change of the spec? Do we need to requeue here?

pkg/services/baremetal/host/host.go
```diff
@@ -1277,11 +1264,13 @@ func (s *Service) actionImageInstalling(ctx context.Context) actionResult {
...
-		s.scope.HetznerBareMetalHost.SetError(infrav1.PermanentError, msg)
+		s.scope.HetznerBareMetalHost.SetError(infrav2.PermanentError, msg)
 		return actionStop{}
 	}
 
-	if s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.UsesImageURLCommand() {
+	// A HetznerBareMetalMachine sets exactly one of customProvisioner or installImage. When
+	// customProvisioner is set, provision with the custom command instead of installimage.
+	if s.scope.HetznerBareMetalMachine.Spec.CustomProvisioner != nil {
 		return s.actionImageInstallingImageURLCommand(ctx, sshClient)
```
janiskemper, line 1274: do we have an issue to rename all of this to custom provisioner?

pkg/services/baremetal/host/host.go
```diff
@@ -1426,35 +1416,36 @@ func (s *Service) actionImageInstallingImageURLCommand(ctx context.Context, sshC
...
+		return actionStop{}
 
 	case sshclient.ImageURLCommandStateNotStarted:
 		data, err := s.scope.GetRawBootstrapData(ctx)
 		if err != nil {
 			return actionError{err: fmt.Errorf("baremetal GetRawBootstrapData failed: %w", err)}
 		}
 
-		command := s.scope.HetznerBareMetalHost.Spec.Status.InstallImage.ImageURLCommand
+		command := s.scope.HetznerBareMetalMachine.Spec.CustomProvisioner.Command
```
janiskemper, line 1437: wouldn't this be validated by the webhooks?

janiskemper (changes requested):

api/v1beta2/hetznerbaremetalhost_types.go
abdullah599: i checked. This was basically copied from v1beta1. I think we can change it to:

1. RobotCredentialsAvailable
2. RobotRateLimitExceeded
3. SSHKeysAvailable
4. RootDeviceHintsValidated
5. ActionCompleted
6. ProvisionSucceeded
7. RebootSucceeded
8. NodeBootIDRetrieved
9. Deleting

The reason is that ActionCompleted usually repeats another condition from this list with a shorter message. When the ssh secret is invalid we mark SSHKeysAvailable false with the real error, and then SetError writes "invalid or not specified information in secret" on ActionCompleted

pkg/scope/baremetalhost.go
abdullah599: for the most part yes, but it does matter for actionDeprovisioning, because that one runs when the hbmm is gone

pkg/scope/baremetalhost.go
abdullah599: Note to self: I checked and all PatchOpts are present in pkg/scope so I should discuss what the best place for all would be.

api/v1beta2/hetznerbaremetalhost_types.go
janiskemper: sounds good

pkg/services/baremetal/host/host.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2282

pkg/scope/baremetalhost.go
janiskemper: you can add a comment to the if statement below that says this

controllers/hetznerbaremetalhost_controller.go
janiskemper: I assume you checked that it can handle nil values right?

abdullah599: yes. klog.KObj checks for a typed nil pointer and returns an empty ObjectRef. I ran it and the log line has Machine="" without a panic

controllers/hetznerbaremetalhost_controller.go
abdullah599: removed

pkg/scope/baremetalhost.go
abdullah599: Done

controllers/hetznerbaremetalhost_controller.go
abdullah599: We will use string literal as discussed

pkg/services/baremetal/host/state_machine.go
abdullah599: I added `SetHostError` on the scope which does SetError plus the log. all call sites that used recordActionFailure go through it now

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -128,29 +130,23 @@ func TestHetznerBareMetalHostReconciler_ReconcileSkipsPausedCluster(t *testing.T
...
 
-func verifyError(host *infrav1.HetznerBareMetalHost, errorType infrav1.ErrorType, errorMessage string) bool {
-	if host.Spec.Status.ErrorType != errorType {
-		return false
-	}
-	if host.Spec.Status.ErrorMessage != errorMessage {
-		return false
-	}
-	return true
+func verifyError(host *infrav2.HetznerBareMetalHost, errorType infrav2.ErrorType, _ string) bool {
```
janiskemper: why do we keep the string in the argument?

controllers/hetznercluster_controller_test.go
```diff
@@ -1708,6 +1709,9 @@ var _ = Describe("reconcileRateLimit", func() {
 		})
 		Expect(reconcileRateLimit(hetznerCluster, testEnv.RateLimitWaitTime)).To(BeFalse())
 		Expect(conditions.Has(hetznerCluster, infrav2.HCloudRateLimitExceededCondition)).To(BeFalse())
+		reachable := deprecatedv1beta1conditions.Get(hetznerCluster, infrav2.HetznerAPIReachableV1Beta1Condition)
```
janiskemper: don't we have the helper for this?

controllers/util.go
```diff
@@ -70,6 +70,22 @@ func reconcileRateLimit(obj conditionsObject, rateLimitWaitTime time.Duration) b
 	return false
 }
 
+// reconcileRobotRateLimit checks whether the Robot API rate limit has been reached and returns
+// whether the controller should wait a bit more. When the wait is over it clears the rate-limit
+// conditions (HetznerAPIReachable marked reachable again, RobotRateLimitExceeded deleted).
+func reconcileRobotRateLimit(bmHost *infrav2.HetznerBareMetalHost, rateLimitWaitTime time.Duration) bool {
+	condition := conditions.Get(bmHost, infrav2.HetznerBareMetalHostRobotRateLimitExceededCondition)
```
janiskemper: since we have "host" hardcoded here, I guess that we should move it to the host controller and if we use it somewhere else, we should double check

janiskemper (commented):
LGTM from my side. @Dhairya-Arora01 can you skim through and check the host_test.go in more detail?

controllers/hetznercluster_controller_test.go
abdullah599: isPresentAndTrueDeprecatedV1Beta1 does a testEnv.Get and this cluster only exists in memory, so it does not fit. I used deprecatedv1beta1conditions.IsTrue and IsFalse here as they also cover the nil case

controllers/util.go
abdullah599: moved

pkg/services/baremetal/host/host_test.go
```diff
@@ -2370,13 +2538,43 @@ var _ = Describe("actionProvisioned NoSSHAfterInstallImage=true", func() {
...
-		Expect(c.Status).To(Equal(corev1.ConditionTrue))
+		Expect(c.Status).To(Equal(metav1.ConditionTrue))
+		cV1Beta1 := deprecatedv1beta1conditions.Get(host, infrav2.RebootSucceededV1Beta1Condition)
+		Expect(cV1Beta1.Message).To(Equal(""))
+		Expect(cV1Beta1.Status).To(Equal(corev1.ConditionTrue))
 		Expect(host.GetAnnotations()).To(BeEmpty())
 	})
 })
 
+var _ = Describe("SetError and ClearError", func() {
```
Dhairya-Arora01: The block name is the same as the describe block on line 48

Dhairya-Arora01 (commented):
host_test.go looks good to me @janiskemper

janiskemper (approved):

PR #2113, opened 2026-06-22
:seedling: Enable proxy protocol for control plane load balancer

### LINKED ISSUE #2112: feat: Enable proxy protocol for control plane load balancer
## Summary

Add support for enabling proxy protocol on the control plane (kube-apiserver) load balancer service.

## Background

The Hetzner Cloud load balancer currently has proxy protocol hardcoded to `false`. Proxy protocol allows the LB to pass the original client IP to the backend, which is needed for accurate client IP logging and rate limiting at the kube-apiserver level.

## Scope

Proxy protocol is enabled **only** on the kube-apiserver LB service. Services listed under `spec.controlPlaneLoadBalancer.extraServices` are never affected — they always keep proxy protocol `false`.

## Design

Setting `spec.controlPlaneLoadBalancer.enableProxyProtocol: true` behaves differently depending on whether the cluster is new or already exists:

### New clusters

The kube-API LB service is created with proxy protocol **immediately** — no annotation check is performed. There are no pre-existing backends that could receive unexpected PROXY-protocol headers, so the safety gate is unnecessary.

### Existing clusters (migration path)

Proxy protocol is activated only when **both** conditions hold:

1. `spec.controlPlaneLoadBalancer.enableProxyProtocol: true` is set on the `HetznerCluster`
2. **Every** control-plane node in the workload cluster carries the annotation
   `capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true"`

The annotation is set by an **external service** (e.g. a node-preparation daemonset). CAPH reads it and never writes it. Enabling proxy protocol is a **one-way operation** — once switched on it is never turned back off.

### Why wait for all CP nodes on existing clusters?

Requiring every CP node to carry the annotation enables a **safe automated transition** from a cluster that does not support proxy protocol to one that does. If proxy protocol were enabled on the LB before all backends were ready, nodes still expecting plain TCP connections would receive malformed PROXY-protocol headers — breaking the control plane. By waiting for all nodes to signal readiness, CAPH ensures the LB is only switched once every backend can handle the new protocol.

### 1. New field in both v1beta1 and v1beta2

Add `enableProxyProtocol bool` to `LoadBalancerSpec` in both API versions (i.e. under `spec.controlPlaneLoadBalancer`). Since it is a plain optional bool it maps 1:1 between versions and the generated converter handles it directly.

### 2. New annotation constant

Add `ProxyProtocolForControlPlaneLoadBalancerAnnotation = "capi.syself.com/proxy-protocol-for-controlplane-loadbalancer"` in `api/v1beta2/hetznercluster_types.go`. CAPH reads this annotation from CP nodes on existing clusters — it never writes it. For new clusters this annotation is never read.

### 3. Effective proxy protocol computation (`controllers/hetznercluster_controller.go`)

Before every LB reconciliation, `computeEffectiveProxyProtocol` runs and sets `ClusterScope.EffectiveProxyProtocolForControlPlane`. This value is only used by the migration path:
- If the spec flag is `false` → effective is `false` (fast path, no WL cluster access)
- Otherwise, list all CP nodes (label `node-role.kubernetes.io/control-plane`) in the workload cluster and check that every one has the annotation
- If the WL cluster is not yet reachable, effective is `false` (retried on next reconcile)

### 4. Load balancer service reconciliation (`pkg/services/hcloud/loadbalancer/loadbalancer.go`)

- If the kube-API LB service **does not yet exist** (new cluster): create it with `enableProxyProtocol` from the spec directly — no annotation check.
- If the kube-API LB service **exists without proxy protocol** (migration): read `scope.EffectiveProxyProtocolForControlPlane`; when `true`, delete the service and recreate it with proxy protocol enabled. **Extra services always remain `false`.**
- Switching from `true` back to `false` is not supported (one-way operation).

### 5. v1beta1 ↔ v1beta2 conversion

The field exists in both versions so the generated converter handles it with a direct field assignment.

### 6. Generated files

Run `make generate` to regenerate CRD YAML, deepcopy, and conversion files.

## Acceptance Criteria

- [ ] `spec.controlPlaneLoadBalancer.enableProxyProtocol` field exists in both v1beta1 and v1beta2 CRDs
- [ ] When the field is `false` (default), LB behavior is unchanged
- [ ] **New cluster** with `enableProxyProtocol: true` → kube-API LB service is created with proxy protocol **immediately**, without waiting for node annotations
- [ ] **Existing cluster** with `enableProxyProtocol: true` but CP nodes lack the annotation → proxy protocol stays **disabled**
- [ ] **Existing cluster**: when the field is `true` and **all** CP nodes have the annotation → proxy protocol is **enabled** on the kube-API LB service and stays enabled permanently
- [ ] Extra LB services (`spec.controlPlaneLoadBalancer.extraServices`) are **never** given proxy protocol
- [ ] v1beta1 ↔ v1beta2 round-trip conversion preserves the field

guettli:
## Summary

- Adds \`spec.controlPlaneLoadBalancer.enableProxyProtocol\` (bool, default \`false\`) to \`HetznerCluster\`
- When \`true\`, enables PROXY protocol on the kube-apiserver LB service — extra services (\`spec.controlPlaneLoadBalancer.extraServices\`) are never affected

**New clusters** (kube-API LB service does not yet exist): the service is created with proxy protocol **immediately** — no annotation check is performed because there are no existing backends that could receive unexpected PROXY-protocol headers.

**Existing clusters** (migration path): CAPH waits until **all** control-plane nodes in the workload cluster carry the annotation \`capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: \"true\"\` before recreating the LB service with proxy protocol. The annotation is set by an external service (e.g. a node-configuration daemonset) — CAPH only reads it, never writes it. This prevents nodes still expecting plain TCP from receiving malformed PROXY-protocol headers during the transition.

- Once proxy protocol is enabled the kube-API LB service is **never turned back off** (one-way operation)
- CRD manifests regenerated

Closes #2112


pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +289,37 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
-				multierr = errors.Join(multierr, fmt.Errorf("failed to delete service from load balancer: %w", err))
-				if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
-					return multierr
-				}
+		if err := s.scope.HCloudClient.DeleteServiceFromLoadBalancer(ctx, lb, listenPort); err != nil {
+			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "DeleteServiceFromLoadBalancer")
+			multierr = errors.Join(multierr, fmt.Errorf("failed to delete service from load balancer: %w", err))
+			if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
+				return multierr
 			}
```
guettli, line 344: Hint for reviewers: `multierr` is returned at the end. It is not ignored.

api/v1beta1/types.go
```diff
@@ -213,6 +213,11 @@ type LoadBalancerSpec struct {
 
 	// Region contains the name of the HCloud location where the load balancer is running.
 	Region Region `json:"region,omitempty"`
+
+	// EnableProxyProtocol enables proxy protocol on the kube-apiserver load balancer service.
+	// See the v1beta2 field for full documentation.
```
janiskemper: let's put the docs in both places. I don't see a reason to not do that

controllers/hetznercluster_controller.go
```diff
@@ -302,6 +309,77 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	return reconcile.Result{}, nil
 }
 
+// computeEffectiveProxyProtocol sets clusterScope.EffectiveProxyProtocolForControlPlane.
```
janiskemper: this function is called for all clusters. The description here focuses on the edge case of migration. What does this function do for existing clusters that already have it enabled?

janiskemper (changes requested):

controllers/hetznercluster_controller.go
guettli: @janiskemper it does not distinguish between that case. It looks at the annotations in the cluster. If there are  none, then the value would be false. But this should not matter. Once the LB was created with that value, then it will not be removed again.

janiskemper: but shouldn't we look at clusters that have it already and skip it? This seems rather inefficient

guettli: First I thought we could use a Condition as a "Cache". But this would fail here:

- user sets proxyProcotol=True for a new cluster
- Because it is a new cluster, the nodes on wl-cluster don't need the annotation. And in this case don't have the anntotation.
- ProxyProtocol gets enabled 
- Everything works
- clusterctl move: Status gets lost
- caph reconciles, checks for annotation: Not found, does not enable proxyProtocol.

I need to think about that again.

guettli: I found a solution. Using status is too fragile, when `clusterctl move` make the status get lost.

@janiskemper please review again.

pkg/scope/cluster.go
```diff
@@ -267,3 +267,47 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
 	_, err = clientSet.Discovery().RESTClient().Get().AbsPath("/readyz").DoRaw(ctx)
 	return err
 }
+
+// AllControlPlaneNodesReadyForProxyProtocol returns true when every control-plane node
+// in the workload cluster carries the annotation
+// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true".
+// Returns false (no error) when the workload cluster is not yet reachable or has no nodes.
+func (s *ClusterScope) AllControlPlaneNodesReadyForProxyProtocol(ctx context.Context) (bool, error) {
```
janiskemper, line 276: question: The CAPI Machine objects also propagate information from the node. Why cannot we use the Machine objects to decide this?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +288,54 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	toCreate, toDelete := utils.DifferenceOfIntSlices(wantServiceListenPorts, slices.Collect(maps.Keys(existingServicesByPort)))
+
+	// Track whether the kube-API service already exists on the LB.
+	// New cluster: service absent → create immediately with EnableProxyProtocol from spec (no annotation check).
+	// Existing cluster migration: service present without proxy protocol → wait for all CP nodes to carry the
+	// annotation before recreating, to avoid sending malformed PROXY-protocol headers to unprepared backends.
+	_, kubeAPIServiceExists := existingServicesByPort[kubeAPIServicePort]
 
-	// delete services which are registered for lb but are not in specs
+	proxyProtocolShouldBeUsed, proxyProtocolAlreadyActive, err := s.proxyProtocolState(ctx, existingServicesByPort, kubeAPIServicePort)
```
janiskemper: you could just give "kubeAPIServiceExists" to this function instead of "existingServicesByPort" AFAIK

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -327,6 +355,25 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 	return multierr
 }
 
+// proxyProtocolState returns whether proxy protocol should be enabled now (shouldBeUsed)
+// and whether it is already active on the existing LB service (alreadyActive).
+// The workload cluster is only contacted when the spec wants proxy protocol but the LB
+// service doesn't have it yet. For new clusters or when already active, no call is made.
+func (s *Service) proxyProtocolState(ctx context.Context, existingServicesByPort map[int]hcloud.LoadBalancerService, kubeAPIServicePort int) (shouldBeUsed, alreadyActive bool, err error) {
+	if !s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.EnableProxyProtocol || kubeAPIServicePort == 0 {
```
janiskemper: you have a check here whether the port is 0. At the same time, in the parent function, you already define "kubeAPIServiceExists" by this time. 

Which meaning does it have if the port is 0? When can that happen? Can that happen at all? Why don't we have to look at that case in the parent function?

janiskemper (changes requested):

pkg/scope/cluster.go
guettli: @janiskemper afaik capi allows syncing metadata from the machine to the node. But there is not way to sync metadata or status from node to machine.

Only the proxy running in the wl-cluster knows if proxyProtocol is enabled or not. I see no way for it to update the capi machine in the mgt-cluster.

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: @janiskemper 

If you only passed kubeAPIServiceExists (bool), you'd lose access to the `Proxyprotocol` field, which is what determines the alreadyActive return value.

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: @janiskemper you are right the `|| kubeAPIServicePort == 0` was not needed here. I removed it.

pkg/scope/cluster.go
```diff
@@ -267,3 +267,47 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get workload cluster rest config: %w", err)
+	}
+	wlClient, err := client.New(restConfig, client.Options{})
+	if err != nil {
+		return false, fmt.Errorf("failed to create workload cluster client: %w", err)
+	}
+
+	nodeList := &corev1.NodeList{}
+	if err := wlClient.List(ctx, nodeList, client.MatchingLabels{"node-role.kubernetes.io/control-plane": ""}); err != nil {
```
janiskemper: I think we should import this label name from CAPI, shouldn't we?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -327,6 +355,25 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 	return multierr
 }
 
+// proxyProtocolState returns whether proxy protocol should get enabled now (shouldGetEnabled)
```
janiskemper: I think it would be easier to read if you inline this function. You can pretty much keep it as is with the explanations also.

janiskemper (changes requested):

pkg/scope/cluster.go
guettli: Done in 36c85a354 — now using `clusterv1.NodeRoleLabelPrefix + "/control-plane"`.

guettli: BTW, the whole string does not exist as constant:

```console
12:01 ❯ rg node-role.kubernetes.io/control-plane vendor
vendor/sigs.k8s.io/cluster-api/internal/webhooks/machine.go:149:20:     if taint.Key == "node-role.kubernetes.io/control-plane" {
vendor/sigs.k8s.io/cluster-api/test/e2e/kcp_md_rollout.go:126:13:           Key:    "node-role.kubernetes.io/control-plane",
vendor/sigs.k8s.io/cluster-api/test/e2e/clusterclass_rollout.go:233:13:         Key:    "node-role.kubernetes.io/control-plane",
vendor/sigs.k8s.io/cluster-api/test/framework/deployment_helpers.go:60:26:  nodeRoleControlPlane = "node-role.kubernetes.io/control-plane"
```

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: done in 3b815cb0d.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +288,68 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	// annotation before recreating, to avoid sending malformed PROXY-protocol headers to unprepared backends.
+	_, kubeAPIServiceExists := existingServicesByPort[kubeAPIServicePort]
+
+	// proxyProtocolShouldGetEnabled: whether proxy protocol should get enabled now.
+	// proxyProtocolAlreadyActive: whether it is already active on the existing LB service.
+	// The workload cluster is only contacted when the spec wants proxy protocol but the LB
+	// service doesn't have it yet. For new clusters or when already active, no call is made.
+	var proxyProtocolShouldGetEnabled, proxyProtocolAlreadyActive bool
+	if s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.EnableProxyProtocol {
+		if existing, ok := existingServicesByPort[kubeAPIServicePort]; ok {
```
janiskemper: this is the same code line as the previous one "_, kubeAPIServiceExists := existingServicesByPort[kubeAPIServicePort]" isn't it?

The below if statement can be written in an easier way and we can just define it like "proxyProtocolAlreadyActive := existing.ProxyProtocol"

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +288,68 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 			}
 		}
 	}
 
-	// create services which are in specs and not yet in API
+	// create services that are in the spec but not yet on the LB
 	for i, listenPort := range toCreate {
-		proxyProtocol := false
+		var proxyProtocol bool
+		if listenPort == kubeAPIServicePort {
```
janiskemper, line 351: add a comment here that we only activate proxy protocol for the kube api service

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +288,68 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+			if kubeAPIServiceExists {
+				// Migration path: kube-API service existed without proxy protocol; require all CP nodes
+				// to carry the annotation before enabling.
+				proxyProtocol = proxyProtocolShouldGetEnabled
```
janiskemper: I think we don't need this new variable and can just use and potentially overwrite this one here no? Why don't we even handle this whole if statement before where we also calculate proxyProtocolShouldGetEnabled?

janiskemper (changes requested):

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: done in 12900d962

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: @janiskemper added:

```go 
			// Proxy protocol is only relevant for the kube-apiserver port (default 6443).
```

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: done in eabb2c779

pkg/scope/cluster.go
```diff
@@ -267,3 +267,47 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
 	_, err = clientSet.Discovery().RESTClient().Get().AbsPath("/readyz").DoRaw(ctx)
 	return err
 }
+
+// AllControlPlaneNodesReadyForProxyProtocol returns true when every control-plane node
+// in the workload cluster carries the annotation
+// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true".
+// Returns false (no error) when the workload cluster is not yet reachable or has no nodes.
```
janiskemper: why do we do that? This function wouldn't get called for new clusters anyway because it is only called after the service for kubeapi server already exists on the load balancer.

For a running cluster this would be an actual error though.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +288,63 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	existingKubeAPIService, kubeAPIServiceExists := existingServicesByPort[kubeAPIServicePort]
+	proxyProtocolAlreadyActive := kubeAPIServiceExists && existingKubeAPIService.Proxyprotocol
+
+	// proxyProtocolShouldGetEnabled: whether proxy protocol should get enabled now.
+	// The workload cluster is only contacted when the spec wants proxy protocol but the LB
+	// service doesn't have it yet. For new clusters or when already active, no call is made.
+	var proxyProtocolShouldGetEnabled bool
+	if s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.EnableProxyProtocol && kubeAPIServiceExists && !proxyProtocolAlreadyActive {
+		var err error
+		proxyProtocolShouldGetEnabled, err = s.scope.AllControlPlaneNodesReadyForProxyProtocol(ctx)
```
janiskemper, line 309: we have to ensure that we reconcile whenever the nodes are finally ready. I think we have a similar case here in this issue, where we also want the HetznerCluster to reconcile based on something that happens on the node: https://github.com/syself/cluster-api-provider-hetzner/issues/2094

Basically, right now, the HetznerCluster would reconcile at a random timepoint and we are not sure when the proxy protocol would be finally enabled.

janiskemper (changes requested):

pkg/services/hcloud/loadbalancer/loadbalancer.go
janiskemper: addition: Since we already have the other issue (it's only merged in the release branch, be careful), we might not have to do much for this PR. We should document though that we actually reconcile when necessary so that proxy protocl would be enabled right after the control planes are up and running. 

What you can do if the other PR solves your problem: Just create it for the main branch as well

pkg/scope/cluster.go
guettli: done in b4d2a7ef4

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: @janiskemper I think this would add too much complexity. I implemented a simpler solution:

Fixed in dc6085156: reconcileServices now returns RequeueAfter: 10s when EnableProxyProtocol is true but not all CP nodes carry the annotation yet. So the controller polls every 10s until ready — no dependency on the 3-minute sync period.

janiskemper: @guettli in which way would that add more complexity? I didn't check in detail - but I believe whatever is implemented in that other PR should already be enough for this use case. So AFAIK (without checking the details) we don't have to do anything extra here and should just add a short comment explaining it

pkg/scope/cluster.go
janiskemper: we still ignore one error. Isn't that the same case?

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: @janiskemper The watch from PR #2107 fires when ServerAvailableCondition becomes True, but the proxy protocol annotation arrives later — after the node joins the cluster. So the reconcile triggered by that watch runs before the annotation is present, and nothing triggers a second one. The 10s requeue closes that gap with minimal code. Adding a watch on workload-cluster Node objects for the annotation would be the event-driven equivalent, but that's the extra complexity I wanted to avoid.

janiskemper: okay

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +290,67 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !proxyProtocolShouldGetEnabled {
```
janiskemper, line 313: you shouldn't requeue here but rather at the end after the reconcilement has happened.. You can just define another variable "shouldRequeue" and set it on true here.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -287,27 +290,67 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+				return reconcile.Result{}, multierr
 			}
 		}
 	}
 
-	// create services which are in specs and not yet in API
+	// create services that are in the spec but not yet on the LB
 	for i, listenPort := range toCreate {
-		proxyProtocol := false
+		var proxyProtocol bool
```
janiskemper: can you keep this  "proxyProtocol := false"? It shows that you want to set it on false rather defining it to set it later.

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: done in ea32ae7b0. Added test, to ensure other services get reconciled while waiting for all cp nodes to have the annotation.

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: done

pkg/scope/cluster.go
guettli: done in e68ab2a64

janiskemper (commented):
LGTM from my side. Let's wait for @Dhairya-Arora01

api/v1beta1/types.go
```diff
@@ -213,6 +213,23 @@ type LoadBalancerSpec struct {
...
+	// annotation check is performed because there are no existing backends that could receive
+	// unexpected PROXY-protocol headers.
+	//
+	// For existing clusters that want to enable proxy protocol after the fact, CAPH waits until
+	// every control-plane node carries the annotation
+	// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true" (set by an external
+	// service, never by CAPH) before recreating the LB service with proxy protocol. This prevents
+	// nodes still expecting plain TCP from receiving malformed PROXY-protocol headers.
+	//
+	// Enabling proxy protocol is a one-way operation — it is never turned back off.
```
Dhairya-Arora01, line 230: shouldn't we have a webhook validating this ?

guettli: @Dhairya-Arora01 yes, this makes sense. Change will be done soon.

guettli: @Dhairya-Arora01 it is implemented now as webhook for beta1. Beta2 webhooks are not called up to now.

janiskemper (dismissed):
plz also add this to v1beta2. Otherwise we will forget to add it later! Or what is your proposal?

janiskemper:
then @Dhairya-Arora01 plz have a look again!

Dhairya-Arora01 (approved):
LGTM

PR #2116, opened 2026-06-23
:seedling: test/e2e: cancel spec context on fast-fail to unblock waits

guettli:
## Summary

The CI failure on [#2113 CI Run](https://github.com/syself/cluster-api-provider-hetzner/actions/runs/27964284038/job/82904734813) was caused by a physical bare metal server failing to boot into rescue mode — completely unrelated to that PR's changes.

However, the failure revealed that the existing **fail-fast mechanism** in `logStatusContinuously` isn't actually fast. When it detects a `PermanentError` on a `HetznerBareMetalHost`, it calls `Fail()` from the goroutine. That records the Ginkgo failure and exits the goroutine, but the main `It` goroutine keeps blocking on `WaitForControlPlaneToBeReady` until the full **1200-second timeout** — wasting ~10 minutes of CI time per occurrence.

### Fix

- Add a `specCancelFn` (protected by `specCancelMu`) in `caph.go`, updated per-spec via `setSpecCancel`/`clearSpecCancel`.
- In `CaphClusterDeploymentSpec`'s `It` block, create a cancellable child context (`specCtx`) and register its cancel. Pass `specCtx` to `ApplyClusterTemplateAndWait`.
- In `logStatusContinuously`, call `fn()` (cancel the spec context) **before** `Fail()` so `WaitForControlPlaneToBeReady` aborts immediately.
- `AfterEach` still uses the original `ctx`, so cleanup is unaffected.

### Result

When hardware reboots fail (PermanentError after ~10 min), the test now aborts within ~30 seconds of detection instead of waiting the full 1200-second timeout.

## Test plan

- [ ] CI baremetal test passes on healthy hardware (context cancellation is a no-op in the happy path)
- [ ] On hardware failure, test fails fast (~30s after PermanentError is set) instead of waiting 20 minutes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/e2e/caph.go
+++ b/test/e2e/caph.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


Dhairya-Arora01 (commented):
@guettli as far as I have checked, the cluster api helpers etc. that we are using in e2e tests won't return immediately when we cancel the context but insted the individual API checks in it would fail faster, e.g. is the cluster ready?" comes back with an error immediately instead of taking a second but the loop that keeps re-checking would keep running.

PR #2124, opened 2026-06-25
:seedling: fix: make installImage.partitions optional, when imageURLCommand gets used.

guettli:
## Summary

- `partitions` is completely ignored when `imageURLCommand` is set — the controller takes a different code path that never builds an installimage config. The CRD previously marked the field required in all cases, forcing users to supply dummy partition data for imageURLCommand-based provisioning.
- Make `partitions` optional in the API types (`omitempty`), dropping it from the CRD `required` list.
- Add webhook validation: partitions must be non-empty when `imageURLCommand` is **not** set (installimage needs at least one partition). When `imageURLCommand` **is** set, partitions are optional — they're ignored by the controller, but may still be set if desired.
- Update tests in both v1beta1 and v1beta2 webhook packages.


Dhairya-Arora01 (approved):
LGTM

api/v1beta1/hetznerbaremetalmachine_types.go
```diff
@@ -203,7 +203,9 @@ type InstallImage struct {
 	PostInstallScript string `json:"postInstallScript,omitempty"`
 
 	// Partitions define the additional Partitions to be created in installimage.
-	Partitions []Partition `json:"partitions"`
+	// Must be non-empty when imageURLCommand is not set.
```
janiskemper, line 206: we now call it "customProvisioner". Can you update it everywhere? If this PR is rebased, it should contain already the API changes and the new name.

janiskemper (changes requested):

PR #2125, opened 2026-06-26
fix: Do not set RebootTriggeredAt to nil too early

guettli:
## Summary

`RebootTriggeredAt` was set to nil to early.

Then `hasJustRebooted()` returned `true` forever when `RebootTriggeredAt` is nil but `ErrorType` is `SSHRebootTriggered`. 

This was introduced in 4f9d7c6bfeefd90059615beaf71699bfe852b723 in April this year:

```diff 
        if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType == infrav1.ErrorTypeConnectionError {
                s.scope.HetznerBareMetalHost.ClearError()
+               s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = nil
        }
```

## How this state is reached

The exact path that produces `ErrorType=SSHRebootTriggered` with `RebootTriggeredAt=nil`:

1. During a reboot, the server becomes temporarily unreachable → `handleIncompleteBoot` sets `ErrorType=ConnectionError`, `RebootTriggeredAt=<time>`.
2. On the next reconcile the server is reachable again → lines 435–437 clear the connection error **and set `RebootTriggeredAt=nil`**.
3. SSH still times out (server hasn't fully rebooted yet) → the `emptyErrorType` case sets `ErrorType=SSHRebootTriggered` but (before this fix) never re-sets `RebootTriggeredAt`.
4. `hasJustRebooted()` sees `SSHRebootTriggered` + nil timestamp → `hasTimedOut(nil, …)` returns `false` → `!false == true` → stuck forever.

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -434,7 +434,6 @@ func (s *Service) handleIncompleteBoot(ctx context.Context, isRebootIntoRescue, 
 	// ssh gave no connection refused error but it is still saved in host status - we can remove it
 	if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType == infrav1.ErrorTypeConnectionError {
 		s.scope.HetznerBareMetalHost.ClearError()
-		s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = nil
 	}
 
 	// Check whether there has been an error message already, meaning that the reboot did not finish in time.
@@ -447,9 +446,6 @@ func (s *Service) handleIncompleteBoot(ctx context.Context, isRebootIntoRescue, 
 			// A timeout error from SSH indicates that the server did not yet finish rebooting.
 			// As the server has no error set yet, set error message and return.
 			s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeSSHRebootTriggered, "ssh timeout error - server has not restarted yet")
-			if s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt == nil {
-				s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
-			}
 			return false, nil
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -447,6 +447,9 @@ func (s *Service) handleIncompleteBoot(ctx context.Context, isRebootIntoRescue,
 			// A timeout error from SSH indicates that the server did not yet finish rebooting.
 			// As the server has no error set yet, set error message and return.
 			s.scope.HetznerBareMetalHost.SetError(infrav1.ErrorTypeSSHRebootTriggered, "ssh timeout error - server has not restarted yet")
+			if s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt == nil {
```
janiskemper: IMO the solution should be rather to not remove the timestamp when there is no connection error, but still a timeout error. At this point we remove an error and replace it here with another error. But since we are still in a reboot process, I don't think we should in between remove the timestamp at all. 

Instead, we should only remove it when it is clear that the ssh connection was successfully established. 

Can you adapt the code accordingly? 

Since there are various occasions of handleIncompleteBoot you would have to be careful to find all occurrences!

janiskemper (changes requested):

pkg/services/baremetal/host/host.go
guettli: This was introduced in 4f9d7c6bfeefd90059615beaf71699bfe852b723 in April this year:

```diff 
        if s.scope.HetznerBareMetalHost.Spec.Status.ErrorType == infrav1.ErrorTypeConnectionError {
                s.scope.HetznerBareMetalHost.ClearError()
+               s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = nil
        }
```

I guess this was not really on purpose.

I will remove that.

guettli: @janiskemper done in 38b7c06371a8d50694ae94ef381acc94dca8cb02. 

There are only two places calling handleIncompleteBoot(). Both set RebootTriggeredAt to nil, when ssh works. No update was needed.

janiskemper:
@guettli can you update the PR title and description again and also add a comment to the utility function you updated? Basically, I would say that this PR is about fixing a bug where we set RebootedAt too early to nil. 

Keeping the change in the utility function wouldn't be necessary but might be a safe guard for the future.

guettli:
> @guettli can you update the PR title and description again and also add a comment to the utility function you updated? Basically, I would say that this PR is about fixing a bug where we set RebootedAt too early to nil.
> 
> Keeping the change in the utility function wouldn't be necessary but might be a safe guard for the future.

@janiskemper

PR title and description got updated.

In ec53299cb I added a comment and log, when the value is nil, which is not expected.

janiskemper (approved):

guettli:
Cherry-picked into release-1.1 via 240eaa585

PR #2128, opened 2026-06-29
:bug: Fix stale HetznerCluster.Status causing cp targets to be skipped on LB attach

### LINKED ISSUE #2094: Fix stale LB target state in HetznerCluster.Status during rollout

guettli:
## Summary

Port of #2107 to main.

- When `ServerAvailableCondition` is not yet `True`, fetch live load-balancer targets via the HCloud API instead of relying on the potentially-stale `HetznerCluster.Status` cache — preventing a target that was removed and re-added with the same IP/server-ID from being skipped due to a stale "already attached" status.
- Once `ServerAvailableCondition` is `True`, use the cached status instead, avoiding an extra HCloud API call on every reconcile.
- Add `Watches` for `HetznerBareMetalMachine` and `HCloudMachine` in the `HetznerCluster` controller, scoped to control-plane machines, so the cluster reconciles promptly when `ServerAvailableCondition` transitions to `True` (target just added) or when a control-plane machine is deleted (target needs to be reflected as removed in status).

Closes https://github.com/syself/cluster-api-provider-hetzner/issues/2094


guettli:
First version was wrong. It looked beta1 conditions. PR will be updated soon ... (done)

Dhairya-Arora01:
> First version was wrong. It looked beta1 conditions. PR will be updated soon ... (done)

Doesn't it still looks for v1beta1 condition  @guettli  ?

guettli:
> > First version was wrong. It looked beta1 conditions. PR will be updated soon ... (done)
> 
> Doesn't it still looks for v1beta1 condition @guettli ?

yes, there went something wrong yesterday. I changed that. But I am still a bit unsure. Looking at:

```go

	if v1beta2conditions.IsTrue(s.scope.BareMetalMachine, infrav1.HetznerBareMetalMachineServerAvailableV1Beta2Condition)
```

`s.scope.BareMetalMachine` is beta1. Is using ` v1beta2conditions.IsTrue()`  correct?

Dhairya-Arora01:
> > > First version was wrong. It looked beta1 conditions. PR will be updated soon ... (done)
> > 
> > 
> > Doesn't it still looks for v1beta1 condition @guettli ?
> 
> yes, there went something wrong yesterday. I changed that. But I am still a bit unsure. Looking at:
> 
> ```go
> 	if v1beta2conditions.IsTrue(s.scope.BareMetalMachine, infrav1.HetznerBareMetalMachineServerAvailableV1Beta2Condition)
> ```
> 
> `s.scope.BareMetalMachine` is beta1. Is using ` v1beta2conditions.IsTrue()` correct?

yes as it will set the condition in status.v1beta2

Dhairya-Arora01 (commented):
LGTM

controllers/hetznercluster_controller.go
```diff
@@ -1017,3 +1027,79 @@ func IgnoreInsignificantHetznerClusterStatusUpdates(logger logr.Logger) predicat
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldGetter, ok := e.ObjectOld.(conditions.Getter)
+			if !ok {
+				return false
+			}
+			newGetter, ok := e.ObjectNew.(conditions.Getter)
+			if !ok {
+				return false
+			}
+			wasTrue := conditions.IsTrue(oldGetter, infrav2.HCloudMachineServerAvailableCondition)
```
abdullah599: This reads HCloudMachineServerAvailableCondition, but the same predicate also runs for the HetznerBareMetalMachine watch. It works only because both constants are the string ServerAvailable. If that string ever changes, the baremetal watch breaks with no compile error. Can we add infrav2.HetznerBareMetalMachineServerAvailableCondition and use it for the baremetal watch?

controllers/hetznercluster_controller.go
```diff
@@ -1017,3 +1027,79 @@ func IgnoreInsignificantHetznerClusterStatusUpdates(logger logr.Logger) predicat
...
+
+	return []reconcile.Request{{
+		NamespacedName: client.ObjectKey{Namespace: hm.Namespace, Name: cluster.Spec.InfrastructureRef.Name},
+	}}
+}
+
+// controlPlaneMachineToHetznerClusterPredicate returns a predicate that fires only when:
+//   - a machine is deleted (so the HetznerCluster can update its LB target status), or
+//   - ServerAvailableCondition transitions to True.
+func controlPlaneMachineToHetznerClusterPredicate() predicate.Funcs {
```
abdullah599, line 1067: Question: This does not filter to control plane machines so why do we have controlplane in name?

abdullah599 (dismissed):
minor nits else LGTM :)

controllers/hetznercluster_controller.go
guettli: @abdullah599 

Good catch, fixed. The predicate now reads the condition type per watched object kind: `infrav1.HCloudMachineServerAvailableV1Beta2Condition` for `HCloudMachine`, `infrav1.HetznerBareMetalMachineServerAvailableV1Beta2Condition` for `HetznerBareMetalMachine`. These are the actual v1beta1 constants that `server.go`/`baremetal.go` use when they set the condition, so the predicate now reads exactly what gets written, instead of comparing against a value that only happened to match by coincidence.

(We first tried adding a parallel `infrav2.HetznerBareMetalMachineServerAvailableCondition` constant, but nothing sets that one either — same problem in a new place. Pointing the predicate straight at the v1beta1 constants that are actually written avoids that.)

controllers/hetznercluster_controller.go
guettli: You're right, and thinking about it more: LB targets can only ever be control-plane nodes, so the predicate
should only fire for control-plane machines in the first place. Updated it to actually filter on
`clusterv1.MachineControlPlaneLabel`, read directly off the watched `HCloudMachine`/`HetznerBareMetalMachine`
object — CAPI already propagates that label (and `clusterv1.ClusterNameLabel`, which this file already
relies on) from the owning `Machine` onto the infra machine at creation time, so no extra `Get` is needed.
Worker machine events (both updates and deletes) are now filtered out entirely instead of causing a
no-op `HetznerCluster` reconcile. Added `TestControlPlaneMachineToHetznerClusterPredicate` to cover the
control-plane vs. worker and HCloudMachine vs. HetznerBareMetalMachine cases.

Also removed `infrav2.HCloudMachineServerAvailableCondition` from `api/v1beta2/conditions_const.go` (added
in #2074, the HCloudMachine v1beta2 migration): it became unused once the predicate switched to the
v1beta1 constants above, since this predicate was its only caller.

abdullah599 (commented):
LGTM

controllers/hetznercluster_controller.go
```diff
@@ -1017,3 +1028,105 @@ func IgnoreInsignificantHetznerClusterStatusUpdates(logger logr.Logger) predicat
...
+		return nil
+	}
+
+	return []reconcile.Request{{
+		NamespacedName: client.ObjectKey{Namespace: bm.Namespace, Name: cluster.Spec.InfrastructureRef.Name},
+	}}
+}
+
+// hcloudMachineToHetznerCluster maps an HCloudMachine to the owning HetznerCluster.
+func (r *HetznerClusterReconciler) hcloudMachineToHetznerCluster(ctx context.Context, o client.Object) []reconcile.Request {
```
abdullah599: this is identical to baremetalMachineToHetznerCluster except the type assert, which only feeds the nil guard. can we merge into one machineToHetznerCluster(ctx, o client.Object) that reads o.GetLabels()/o.GetNamespace(), and point both watches at it

abdullah599 (dismissed):
minor nit

controllers/hetznercluster_controller.go
guettli: @abdullah599 

done in c30e4e8c297dd4ca888e7f9fb3729e08eaff5060.

Please approve again.

abdullah599 (approved):

PR #2129, opened 2026-06-29
:seedling: Read installImage value from hbmm spec, not hbmh status.

guettli:
Read installImage value from hbmm spec, not hbmh status.


guettli:
Will be done later. Part of beta2: https://github.com/syself/cluster-api-provider-hetzner/issues/2103

PR #2130, opened 2026-06-29
✨ DeviceStringType field for ImageURLCommand (sda vs WWN)

### LINKED ISSUE #2127: `DeviceStringType` field for `ImageURLCommand` (sda vs WWN)
# Plan: `DeviceStringType` field for `ImageURLCommand`

`DeviceStringType` is a baremetal-only feature. HCloud always passes `sda`; WWN is not supported
there.

## Why?

The user configures the rootDeviceHints in the hbmh spec. This is done via WWN, not device names
like "sda".

CAPH currently translates the WWN to a device name. This is fragile, because device names can
change when a Linux server reboots. The disk called "sda" before, might be called "sdb" after a
reboot.

To eliminate sources of failures, we want to send the value of the customer (WWN) to the
imageURLCommand.

---

### Baremetal path

`host.go:1429` calls `getDeviceNames(wwns, storageDevices)`, which finds storage devices whose
WWN matches a root-device-hint WWN and returns their **short names** (e.g., `sda`). Those short
names are passed as the last positional argument to `sshClient.StartImageURLCommand`.

We want to make it possible that this conversion from WWN to device name gets done in CAPH.

### HCloud path

`server.go:897` hardcodes `[]string{"sda"}`. HCloud VMs always boot from `sda`; this does not
change. Disks of hcloud machines do not have a WWN.

### SSH layer

`ssh_client.go:863–867`: device names are written into the shell command as the 4th positional
argument (space-joined):

Example:

```bash
nohup /root/image-url-command '<imageURL>' /root/bootstrap.data '<machineName>' 'sda'
```

With `DeviceStringType: "wwn"` on baremetal the argument becomes `'eui.00253885910c8cec'` or
similar.

---

### 1. API types — baremetal only

Add to `api/v1beta2/hetznerbaremetalmachine_types.go` (`InstallImage`):

```go
// DeviceStringType instructs caph to either use the short device name, or the WWN when calling
// ImageURLCommand. It is not used when ImageURLCommand is not set. Allowed values are "short"
// or "wwn". Example: When "short" is used, ImageURLCommand receives "sda"; when "wwn" is
// used, it receives "eui.00253885910c8cec" or "0x500a07511bb48b25".
DeviceStringType string `json:"deviceStringType,omitempty"`
```

Add the same field to beta1.

### 2. Baremetal logic — `host.go`

At line 1429 (in `actionImageInstallingImageURLCommand()`), switch on `DeviceStringType` (two call
sites: lines 1429 and 1819):

```go
var deviceNames []string
switch s.scope.HetznerBareMetalMachine.Spec.InstallImage.DeviceStringType {
case "wwn":
    deviceNames = s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.ListOfWWN()
default: // "short" or ""
    deviceNames = getDeviceNames(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.ListOfWWN(), storage)
}
```

No new helper function needed — the WWNs are already available from `RootDeviceHints`.

### 3. CEL validation (no webhook needed)

Use a CEL validation rule on the field instead of webhook validation — CEL is sufficient for this constraint.

Note: do not use an enum type for `DeviceStringType`. According to the [Kubernetes API
conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-and-constants),
string enums should be avoided in favour of plain strings with documented allowed values. The
allowed values (`""`, `"short"`, `"wwn"`) are expressed in the field comment and enforced via a CEL
rule.

### 4. CRD regeneration

```
make generate
```

Regenerates:

- `api/v1beta1/zz_generated.conversion.go` (picks up the new field automatically)
- `config/crd/bases/` YAML files

### 5. Tests

- `pkg/services/baremetal/host/host_test.go` — cases for `DeviceStringType: "wwn"` verifying the
  correct WWN is passed to `StartImageURLCommand`.
- CRD/CEL validation test — valid and invalid `DeviceStringType` values (e.g. rejected unknown
  value, accepted `""`, `"short"`, `"wwn"`).

### 6. Docs

Update `docs/caph/04-developers/06-image-url-command.md` to explain the `DeviceStringType` field
and its allowed values.

guettli:
Closes #2127

## Summary

* Adds `DeviceStringType` Go type (named string) with constants `DeviceStringTypeShort` and `DeviceStringTypeWWN` to both `api/v1beta1` and `api/v1beta2`
* Adds `DeviceStringType DeviceStringType` (JSON: `deviceStringType`) with a CEL validation rule restricting values to `""`, `"short"`, and `"wwn"`
* Implements the switch in `actionImageInstallingImageURLCommand`: when `DeviceStringType == "wwn"`, passes WWNs directly from `RootDeviceHints` to `StartImageURLCommand` instead of resolving to short device names
* Regenerates CRDs and conversion code via `make generate`
* Adds a test for the `"wwn"` path verifying the correct WWN is passed to `StartImageURLCommand`
* Updates docs explaining `deviceStringType` and why WWNs are more stable than short device names

### DIFF
--- a/api/v1beta1/hetznerbaremetalmachine_types.go
+++ b/api/v1beta1/hetznerbaremetalmachine_types.go
@@ -47,11 +47,15 @@ var errUnknownSuffix = errors.New("unknown suffix")
 type ImageType string
 
 // DeviceStringType controls what CAPH passes as the device argument to ImageURLCommand.
-// The zero value ("") passes the short device name (e.g. "sda"). Set to "wwn" to pass the WWN instead.
+// Allowed values are "" (same as "short"), "short", and "wwn".
 type DeviceStringType string
 
-// DeviceStringTypeWWN passes the WWN (e.g. "eui.00253885910c8cec") to ImageURLCommand.
-const DeviceStringTypeWWN DeviceStringType = "wwn"
+const (
+	// DeviceStringTypeShort passes the short device name (e.g. "sda") to ImageURLCommand.
+	DeviceStringTypeShort DeviceStringType = "short"
+	// DeviceStringTypeWWN passes the WWN (e.g. "eui.00253885910c8cec") to ImageURLCommand.
+	DeviceStringTypeWWN DeviceStringType = "wwn"
+)
 
 const (
 	// ImageTypeTar defines the image type for tar files.
@@ -187,11 +191,11 @@ type InstallImage struct {
 	// +optional
 	ImageURLCommand string `json:"imageURLCommand,omitempty"`
 
-	// DeviceStringType instructs caph to pass the short device name (default, "") or the WWN when
-	// calling ImageURLCommand. It is not used when ImageURLCommand is not set. Example: when "" is
-	// used, ImageURLCommand receives "sda"; when "wwn" is used, it receives "eui.00253885910c8cec"
-	// or "0x500a07511bb48b25".
-	// +kubebuilder:validation:XValidation:rule="self == '' || self == 'wwn'",message="DeviceStringType must be empty or 'wwn'"
+	// DeviceStringType instructs caph to either use the short device name, or the WWN when calling
+	// ImageURLCommand. It is not used when ImageURLCommand is not set. Allowed values are "short"
+	// or "wwn". Example: When "short" is used, ImageURLCommand receives "sda"; when "wwn" is
+	// used, it receives "eui.00253885910c8cec" or "0x500a07511bb48b25".
+	// +kubebuilder:validation:XValidation:rule="self == '' || self == 'short' || self == 'wwn'",message="DeviceStringType must be empty, 'short', or 'wwn'"
 	// +optional
 	DeviceStringType DeviceStringType `json:"deviceStringType,omitempty"`
 
--- a/api/v1beta1/hetznerbaremetalmachine_types.go
+++ b/api/v1beta1/hetznerbaremetalmachine_types.go
@@ -47,15 +47,11 @@ var errUnknownSuffix = errors.New("unknown suffix")
 type ImageType string
 
 // DeviceStringType controls what CAPH passes as the device argument to ImageURLCommand.
-// Allowed values are "" (same as "short"), "short", and "wwn".
+// The zero value ("") passes the short device name (e.g. "sda"). Set to "wwn" to pass the WWN instead.
 type DeviceStringType string
 
-const (
-	// DeviceStringTypeShort passes the short device name (e.g. "sda") to ImageURLCommand.
-	DeviceStringTypeShort DeviceStringType = "short"
-	// DeviceStringTypeWWN passes the WWN (e.g. "eui.00253885910c8cec") to ImageURLCommand.
-	DeviceStringTypeWWN DeviceStringType = "wwn"
-)
+// DeviceStringTypeWWN passes the WWN (e.g. "eui.00253885910c8cec") to ImageURLCommand.
+const DeviceStringTypeWWN DeviceStringType = "wwn"
 
 const (
 	// ImageTypeTar defines the image type for tar files.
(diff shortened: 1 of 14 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -1418,15 +1418,23 @@ func (s *Service) actionImageInstallingImageURLCommand(ctx context.Context, sshC
...
-		}
-
 		// get device names from storage device
-		deviceNames := getDeviceNames(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.ListOfWWN(), storage)
+		var deviceNames []string
+		switch s.scope.HetznerBareMetalMachine.Spec.InstallImage.DeviceStringType {
+		case infrav1.DeviceStringTypeWWN:
+			deviceNames = s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.ListOfWWN()
+			if len(deviceNames) == 0 {
+				return actionError{err: fmt.Errorf("DeviceStringType is %q but no WWN is configured in rootDeviceHints", infrav1.DeviceStringTypeWWN)}
```
janiskemper, line 1428: is this something you expect ever to happen? I guess not right? Because then an error is fine but we should add a comment that this is not expected. 

If we expect it to happen, we need a condition etc. and proper lifecycle handling.

pkg/services/baremetal/host/host.go
```diff
@@ -1418,15 +1418,23 @@ func (s *Service) actionImageInstallingImageURLCommand(ctx context.Context, sshC
...
-		// Device names can change during restart.
-		storage, err := obtainHardwareDetailsStorage(ctx, sshClient)
-		if err != nil {
-			return actionError{err: fmt.Errorf("failed to obtain hardware details storage: %w", err)}
-		}
-
 		// get device names from storage device
-		deviceNames := getDeviceNames(s.scope.HetznerBareMetalHost.Spec.RootDeviceHints.ListOfWWN(), storage)
+		var deviceNames []string
+		switch s.scope.HetznerBareMetalMachine.Spec.InstallImage.DeviceStringType {
```
janiskemper, line 1422: can you add examples again to the code here, similar to what you have in the API? This would make it easier to understand

pkg/services/baremetal/host/host.go
guettli: no, this is already validated. It is not expected.

pkg/services/baremetal/host/host.go
guettli: @janiskemper I added examples ind462979d1 is that what you wanted?

janiskemper (commented):
LGTM. Plz check @Dhairya-Arora01

api/v1beta1/hetznerbaremetalmachine_types.go
```diff
@@ -180,6 +191,14 @@ type InstallImage struct {
 	// +optional
 	ImageURLCommand string `json:"imageURLCommand,omitempty"`
 
+	// DeviceStringType instructs caph to either use the short device name, or the WWN when calling
```
Dhairya-Arora01: CAPH should be uppercase.

api/v1beta1/hetznerbaremetalmachine_types.go
```diff
@@ -180,6 +191,14 @@ type InstallImage struct {
...
+	// ImageURLCommand. It is not used when ImageURLCommand is not set. Allowed values are "",
+	// "short" (same as ""), and "wwn". Example: When "" or "short" is used, ImageURLCommand
+	// receives "sda"; when "wwn" is used, it receives "eui.00253885910c8cec" or "0x500a07511bb48b25".
+	// +kubebuilder:validation:XValidation:rule="self == '' || self == 'short' || self == 'wwn'",message="DeviceStringType must be empty, 'short', or 'wwn'"
```
Dhairya-Arora01: `// +kubebuilder:validation:Enum=short;wwn` wouldn't this be simpler and equivalent

guettli: @Dhairya-Arora01 enums are discouraged. See: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas

api/v1beta1/hetznerbaremetalmachine_types.go
guettli: @Dhairya-Arora01 done in 596031450

Dhairya-Arora01 (dismissed):

api/v1beta1/hetznerbaremetalmachine_types.go
janiskemper: We have 64 occurences of "kubebuilder:validation:Enum", including many from CAPI in the vendor directory. I don't think that we need to think about backward compatibility if we already have this in many cases.

@guettli do you think backward compatibility is still relevant if our existing code base contains this already heavily? If not, I would suggest to align it and to use it here as well.

guettli: PR uses enums now. See a8a0654.

janiskemper (approved):

guettli:
Backport in: [✨ DeviceStringType field for ImageURLCommand (sda vs WWN) (backport #2130) by guettli · Pull Request #2131 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/2131)

PR #2132, opened 2026-06-30
:seedling: backport: Enable proxy protocol for control plane load balancer (#2113)

guettli:
## Summary

Back-port of #2113 to `release-1.1`.

Since `release-1.1` has no `v1beta2` API and webhooks live in `api/v1beta1/` (not `pkg/webhook/`), the changes were adapted manually:

- `ProxyProtocolForControlPlaneLoadBalancerAnnotation` constant added to `api/v1beta1/hetznercluster_types.go`
- `EnableProxyProtocol bool` field added to `LoadBalancerSpec` in `api/v1beta1/types.go`
- `AllControlPlaneNodesReadyForProxyProtocol` added to `ClusterScope` using `infrav1` (no `infrav2` in this branch)
- `reconcileServices` updated with proxy protocol enable/migration logic (returns `reconcile.Result` now)
- Webhook validation added to `api/v1beta1/hetznercluster_webhook.go` to prevent disabling proxy protocol once enabled
- CRD manifests updated for both `hetznerclusters` and `hetznerclustertemplates`
- Webhook test added to `api/v1beta1/hetznercluster_webhook_test.go`

### DIFF
--- a/api/v1beta1/hetznercluster_types.go
+++ b/api/v1beta1/hetznercluster_types.go
@@ -44,6 +44,11 @@ const (
 	// format to use for baremetal nodes. If "true" "hrobot://" will be used. If not set or empty,
 	// then the old format ("hcloud://bm-") gets used.
 	UseHrobotProviderIDForBaremetalAnnotation = "capi.syself.com/use-hrobot-provider-id-for-baremetal"
+
+	// ProxyProtocolForControlPlaneLoadBalancerAnnotation is set on control-plane nodes by an external
+	// service (e.g. a node-configuration daemonset) to signal that the node is ready to receive
+	// PROXY-protocol headers. CAPH reads this annotation but never writes it.
+	ProxyProtocolForControlPlaneLoadBalancerAnnotation = "capi.syself.com/proxy-protocol-for-controlplane-loadbalancer"
 )
 
 // HetznerClusterSpec defines the desired state of HetznerCluster.
--- a/api/v1beta1/hetznercluster_types.go
+++ b/api/v1beta1/hetznercluster_types.go
@@ -45,9 +45,15 @@ const (
 	// then the old format ("hcloud://bm-") gets used.
 	UseHrobotProviderIDForBaremetalAnnotation = "capi.syself.com/use-hrobot-provider-id-for-baremetal"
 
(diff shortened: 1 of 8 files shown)


api/v1beta1/hetznercluster_types.go
```diff
@@ -44,6 +44,11 @@ const (
 	// format to use for baremetal nodes. If "true" "hrobot://" will be used. If not set or empty,
 	// then the old format ("hcloud://bm-") gets used.
 	UseHrobotProviderIDForBaremetalAnnotation = "capi.syself.com/use-hrobot-provider-id-for-baremetal"
+
+	// ProxyProtocolForControlPlaneLoadBalancerAnnotation is set on control-plane nodes by an external
```
janiskemper: why do you not have the same comment as in the original PR here?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -285,27 +289,29 @@ func (s *Service) reconcileLBProperties(ctx context.Context, lb *hcloud.LoadBala
 	return multierr
 }
 
-func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer) error {
+func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer) (reconcile.Result, error) {
```
janiskemper, line 292: there is a diff between these changes and the original changes. Why?

api/v1beta1/hetznercluster_types.go
guettli: Fixed — expanded to the full comment from the original PR (explaining migration vs. new-cluster distinction).

pkg/services/hcloud/loadbalancer/loadbalancer.go
guettli: Fixed — reordered locals to match the original PR structure. The only intentional difference that remains is the nil check on `ControlPlaneEndpoint`: in v1beta1 it is a pointer (`*clusterv1beta1.APIEndpoint`), so it can be nil on new clusters, whereas in v1beta2 it is a value type.

janiskemper (approved):

PR #2133, opened 2026-07-01
:seedling: Always accept Loopback IP in CSR.

guettli:
The loopback address 127.0.0.1 is always valid in a kubelet CSR, regardless of the machine's listed addresses.


guettli:
no longer needed.

PR #2140, opened 2026-07-02
:seedling:  show bare metal machine IP correctly as ExternalIP

### LINKED ISSUE #2137: Bare metal machine IP is shown as INTERNAL-IP with a CIDR suffix instead of EXTERNAL-IP
The IP address type of bare-metal machines is broken since "ages". PR #2140 fixes it. No annotation or field is needed to enable the fix: it applies automatically to every new `HetznerBareMetalMachine`. Existing machines keep reporting addresses the old (incorrect) way — see "Fix" below for why.

For `HetznerBareMetalMachine`s, the NIC IP address discovered on the host is always reported as
`InternalIP` (`Machine.status.addresses[].type), and the raw value from `ip addr show` (which
includes the CIDR prefix length, e.g. `/26`) is stored verbatim. This makes bare metal machines show
up differently than hcloud machines in `kubectl get machines`:

## Current state

```text
NAME                            NODE NAME             PROVIDER ID           INTERNAL-IP          EXTERNAL-IP
foo-jxzdg-75rxh                 foo-jxzdg-75rxh       hcloud://100000001                         198.51.100.11
foo-md-2-4xwl7-wrq9n-6rvzq      bm-foo-9000001        hcloud://bm-9000001   198.51.100.20/26  <--- wrong
```

Note the bare metal machine `foo-md-2-4xwl7-wrq9n-6rvzq`: its IP (`198.51.100.20/26`) is in the
`INTERNAL-IP` column and still carries the `/26` CIDR suffix.

## Target state

```text
NAME                            NODE NAME             PROVIDER ID           INTERNAL-IP    EXTERNAL-IP
foo-jxzdg-75rxh                 foo-jxzdg-75rxh      hcloud://100000001                   198.51.100.11
foo-md-2-4xwl7-wrq9n-6rvzq      bm-foo-9000001       hcloud://bm-9000001                  198.51.100.20
```

The bare metal machine's IP should show up in the `EXTERNAL-IP` column, without the CIDR suffix,
just like the hcloud machines.

## Root cause

1. The NIC IP is scraped from the host over SSH with plain `ip addr show`, which prints addresses in
   CIDR notation. The prefix length is never stripped:
   - [pkg/services/baremetal/client/ssh/nic-info.sh](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/pkg/services/baremetal/client/ssh/nic-info.sh)
     (`ip -4 addr show dev "$iname" | awk '/inet /{print $2}'` and the IPv6 equivalent)
   - [pkg/services/baremetal/client/ssh/ssh_client.go:261-269](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/pkg/services/baremetal/client/ssh/ssh_client.go#L261-L269)
     (`GetHardwareDetailsNics`, runs the embedded script)
   - [pkg/services/baremetal/host/host.go:930-984](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/pkg/services/baremetal/host/host.go#L930-L984)
     (`obtainHardwareDetailsNics`, copies `nic.IP` verbatim into `infrav1.NIC.IP`)

2. Every NIC address is unconditionally tagged as `MachineInternalIP`, regardless of whether it is
   actually a public (external) or private (internal) address:
   - [pkg/services/baremetal/baremetal/baremetal.go:942-960](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/pkg/services/baremetal/baremetal/baremetal.go#L942-L960)
     (`nodeAddresses`)

   For comparison, the hcloud path correctly distinguishes public vs. private server IPs and tags
   them `MachineExternalIP` / `MachineInternalIP` respectively:
   - [pkg/services/hcloud/server/server.go:2107-2145](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/pkg/services/hcloud/server/server.go#L2107-L2145)
     (`statusAddresses`)

The `NIC.IP` API doc comment already promises a bare address ("This will be an IPv4 or IPv6
address"), so the CIDR suffix is a bug relative to the field's own contract, not just a display
issue:
[api/v1beta2/hetznerbaremetalhost_types.go:488-494](https://github.com/syself/cluster-api-provider-hetzner/blob/5fb187f5dfd3bc16ce9b7bb549a7fdee4917ad80/api/v1beta2/hetznerbaremetalhost_types.go#L488-L494).

## Fix

Implemented in #2140. No CRD/API change is needed — there is no annotation and no spec field:

- Strip the CIDR suffix (`/xx`) when parsing the `ip=` field in `obtainHardwareDetailsNics` (or in
  `nic-info.sh` itself), so `infrav1.NIC.IP` only ever contains a bare IP address.
- In `nodeAddresses`, classify each NIC address as `MachineExternalIP` when it is a public/routable
  address and `MachineInternalIP` when it is a private one (e.g. RFC1918/ULA or the machine's
  private network range), mirroring the distinction already made in
  `pkg/services/hcloud/server/server.go`.

**Existing machines must not change.** Flipping an already-running machine's reported address
type from `InternalIP` to `ExternalIP` (or vice versa) could surprise anything in the management
cluster that reads `status.addresses` — see "Caution" below. So the corrected classification is
applied only the first time a `HetznerBareMetalMachine` gets an IP-type address
(`InternalIP`/`ExternalIP`) in `status.addresses`, i.e. for a genuinely new machine. Once a
machine has reported such an address, its type is effectively frozen: on later reconciles it
keeps being computed the old (incorrect) way — CIDR suffix kept, always `InternalIP` — for as
long as that machine exists. There is no annotation or field to opt in or out of this; it is
purely based on whether the machine already reported an address before this fix was deployed.

In other words: after upgrading, new machines get the correct `ExternalIP`/`InternalIP`
classification automatically; machines that already existed keep the old, wrong classification
forever. This means a cluster can have a mix of both right after the upgrade, until old machines
are replaced (e.g. via a `MachineDeployment` rollout).

## Caution, but it only affects the management cluster

`Machine.status.addresses` (and `HetznerBareMetalMachine.status.addresses`) lives entirely in the
**management cluster**. It has no effect on the **workload cluster**: workload-cluster `Node`
objects are populated independently (e.g. by `hcloud-cloud-controller-manager` running inside the
workload cluster), so anything reading them there — CNI/CiliumNetworkPolicy, firewalls, etc. — never
sees this field and is not affected by this change.

The only real consumer of `status.addresses` inside this repo is `GuestCSRReconciler`
(`pkg/csr/csr.go: ValidateKubeletCSR`), which uses it to check that a kubelet-serving CSR's IP
SANs are among the machine's known addresses before approving it. That check treats `InternalIP`
and `ExternalIP` identically and already strips any CIDR suffix itself
(`strings.Split(address.Address, "/")[0]`), so it is unaffected by either part of this fix.

If you have your own automation in the management cluster that reads a bare-metal machine's
`InternalIP` from `status.addresses` and expects a specific value or type, it will see a
different value/type once that machine gets the corrected classification (i.e. for new machines
going forward).

guettli:
## Summary
- Bare metal machine NIC IPs were always reported as `InternalIP` in `Machine.status.addresses`, with the raw `/xx` CIDR suffix from `ip addr show` kept verbatim — unlike hcloud machines, which correctly report `ExternalIP`/`InternalIP` without a CIDR suffix.
- Fixes this: strips the CIDR suffix and classifies each NIC address as `ExternalIP` (public/routable) or `InternalIP` (private, RFC1918/ULA), mirroring the existing distinction in `pkg/services/hcloud/server/server.go`.
- No CRD/API change: no annotation, no spec field. The corrected behavior is applied automatically — there is nothing to opt in to.
- **Existing machines are unaffected.** The corrected classification is only applied the first time a `HetznerBareMetalMachine` reports an IP-type address (`InternalIP`/`ExternalIP`) in `status.addresses` — i.e. for a genuinely new machine. Once a machine has reported such an address, it keeps being computed the old (incorrect) way — CIDR suffix kept, always `InternalIP` — for the rest of its lifetime.
- This only changes `Machine.status.addresses` in the **management cluster**. It has no effect on the workload cluster: Node objects there are populated independently (e.g. by hcloud-cloud-controller-manager), so tools reading them — CNI/Cilium, firewalls, etc. — never see this field and are not affected. The one real consumer inside this repo, `GuestCSRReconciler` (kubelet CSR validation in `pkg/csr/csr.go`), treats `InternalIP`/`ExternalIP` identically and already strips any CIDR suffix itself, so it is unaffected too.
- If you have your own automation in the management cluster that reads a `HetznerBareMetalMachine`'s `InternalIP` from `status.addresses` and expects a specific value/type, it will see a different value/type for new machines going forward.
- After upgrading, a cluster can therefore have a mix of old-style and new-style machines until the old ones are replaced (e.g. via a `MachineDeployment` rollout).

related to #2137 (but we don't want to close the issue until we merge this to main)


janiskemper (commented):
LGTM from high-level

Dhairya-Arora01 (approved):

PR #2141, opened 2026-07-02
:seedling: pre-commit config, for fast local-first CI.

guettli:
pre-commit config, for fast local-first CI.

See: https://pre-commit.com/

It runs a small, fast CI after you started `git commit`. It helps to avoid common failures, like not running `make generate`.


janiskemper:
@batistein should review it because he was having a similar topic in another repo

guettli:
Currently, pre-commit does not work well with ai-agent. Maybe we add this later.

PR #2142, opened 2026-07-03
:seedling: reword to custom provisioner.

guettli:
Rewording to "custom provisioner".


janiskemper:
I don't think we can change the conditions etc like this.. I would not do this right now. We shouldn't do anything breaking and can do it for v1beta2

PR #2148, opened 2026-07-06
✨ Add onExhaustion Retire option to bare metal remediation

### LINKED ISSUE #2144: Add a `retire` ending to bare metal remediation so a failed server leaves the cluster instead of being reused
## Summary

Add an `onExhaustion` option to the bare metal remediation strategy so that, when reboots do not fix a server, the user can retire it (set a permanent error, which removes the machine and keeps the host out of the pool) instead of today's behavior of deleting the machine and letting the same server be picked again.

## The Problem

When bare metal remediation runs out of reboots and the node is still unhealthy, CAPH deletes the machine and frees the host back into the selectable pool, so it gets provisioned again. This is the right move for a temporary problem. For a real hardware failure a reboot never helps, so the same broken host is picked again, fails again, and MachineHealthCheck keeps remediating it in a loop. The broken server never leaves the cluster, and every workload scheduled onto it is disrupted again.

## The Solution

CAPH already has the permanent error feature. `HetznerBareMetalHost.SetError(PermanentError, msg)` sets the error and message, adds the `capi.syself.com/permanent-error` annotation with the reason, deletes the machine, and keeps the host out of selection (the permanent error is not cleared on deprovision, and `skipHost` skips a host that has an error). A human clears it by removing the annotation.

We add a choice for what remediation does when the reboots run out:

- **New strategy field** `onExhaustion` with values `Reuse` (default, today's behavior) and `Retire`.
- **Retire** sets a permanent error on the host with the reason. The existing permanent error feature then removes the machine and keeps the host out of the pool until a human removes the annotation.

This is additive. Behavior only changes when the user sets `onExhaustion: Retire`.

### 1. Add the `onExhaustion` field to the strategy

In `api/v1beta2/remediation_types.go` and its v1beta1 mirror, add to `RemediationStrategy` a typed `OnExhaustion` field with values `Reuse` and `Retire`, `+kubebuilder:default=Reuse`, enum-validated.
### 2. Retire the host when reboots run out

In the exhausted path of `handlePhaseWaiting` (`pkg/services/baremetal/remediation/remediation.go`), where the node is still unhealthy after the timeout, if `OnExhaustion == Retire`, call `SetError(infrav1.PermanentError, reason)` on the host and patch it, with `reason` like "retired by remediation after N failed reboots". The permanent error deletes the machine through the existing `HasFatalError` path, so no separate machine-deletion signal is needed. `Reuse` (or empty) keeps today's behavior.

### 3. Reason for the permanent error

When the MachineHealthCheck marks a node unhealthy, it records the node condition that tripped on the Machine's `HealthCheckSucceeded` condition, in a message that names the condition, its status, and its reason. CAPI sets this message [here](https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.1/internal/controllers/machinehealthcheck/machinehealthcheck_targets.go#L146-L151).

In the retire path, read that message from the owner Machine, which the controller already fetches, and pass it as the `reason` to `SetError`. If the Machine is missing or the message is empty, fall back to the `retired by remediation after N failed reboots` message from step 2. This way the permanent error says why the host was retired, not only that the reboots ran out.

## Expected Behavior

| Scenario | Behavior |
|----------|----------|
| `onExhaustion: Reuse` (or unset), reboots fail | Machine deleted, host freed, host can be picked again (today's behavior) |
| `onExhaustion: Retire`, reboots fail | Permanent error set on host with the reason from the Machine's `HealthCheckSucceeded` condition, machine deprovisioned, host kept out of the pool until the `capi.syself.com/permanent-error` annotation is removed |

## Related

With the `retryLimit: 0` fix (separate issue https://github.com/syself/cluster-api-provider-hetzner/issues/2143), a `retryLimit: 0` + `onExhaustion: Retire` template goes straight to retire with no reboot, which is the hardware-failure case. This issue does not depend on that fix, but the two together cover that use case.

abdullah599:
Added onExhaustion Retire option to bare metal remediation

Fixes #2144

None

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta1/hetznerbaremetalremediation_types.go
```diff
@@ -33,7 +33,44 @@ const (
...
-	Strategy *RemediationStrategy `json:"strategy,omitempty"`
+	Strategy *BareMetalRemediationStrategy `json:"strategy,omitempty"`
+}
+
+// OnExhaustionAction defines what remediation does when the retries run out and
+// the node is still unhealthy.
+type OnExhaustionAction string
+
+const (
+	// OnExhaustionReuse deletes the machine and frees the host back into the
```
janiskemper, line 44: ReUse IMO

api/v1beta1/hetznerbaremetalremediation_types.go
```diff
@@ -33,7 +33,44 @@ const (
...
+type BareMetalRemediationStrategy struct {
+	// The shared remediation fields (type, retryLimit, timeout, cooldown).
+	RemediationStrategy `json:",inline"`
+
+	// OnExhaustion selects what happens when remediation runs out of retries and
+	// the node is still unhealthy. Reuse (default) deletes the machine and frees
+	// the host to be provisioned again. Retire sets a permanent error on the host,
+	// which deletes the machine and keeps the host out of the pool until a human
+	// removes the capi.syself.com/permanent-error annotation.
+	// +kubebuilder:default=Reuse
```
janiskemper: AFAIK we shouldn't set a default here because this would interfere with GitOps controllers not setting a default. The better strategy will be to leave it empty as default and say that this behavior leads to "re-use"

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,46 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 			"reboot remediation succeeded: Node is healthy again")
 	}
 
+	// Reboots are exhausted and the node is still unhealthy. When the strategy
+	// asks to retire, set a permanent error on the host instead of freeing it:
+	// this deletes the machine through the existing HasFatalError path and keeps
```
janiskemper: can you add that this has the same effect on this remediation as if we set the condition below on failed? I didn't understand that you mean this until I read the code and wondered why we don't set the condition anymore in this path.

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,46 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
...
+func (s *Service) retireHost(ctx context.Context, host *infrav1.HetznerBareMetalHost) error {
+	patchHelper, err := v1beta1patch.NewHelper(host, s.scope.Client)
+	if err != nil {
+		return fmt.Errorf("failed to init patch helper: %s %s/%s %w", host.Kind, host.Namespace, host.Name, err)
+	}
+
+	// RetryCount is the number of reboots that were attempted. It is 0 when
+	// retryLimit is 0 (retire without rebooting), so phrase the reason accordingly.
+	retryCount := s.scope.BareMetalRemediation.Status.RetryCount
+	reason := "retired by remediation: retryLimit is 0, node retired without a reboot attempt"
```
janiskemper: this is not a great reason. In theory, the reason exists somewhere because this remediation path was triggered - knowing that retiring the host is desired. 

However, how would we be able to find that out?

pkg/services/baremetal/remediation/remediation_test.go
```diff
@@ -110,3 +117,120 @@ var _ = Describe("Test AddRebootAnnotation", func() {
 		}),
 	)
 })
+
+var _ = Describe("Test handlePhaseWaiting onExhaustion", func() {
+	// Deterministic unit test with a fake client (no host controller running, so
+	// nothing overwrites the permanent error). It exercises the exhausted path:
+	// reboots are used up and the node is still unhealthy, so the strategy decides
+	// whether the host is reused (today's behavior) or retired.
```
janiskemper: it's not "today's behavior"

pkg/services/baremetal/remediation/remediation_test.go
```diff
@@ -110,3 +117,120 @@ var _ = Describe("Test AddRebootAnnotation", func() {
...
+				WithStatusSubresource(machine).
+				Build()
+
+			service := &Service{scope: &scope.BareMetalRemediationScope{
+				Client:               c,
+				Machine:              machine,
+				BareMetalRemediation: remediation,
+			}}
+
+			// The machine has no MachineNodeHealthy condition, so the node counts as
```
janiskemper: I don't understand this comment.

pkg/services/baremetal/remediation/remediation_test.go
```diff
@@ -110,3 +117,120 @@ var _ = Describe("Test AddRebootAnnotation", func() {
...
+			Expect(res.RequeueAfter).To(BeZero())
+
+			// Either way, remediation stops.
+			Expect(remediation.Status.Phase).To(Equal(infrav1.PhaseDeleting))
+
+			updatedHost := &infrav1.HetznerBareMetalHost{}
+			Expect(c.Get(ctx, client.ObjectKeyFromObject(host), updatedHost)).To(Succeed())
+			if tc.expectHostPermanentError {
+				Expect(updatedHost.Spec.Status.ErrorType).To(Equal(infrav1.PermanentError))
+				// Pin the retire branch's real output: the reason differs for 0 reboots
```
janiskemper: what does this mean?

api/v1beta1/hetznerbaremetalremediation_types.go
abdullah599: I would disagree here. It is a single word: https://www.merriam-webster.com/dictionary/reuse

pkg/services/baremetal/remediation/remediation.go
abdullah599: We decided to ignore this comment.

ref: https://chat.syself.org/#narrow/stream/20-provider.2Fhetzner/topic/Permanent.20error.20when.20reboot.20timed.20out.3F/near/593204

pkg/services/baremetal/remediation/remediation_test.go
janiskemper: I'm still not sure what "reason" is.

api/v1beta1/hetznerbaremetalremediation_types.go
janiskemper: Interesting. I don't like how that looks. I prefer re-use 😆

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,45 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
 			"reboot remediation succeeded: Node is healthy again")
 	}
 
+	// Reboots are exhausted and the node is still unhealthy. Retire also gets the
+	// machine deleted (like the reuse path below), but via a permanent error on the
```
janiskemper: you write here "reuse path" and I think you mean just everything after this if statement, because that's the default. However, that's not understandable from this

api/v1beta1/hetznerbaremetalremediation_types.go
```diff
@@ -33,7 +33,43 @@ const (
...
+type OnExhaustionAction string
+
+const (
+	// OnExhaustionReuse deletes the machine and frees the host back into the
+	// selectable pool, so it can be provisioned again. This is the behavior when
+	// onExhaustion is unset, and the right choice for temporary problems.
+	OnExhaustionReuse OnExhaustionAction = "Reuse"
+
+	// OnExhaustionRetire sets a permanent error on the host, which deletes the
+	// machine and keeps the host out of the pool until a human removes the
```
janiskemper: rewrite this. The human should from now on (see other PR) put the server out of maintenance mode

docs/caph/03-reference/07-hetzner-bare-metal-remediation-template.md
```diff
@@ -9,9 +9,10 @@ In `HetznerBareMetalRemediationTemplate` you can define all important properties
...
+| `template.spec.strategy.retryLimit`   | `int`    | `0`      | no       | Set maximum of remediation retries. Zero retries if not set.                                                                                                                                                                                                                       |
+| `template.spec.strategy.timeout`      | `string` |          | yes      | Timeout of one remediation try. Should be of the form "10m", or "40s"                                                                                                                                                                                                              |
+| `template.spec.strategy.onExhaustion` | `string` |          | no       | What to do when retries run out and the node is still unhealthy. `Reuse` deletes the machine and frees the host to be provisioned again. Note: when unset, behaves like `Reuse`. `Retire` sets a permanent error on the host, which deletes the machine and keeps the host out of the pool until a human removes the `capi.syself.com/permanent-error` annotation. |
```
janiskemper: this is also outdated

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,55 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
...
+		return reconcile.Result{}, s.retireHost(ctx, host, capiMachine)
+	}
+
 	return reconcile.Result{}, s.setOwnerRemediatedConditionToFailed(ctx, "because retryLimit is reached and reboot timed out")
 }
 
+// retireHost sets a permanent error on the host so it leaves the cluster instead of
+// being reused. The permanent error deletes the machine through the HasFatalError path,
+// and skipHost keeps the host out of selection (the error is not cleared on deprovision)
+// until a human removes the permanent-error annotation.
```
janiskemper: this is outdated

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,55 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
...
+	patchHelper, err := v1beta1patch.NewHelper(host, s.scope.Client)
+	if err != nil {
+		return fmt.Errorf("failed to init patch helper: %s %s/%s %w", host.Kind, host.Namespace, host.Name, err)
+	}
+
+	// The MachineHealthCheck writes the node condition that made the machine unhealthy onto
+	// the Machine's HealthCheckSucceeded condition. Use it as the reason so the permanent
+	// error explains why the host was retired. Fall back to how remediation ended when the
+	// Machine or the message is missing.
+	reason := ""
```
janiskemper: I think `var reason string` is a bit nicer

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -207,9 +207,55 @@ func (s *Service) handlePhaseWaiting(ctx context.Context) (res reconcile.Result,
...
+		reason = conditions.GetMessage(capiMachine, clusterv1.MachineHealthCheckSucceededCondition)
+	}
+	if reason == "" {
+		// RetryCount is the number of reboots attempted; it is 0 when retryLimit is 0.
+		reason = "retryLimit is 0, node retired without a reboot attempt"
+		if retryCount := s.scope.BareMetalRemediation.Status.RetryCount; retryCount > 0 {
+			reason = fmt.Sprintf("node still unhealthy after %d failed reboot(s)", retryCount)
+		}
+	}
+	message := fmt.Sprintf("retired by remediation: %s", reason)
```
janiskemper: I don't think we should do that. It is not important which controller did this action IMO and this string will be then the only thing shown to users reading the condition via printcolumns

janiskemper (commented):
@Dhairya-Arora01 plz review

Dhairya-Arora01 (dismissed):

pkg/services/baremetal/remediation/remediation.go
janiskemper: we implemented it actually via a Machine HealthCheck condition - just for completion

janiskemper:
we decided to change the maintenance-mode feature again. Therefore we have to update this PR also again, right?

Dhairya-Arora01 (approved):

PR #2156, opened 2026-07-09
✨ Boot HCloud imageURL machines directly into rescue system

### LINKED ISSUE #2153: Boot Hetzner Cloud machines directly into rescue by creating the server powered off
## Summary

Create the rescue server with `start_after_create=false`, enable the rescue system while the server is still powered off, and then power it on so its first boot goes straight into rescue. This removes one full operating system boot and cuts the time to reach the rescue system by about 60 to 90 seconds per machine.

## The Problem

When CAPH provisions an HCloud machine from an image URL, it walks a boot state machine. The states are defined in `api/v1beta1/types.go` and dispatched in `Reconcile` in `pkg/services/hcloud/server/server.go`:

```
Unset -> Initializing -> EnablingRescue -> BootingToRescue -> RunningImageCommand -> BootingToRealOS -> OperatingSystemRunning
```

To reach the rescue system today, three things happen.

First, in `handleBootStateUnset`, CAPH creates the server through `createServerFromImageURL`. That call uses the placeholder image `ubuntu-24.04` (the `preRescueOSImage` constant). Inside `createServer` the option `StartAfterCreate` is hardcoded to `true`, so Hetzner creates the server and powers it on right away. It boots a full Ubuntu. The state then becomes `Initializing`.

Second, in `handleBootStateInitializing`, CAPH polls the server every 10 seconds until its status is `Running`. In other words, it waits for the whole Ubuntu boot to finish. Only after Ubuntu is fully up does it call `EnableRescueSystem`, store the returned action ID, and move to `EnablingRescue`.

Third, in `handleBootStateEnablingRescue`, CAPH waits for the enable rescue action to finish and then reboots the server over SSH. This time the server boots into the rescue system, and the state becomes `BootingToRescue`.

The waste is the whole Ubuntu boot in the first two steps. We boot a complete operating system, wait around 90 to 120 seconds for it to be `Running` (plus up to 10 seconds of polling delay), and then reboot into a different system. The Ubuntu boot is thrown away. It exists only because the create call powers the server on before we get a chance to enable rescue.

## The Solution

The Hetzner Cloud API lets us create a server without powering it on, by setting `start_after_create=false`. So we can create the rescue server powered off, enable rescue while it is still off, and then power it on. Enabling rescue only sets a flag that says the next boot is the rescue system. Because the server was never started before, the power on is its first boot, and it goes straight into rescue. We never boot Ubuntu.

* The rescue path (image URL) creates the server powered off and never boots the placeholder Ubuntu.
* The `Initializing` state keeps its meaning: wait until the server is ready, then enable rescue. It now waits for the create action to finish instead of waiting for Ubuntu to reach status `Running`.
* We power the stopped server on instead of rebooting a running one over SSH.
* The snapshot path (image name) is unchanged. It keeps booting the real image directly. There is no rescue step to wait for on that path, so the create action does not help there. Making that path also wait on actions instead of polling the server status can be a separate follow up.

Here is why this works at the API level. After create with `start_after_create=false` the server reports status `off`. The `enable_rescue` action succeeds while the server is off and sets `rescue_enabled=true`. The first power on then boots into rescue. The last section, What the PoC confirmed, shows how we tested this.

```
Now (slow):
  create (start=true) -> Ubuntu boots -> wait Running -> enable rescue -> reboot -> rescue
  \_________________________ about 90 to 120 s to reach rescue __________________________/

New (fast):
  create (start=false) -> enable rescue -> power on -> rescue
  \___________ about 30 to 50 s to reach rescue ____________/
```

The new rescue flow as a boot state machine. The states stay the same. `Initializing` and `EnablingRescue` each wait on the action the previous state started:

```mermaid
flowchart LR
    U["Unset<br/>create server powered off,<br/>store create action"]
    I["Initializing<br/>wait create action,<br/>enable rescue"]
    E["EnablingRescue<br/>wait enable rescue action,<br/>power on"]
    B["BootingToRescue"]
    R["RunningImageCommand"]
    RO["BootingToRealOS"]
    OS["OperatingSystemRunning"]
    U --> I --> E --> B --> R --> RO --> OS
```

## Implementation Details

All changes are in `pkg/services/hcloud/server/server.go` unless a different file is named. Do not touch the snapshot path (`createServerFromImageName` and `handleBootingToRealOS`).

### 1. Return the create action from the client

`CreateServer` in `pkg/services/hcloud/client/client.go` returns only `*hcloud.Server` and throws away the create action. We need that action later, in the `Initializing` state (step 4), to wait for the server to finish provisioning before enabling rescue. Change `CreateServer` to return the full `hcloud.ServerCreateResult`, which holds the server and the create action.

### 2. Make StartAfterCreate a parameter

`createServer` hardcodes `startAfterCreate := true`. Make it a function argument so each path decides.

* `createServerFromImageName` (snapshot) passes `true`, so it boots the real image as it does today.
* `createServerFromImageURL` (rescue) passes `false`, so it does not boot the placeholder Ubuntu.

Keep passing `ubuntu-24.04` on the rescue path. The create API still requires an image field, but it will never boot.

### 3. Rescue path: store the create action

In `createServerFromImageURL`, store the create action ID in a new `ActionIDCreate` field on `HCloudMachineStatusExternalIDs` (`api/v1beta1/hcloudmachine_types.go`), next to the existing `ActionIDEnableRescueSystem`. The state stays `Initializing`, as today.

### 4. Initializing: wait for the create action instead of the server status

`handleBootStateInitializing` currently polls the server every 10 seconds until its status is `Running`, because it had to wait for the Ubuntu boot. There is no boot anymore, so replace that status check with a wait on the create action. Keep the existing 6 minute timeout guard at the top of the handler, and keep the rest of the handler (the `EnableRescueSystem` call, storing the action ID, and the move to `EnablingRescue`) as it is. The exact retry intervals are in Retries and backoff below.

* Poll `ActionIDCreate` with `GetAction`, the same way `EnablingRescue` already waits for the enable rescue action. While the create action is still running, requeue and wait, because the server is still being provisioned.
* If the create action finished with an error, set an error and remediate, the same way a failed enable rescue action is handled today.
* Once the create action is finished successfully, the server is provisioned and unlocked, so call `EnableRescueSystem` with the same options used today (`ServerRescueTypeLinux64` and our SSH keys), store the returned action ID, and set the state to `EnablingRescue`, all as the handler already does. If the call still returns a locked error, requeue and retry, bounded by the timeout.

The create action needs no done marker like the `actionDone` sentinel of the enable rescue action, because the handler leaves the state in the same reconcile in which the action is finished.

### 5. EnablingRescue: power on instead of the SSH reboot

`handleBootStateEnablingRescue` keeps its structure: the 6 minute timeout guard, the wait for the enable rescue action with `GetAction`, and the check that `server.RescueEnabled` is true. Only the last step changes.

* Remove the reboot over SSH block (the section with the comment "Reboot via ssh, avoid API calls to hcloud (rate-limit)"). The server was never running, so the power on is the first boot and goes straight into rescue.
* Remove the 4 second requeue between the enable rescue action finishing and the reboot, with the comment that the reboot might get ignored otherwise. That delay only existed to protect the reboot. We power on instead, and testing against the API showed it accepts the next operation right after the action finishes, so the new flow adds no delay there.
* Call `PowerOnServer` as soon as the enable rescue action is finished and `RescueEnabled` is confirmed, then set the state to `BootingToRescue`. If the server is locked, requeue and retry, bounded by the timeout, using the same locked error check as `handleServerStatusOff`.

We do not need to wait on the power on action. `PowerOnServer` returns only an error today and discards its action, and we do not need to change that, because `BootingToRescue` already waits for the server to boot into rescue.

### 6. Retries and backoff

When a step fails, requeue and retry. The interval depends on the kind of failure.

* A `locked` error: Hetzner returns the error code `locked` when the server is busy right after create, so `enable_rescue` or `poweron` is rejected. Detect it with `hcloud.IsError(err, hcloud.ErrorCodeLocked)`, the same check `handleServerStatusOff` already uses today. When it is locked, requeue after 5 seconds and keep retrying. The lock is a short window that clears on its own, so a fixed short requeue is enough and we do not back off here.
* Any other error, a rate limit or something unexpected: return the error instead of a requeue. controller-runtime then applies its exponential backoff, the standard Kubernetes workqueue limiter. The existing `handleRateLimit` helper already does this and marks the rate limit condition when the error is a rate limit. Backoff fits here because these failures can last a while, unlike a short locked window.
* Overall limit: `Initializing` and `EnablingRescue` each keep their own 6 minute timeout. If the machine never gets through a state, the controller sets an error and CAPI remediates the machine, meaning it deletes and recreates it. So the worst case is retrying for up to 6 minutes in a state and then rebuilding.

The PoC never hit a `locked` error, even with ten servers created at once, so the locked retry is insurance and not the normal path.

## What the PoC confirmed

Before changing the controller, we tested the risky part with a small standalone program. It created servers with `start_after_create=false`, waited until each server finished provisioning, then called `enable_rescue` right away. We ran it one server at a time with tight timing, and then ten servers at once, like a scale up.

The result was clean in every server, including all ten at once. `enable_rescue` worked on the first try right after the server was ready, the API never returned `locked`, and no server needed a retry. The create action finishing and the status reaching `off` and unlocked happened at the same time, so either one is a valid signal for "the server is ready". We wait on the create action, to stay consistent with the enable rescue and other steps that already wait on actions. The first power on then cleared the `rescue_enabled` flag, which is what the controller waits for.

One thing is still open. The program only checked that the `rescue_enabled` flag clears, it did not log into the rescue system to confirm the hostname is `rescue`. A real boot takes longer than the flag change, so the exact end to end time is not confirmed yet. During implementation, confirm over SSH that the server actually reaches rescue after the first power on, and measure the real provisioning time against main.

abdullah599:
Boot HCloud imageURL machines directly into rescue system

Fixes #2153

Beyond the issue:

- After the enable rescue action finishes, the handler requeues once (`requeueImmediately`) instead of powering on in the same reconcile. This keeps the `EnablingRescueActionDone` condition observable and runs the `RescueEnabled` check on a fresh server object. Same pattern as `handleBootingToRealOS` ("Show changes in Status and go to next BootState").
- `GetAction` in the hcloud client now returns an error when the action does not exist. hcloud-go returns a nil action with no error on 404 ([`ResourceActionClient.GetByID`, v2.32.0](https://github.com/hetznercloud/hcloud-go/blob/v2.32.0/hcloud/action.go#L156-L170)), which would panic both action-wait handlers at `action.Finished`.
- `Initializing` got the same "action ID missing" guard as `EnablingRescue`, and both guard messages are reworded to say what is missing and that the machine will be remediated.
- The timing decisions were verified with a standalone PoC program that runs the exact controller sequence against the real Hetzner API, with the same hcloud-go version as CAPH (v2.32.0): create with `start_after_create=false`, wait for the create action, call `enable_rescue` immediately, wait for its action, power on immediately with no delay, then confirm the server starts and `rescue_enabled` clears (booted into rescue).

  Results over 5 servers: create action finished in 3.4 to 6.6 s, `enable_rescue` accepted on the first try 5/5, power on accepted directly after the enable action 5/5, server status changed from `off` to `running` right after the power on (the power on took effect) 5/5, rescue reached (`rescue_enabled` cleared) 5/5, `locked` never seen. 40 to 49 s from create to rescue, against 90 to 120 s today.

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/api/v1beta1/hcloudmachine_types.go
+++ b/api/v1beta1/hcloudmachine_types.go
@@ -208,9 +208,9 @@ type HCloudMachineStatusExternalIDs struct {
 	// +optional
 	ActionIDEnableRescueSystem int64 `json:"actionIdEnableRescueSystem,omitzero"`
 
-	// ActionIDCreate is the hcloud API Action result of CreateServer.
+	// ActionIDCreateServer is the hcloud API Action result of CreateServer.
 	// +optional
-	ActionIDCreate int64 `json:"actionIdCreate,omitzero"`
+	ActionIDCreateServer int64 `json:"actionIdCreateServer,omitzero"`
 }
 
 // HCloudMachine is the Schema for the hcloudmachines API.
--- a/api/v1beta1/hcloudmachine_types.go
+++ b/api/v1beta1/hcloudmachine_types.go
@@ -207,6 +207,10 @@ type HCloudMachineStatusExternalIDs struct {
 	// ActionIDEnableRescueSystem is the hcloud API Action result of EnableRescueSystem.
 	// +optional
 	ActionIDEnableRescueSystem int64 `json:"actionIdEnableRescueSystem,omitzero"`
+
+	// ActionIDCreate is the hcloud API Action result of CreateServer.
+	// +optional
+	ActionIDCreate int64 `json:"actionIdCreate,omitzero"`
 }
 
 // HCloudMachine is the Schema for the hcloudmachines API.
--- a/pkg/services/hcloud/client/client.go
+++ b/pkg/services/hcloud/client/client.go
@@ -54,7 +54,7 @@ type Client interface {
 	AddServiceToLoadBalancer(context.Context, *hcloud.LoadBalancer, hcloud.LoadBalancerAddServiceOpts) error
 	DeleteServiceFromLoadBalancer(context.Context, *hcloud.LoadBalancer, int) error
 	ListImages(context.Context, hcloud.ImageListOpts) ([]*hcloud.Image, error)
-	CreateServer(context.Context, hcloud.ServerCreateOpts) (*hcloud.Server, error)
+	CreateServer(context.Context, hcloud.ServerCreateOpts) (hcloud.ServerCreateResult, error)
 	AttachServerToNetwork(context.Context, *hcloud.Server, hcloud.ServerAttachToNetworkOpts) error
 	ListServers(context.Context, hcloud.ServerListOpts) ([]*hcloud.Server, error)
 	GetServer(context.Context, int64) (*hcloud.Server, error)
@@ -227,9 +227,9 @@ func (c *realClient) ListImages(ctx context.Context, opts hcloud.ImageListOpts) 
 	return c.client.Image.AllWithOpts(ctx, opts)
 }
 
-func (c *realClient) CreateServer(ctx context.Context, opts hcloud.ServerCreateOpts) (*hcloud.Server, error) {
+func (c *realClient) CreateServer(ctx context.Context, opts hcloud.ServerCreateOpts) (hcloud.ServerCreateResult, error) {
 	res, _, err := c.client.Server.Create(ctx, opts)
-	return res.Server, err
+	return res, err
 }
 
 func (c *realClient) AttachServerToNetwork(ctx context.Context, server *hcloud.Server, opts hcloud.ServerAttachToNetworkOpts) error {
@@ -357,6 +357,11 @@ func (c *realClient) GetAction(ctx context.Context, actionID int64) (*hcloud.Act
 			return action, fmt.Errorf("%w: getting hcloud action failed: %w", ErrUnauthorized, err)
 		}
 		return action, fmt.Errorf("getting hcloud action failed: %w", err)
+	}
+	// GetByID returns a nil action and no error when the action does not exist.
+	// Return an error instead, so callers can rely on the action being non-nil when err is nil.
+	if action == nil {
+		return nil, fmt.Errorf("hcloud action %d not found", actionID)
 	}
 	return action, nil
 }
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -615,7 +615,7 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 			})
 			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 		}
-		return res, fmt.Errorf("EnableRescueSystem failed: %w", err)
+		return res, handleRateLimit(hm, err, "EnableRescueSystem", "failed to enable rescue system")
 	}
 
 	// The API of hetzner is async. We get an Action-ID as result. We need to wait until the action
@@ -1705,6 +1705,7 @@ func (s *Service) createServerFromImageName(ctx context.Context) (*hcloud.Server
 		return nil, nil, err
 	}
 
+	// The imageName flow does not use the rescue system, so the server can start right after create.
 	result, err := s.createServer(ctx, userData, image, true)
 	if err != nil {
 		return nil, nil, err
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -75,6 +75,24 @@ type Service struct {
 	scope *scope.MachineScope
 }
 
+// setBootState sets the BootState and logs the transition, together with how long the
+// machine was in the previous state.
+func (s *Service) setBootState(bootState infrav1.HCloudBootState) {
+	hm := s.scope.HCloudMachine
+	if hm.Status.BootState == bootState {
+		return
+	}
+	durationOfPreviousState := time.Duration(0)
+	if !hm.Status.BootStateSince.IsZero() {
+		durationOfPreviousState = time.Since(hm.Status.BootStateSince.Time).Round(time.Millisecond)
+	}
+	s.scope.Info("BootState changed",
+		"from", hm.Status.BootState,
+		"to", bootState,
+		"durationOfPreviousState", durationOfPreviousState)
+	hm.SetBootState(bootState)
+}
+
 // NewService outs a new service with machine scope.
 func NewService(scope *scope.MachineScope) *Service {
 	return &Service{
@@ -195,6 +213,14 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		}
 	}
 
+	sinceBootStateChanged := time.Duration(0)
+	if !s.scope.HCloudMachine.Status.BootStateSince.IsZero() {
+		sinceBootStateChanged = time.Since(s.scope.HCloudMachine.Status.BootStateSince.Time).Round(time.Millisecond)
+	}
+	s.scope.V(1).Info("Reconciling BootState",
+		"bootState", s.scope.HCloudMachine.Status.BootState,
+		"sinceBootStateChanged", sinceBootStateChanged)
+
 	switch s.scope.HCloudMachine.Status.BootState {
 	case infrav1.HCloudBootStateUnset:
 		return s.handleBootStateUnset(ctx)
@@ -269,9 +295,9 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 
 		var msg string
 		if !hm.Status.Ready {
-			hm.SetBootState(infrav1.HCloudBootStateBootingToRealOS)
+			s.setBootState(infrav1.HCloudBootStateBootingToRealOS)
 		} else {
-			hm.SetBootState(infrav1.HCloudBootStateOperatingSystemRunning)
+			s.setBootState(infrav1.HCloudBootStateOperatingSystemRunning)
 		}
 		msg = fmt.Sprintf("Updating old resource (pre BootState) %s", hm.Status.BootState)
 
@@ -595,7 +621,7 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 	// is done. After that we can power the server on, so that it boots into the rescue system.
 	hm.Status.ExternalIDs.ActionIDEnableRescueSystem = result.Action.ID
 
-	hm.SetBootState(infrav1.HCloudBootStateEnablingRescue)
+	s.setBootState(infrav1.HCloudBootStateEnablingRescue)
 
 	v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
 		"WaitForRescueSystem", clusterv1beta1.ConditionSeverityInfo,
@@ -805,7 +831,7 @@ func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcl
 
 	s.scope.Info("Power on started (first boot, into rescue system)")
 
-	hm.SetBootState(infrav1.HCloudBootStateBootingToRescue)
+	s.setBootState(infrav1.HCloudBootStateBootingToRescue)
 	v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
 		"BootingToRescue", clusterv1beta1.ConditionSeverityInfo,
 		"power on to rescue started")
@@ -1010,7 +1036,7 @@ func (s *Service) handleBootStateBootingToRescue(ctx context.Context, server *hc
 		Reason:  infrav1.HCloudMachineHCloudImageURLCommandRunningV1Beta2Reason,
 		Message: "custom provisioner running",
 	})
-	hm.SetBootState(infrav1.HCloudBootStateRunningImageCommand)
+	s.setBootState(infrav1.HCloudBootStateRunningImageCommand)
 	return reconcile.Result{RequeueAfter: 55 * time.Second}, nil
 }
 
@@ -1124,7 +1150,7 @@ func (s *Service) handleBootStateRunningImageCommand(ctx context.Context, server
 			return reconcile.Result{}, fmt.Errorf("reboot after ImageURLCommand failed: %w", rebootErr)
 		}
 
-		hm.SetBootState(infrav1.HCloudBootStateBootingToRealOS)
+		s.setBootState(infrav1.HCloudBootStateBootingToRealOS)
 
 		v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
 			"BootingToRealOS", clusterv1beta1.ConditionSeverityInfo,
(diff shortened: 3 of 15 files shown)


api/v1beta1/hcloudmachine_types.go
```diff
@@ -207,6 +207,10 @@ type HCloudMachineStatusExternalIDs struct {
 	// ActionIDEnableRescueSystem is the hcloud API Action result of EnableRescueSystem.
 	// +optional
 	ActionIDEnableRescueSystem int64 `json:"actionIdEnableRescueSystem,omitzero"`
+
+	// ActionIDCreate is the hcloud API Action result of CreateServer.
+	// +optional
+	ActionIDCreate int64 `json:"actionIdCreate,omitzero"`
```
janiskemper: I think it's appropriate to name it "CreateServer", because it is part of the HCloudMachine object - aka a Kubernetes object - while we talk here about a VM on Hetzner. 

EDIT:
You can also check all conditions etc. where you talk about the action

pkg/services/hcloud/client/client.go
```diff
@@ -358,5 +358,10 @@ func (c *realClient) GetAction(ctx context.Context, actionID int64) (*hcloud.Act
 		}
 		return action, fmt.Errorf("getting hcloud action failed: %w", err)
 	}
+	// GetByID returns a nil action and no error when the action does not exist.
+	// Return an error instead, so callers can rely on the action being non-nil when err is nil.
+	if action == nil {
```
janiskemper, line 363: is that maybe dangerous because the server exists already and therefore it is not idempotent

abdullah599:
here is every wait in the new flow. All of this applies only to machines with `spec.imageURL`, because only they boot through the rescue system. Machines with `spec.imageName` are unchanged: they get created with `start_after_create=true` and boot the real image directly. Each state also keeps its own 6 minute timeout. a machine that does not leave a state in time gets remediated.

## Unset: create the server

For an `imageURL` machine the server gets created with `start_after_create=false` and the create action ID is stored in the status. After the create call the controller waits 15 seconds before the next reconcile. 

> [!note]
> The create action took around 4-6 seconds for a small server in the PoC. It is set to 15 seconds to leave enough margin before the first check.

## Initializing: wait for the create action, enable rescue

| When | Requeue |
|---|---|
| create action not finished yet | 5 seconds |
| `EnableRescueSystem` returns locked | 5 seconds |
| enable rescue accepted, action started | 1 second, then `EnablingRescue` |
| `GetAction` fails with an unexpected error | error is returned, controller backoff |
| create action finished with an error, or the stored action ID is missing | no requeue, machine gets remediated |

Once the create action is finished it gets marked as done in the status, so a locked retry of `EnableRescueSystem` does not fetch the finished action again. The 1 second requeue at the end just hands off to `EnablingRescue`, which waits for the enable rescue action (about 3 seconds in the PoC).

## EnablingRescue: wait for the enable rescue action, power on

| When | Requeue |
|---|---|
| enable rescue action not finished yet | 5 seconds |
| action finished | 1 second, so the next reconcile reads a fresh server object and the `EnablingRescueActionDone` condition is visible |
| `RescueEnabled` still false on the fresh read | 10 seconds |
| `PowerOnServer` returns locked | 5 seconds |
| power on accepted | 10 seconds, state becomes `BootingToRescue` |

## BootingToRescue: wait until the server is in rescue

| When | Requeue |
|---|---|
| `RescueEnabled` still true, server still booting | 5 seconds |
| SSH not reachable yet | 5 seconds |
| hostname check failed | 5 seconds |

The boot into rescue took about 30 seconds in the PoC, so this state polls about six times. Everything after this state is unchanged by this PR.

pkg/services/hcloud/server/server.go
```diff
@@ -384,15 +384,22 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
 	// These values get only used **once** after the server got created.
-
-	requeueAfter := 140 * time.Second
-	if image.RapidDeploy {
-		requeueAfter = 10 * time.Second
+	var requeueAfter time.Duration
+	if hm.Status.BootState == infrav1.HCloudBootStateInitializing {
+		// The imageURL flow created the server powered off. Only the create action needs
+		// to finish before the rescue system can be enabled, which takes a few seconds.
+		requeueAfter = 5 * time.Second
```
janiskemper: 15 seconds are fine here (Sven said it)

pkg/services/hcloud/server/server.go
janiskemper: we can reduce this to 1 second I guess or requeueImmediately

janiskemper: let's do 5 sec here

pkg/services/hcloud/server/server.go
```diff
@@ -946,7 +1036,7 @@ func (s *Service) handleBootStateBootingToRescue(ctx context.Context, server *hc
 		Reason:  infrav1.HCloudMachineHCloudImageURLCommandRunningV1Beta2Reason,
 		Message: "custom provisioner running",
 	})
-	hm.SetBootState(infrav1.HCloudBootStateRunningImageCommand)
+	s.setBootState(infrav1.HCloudBootStateRunningImageCommand)
 	return reconcile.Result{RequeueAfter: 55 * time.Second}, nil
```
janiskemper: let's do 20 here

pkg/services/hcloud/server/server.go
janiskemper: let's do 10 here

pkg/services/hcloud/server/server.go
```diff
@@ -1060,7 +1150,7 @@ func (s *Service) handleBootStateRunningImageCommand(ctx context.Context, server
 			return reconcile.Result{}, fmt.Errorf("reboot after ImageURLCommand failed: %w", rebootErr)
 		}
 
-		hm.SetBootState(infrav1.HCloudBootStateBootingToRealOS)
+		s.setBootState(infrav1.HCloudBootStateBootingToRealOS)
```
janiskemper, line 1157: we requeue immediately after doing the ssh reboot and then in the next step we just check whether the server is showing status "running". Would this be done immediately, even if the ssh reboot is ongoing? We should document that.

pkg/services/hcloud/client/client.go
janiskemper: I realized that this is not part of the CreateServer method, so it doens't matter.

pkg/services/hcloud/server/server.go
```diff
@@ -384,15 +410,22 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
-	// Provisioning from a snapshot takes roughly 140 seconds.
-	// We do not want to do too many api-calls (rate-limiting). So we differentiate
-	// between both cases.
 	// These values get only used **once** after the server got created.
-
-	requeueAfter := 140 * time.Second
-	if image.RapidDeploy {
-		requeueAfter = 10 * time.Second
+	var requeueAfter time.Duration
+	if hm.Status.BootState == infrav1.HCloudBootStateInitializing {
```
janiskemper: you ask in the part of the code that is handling one state whether the state is something else :D You do this because you know that the state was updated already after the server was created, right? But this is not clear here at all.

pkg/services/hcloud/server/server.go
```diff
@@ -452,36 +485,109 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
...
+				Reason:  infrav1.HCloudMachineGettingCreateActionFailedV1Beta2Reason,
+				Message: err.Error(),
+			})
+			return reconcile.Result{}, err
+		}
+
+		if action.Finished.IsZero() {
+			// not finished yet.
+			v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
+				"WaitingForCreateAction", clusterv1beta1.ConditionSeverityInfo,
```
janiskemper: "CreatingServer" is maybe a nicer way to formulate this for users. Btw: this reason is not stored as constant, like the others.

Can you maybe go through the code and double check this?

pkg/services/hcloud/server/server.go
```diff
@@ -452,36 +485,109 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
...
+		if err != nil {
+			err = fmt.Errorf("action %+v failed (wait for server create): %w", action, err)
+			msg := err.Error()
+			s.scope.Error(err, "")
+			remediateErr := s.scope.SetErrorAndRemediate(ctx, msg)
+			if remediateErr != nil {
+				return reconcile.Result{}, remediateErr
+			}
+			v1beta1conditions.MarkFalse(hm, infrav1.ServerProvisionedCondition,
+				"CreateActionFailed", clusterv1beta1.ConditionSeverityWarning,
```
janiskemper: is it ServerCreationFailed or CreateActionFailed?

pkg/services/hcloud/server/server.go
```diff
@@ -688,68 +808,38 @@ func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcl
...
-			Status:  metav1.ConditionUnknown,
-			Reason:  infrav1.HCloudMachineRebootViaSSHFailedV1Beta2Reason,
-			Message: err.Error(),
-		})
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		return reconcile.Result{}, handleRateLimit(hm, err, "PowerOnServer", "failed to power on server")
 	}
 
-	s.scope.Info("Reboot started (via ssh)")
+	s.scope.Info("Power on started (first boot, into rescue system)")
```
janiskemper: I think we can remove this log

pkg/services/hcloud/server/server.go
janiskemper: @abdullah599 the reasons are not constants. Can you check whether we miss this in a lot of places? If all others are pretty much defined as constants, we should do it here as well for sure.

pkg/services/hcloud/server/server.go
janiskemper: @abdullah599 IMO this should be "server create failed", or not? Because "action failed" sounds a bit as if it is not the same

pkg/services/hcloud/server/server.go
```diff
@@ -1605,19 +1705,18 @@ func (s *Service) createServerFromImageName(ctx context.Context) (*hcloud.Server
 		return nil, nil, err
 	}
 
-	server, err := s.createServer(ctx, userData, image)
+	result, err := s.createServer(ctx, userData, image, true)
```
Dhairya-Arora01, line 1709: Also add a comment on here about why we set `startAfterCreate` to true.

pkg/services/hcloud/server/server.go
abdullah599: Discussed in chat: Only v1beta2 reasons should be constant

pkg/services/hcloud/server/server.go
```diff
@@ -494,15 +601,28 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
...
+			v1beta2conditions.Set(hm, metav1.Condition{
+				Type:    infrav1.HCloudMachineServerProvisionedV1Beta2Condition,
+				Status:  metav1.ConditionFalse,
+				Reason:  infrav1.HCloudMachineEnablingRescueSystemFailedV1Beta2Reason,
+				Message: "EnableRescueSystem: server locked. Will retry",
+			})
+			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+		}
 		return res, fmt.Errorf("EnableRescueSystem failed: %w", err)
 	}
```
Dhairya-Arora01, line 619: shouldn't we handle rate limit here ?

Dhairya-Arora01 (dismissed):

Dhairya-Arora01 (approved):

PR #2164, opened 2026-07-13
Set fatal error instead of permanent/provisioning error when reboot times out

### LINKED ISSUE #2147: Set a fatal error instead of a permanent error when a reboot times out
## Summary

When the reboot into the rescue system times out during host registration, set a fatal error instead of a permanent error, so the host is deprovisioned and returned to the pool for another machine to pick up, instead of being removed from the pool until a human clears it. The server reboots a second time into the installed OS after image install; when that reboot times out, set a fatal error there too, instead of leaving the machine to retry on the same host.

## The Problem

During provisioning, CAPH reboots a bare metal server into Hetzner's rescue system so it can install the image. The `actionRegistering` step waits for the server to come up in rescue and escalates the reboot step by step (ssh → software → hardware). If the hardware reboot does not reach rescue within the timeout (`hardwareResetTimeout`, 10 minutes), CAPH gives up and records a **permanent error** with the message:

> The rescue system could not be reached. Please ensure that the machine tries to boot from network before booting from disk. This setting needs to be enabled permanently in the BIOS.

A permanent error and a fatal error behave differently:

- Both delete the associated `HetznerBareMetalMachine`.
- A **fatal error** is cleared when the host is deprovisioned. The host goes back to `StateNone` with an empty error message and becomes available again, so another machine can pick it up.
- A **permanent error** is *not* cleared on deprovisioning. The host keeps its error message and the `capi.syself.com/permanent-error` annotation. Host selection skips any host that has an error message, so the host stays out of the pool until someone removes the annotation by hand.

The problem is that "the rescue system could not be reached" is usually **not** a permanent condition. In most cases the server was just slow to reboot. Treating a slow reboot as a permanent error takes a healthy host out of the pool and needs manual work to bring it back. This false positive happens often.

The server reboots a second time into the installed OS after the image is installed. The `actionEnsureProvisioned` step waits for the server to come up with the expected hostname and escalates the reboot the same way. If that reboot times out, CAPH records a **provisioning error** with the message:

> reboot to node timed out - please check if server is working properly

This does not delete the machine and does not return the host to the pool. The machine just keeps retrying on the same host with a backoff that grows to hours. This is the same slow-reboot situation and should be handled the same way.

## The Solution

Stop setting a permanent error when the reboot into rescue times out, and stop setting a provisioning error when the reboot into the OS times out. Set a **fatal error** in both cases instead.

- The machine is still deleted, same as today.
- The host is deprovisioned, its error is cleared, and it returns to the pool.
- A new machine picks up the host and retries provisioning from the start.

This trades a false positive that needs manual recovery for an automatic retry.

**Accepted drawback:** a server that truly cannot boot into rescue (for example a wrong or unstable boot order in the BIOS) will now retry forever: provision → reboot times out → fatal error → machine deleted → host deprovisioned → host picked up again → provision. This is acceptable. Such a server is visibly stuck, so the problem is not hidden. The longer term fix is a separate initial check that verifies a new server can reach rescue consistently before it is added to the pool.

### 1. Change the error type for the reboot timeout

In `pkg/services/baremetal/host/host.go`, `actionRegistering` turns a failed boot into rescue into a permanent error:

```go
failed, err := s.handleIncompleteBoot(ctx, true, isSSHTimeoutError, isSSHConnectionRefusedError)
if failed {
    return s.recordActionFailure(infrav1.PermanentError, err.Error())
}
```

Change `infrav1.PermanentError` to `infrav1.FatalError` in this call.

This single call covers the reboot-into-rescue timeout (the message above) and the related "connection refused, cannot reach rescue via ssh" case. Both mean "cannot reach rescue during registration" and both should become fatal errors.

### 2. Change the error type for the OS reboot timeout

In `pkg/services/baremetal/host/host.go`, `actionEnsureProvisioned` turns a failed boot into the OS into a provisioning error:

```go
failed, err := s.handleIncompleteBoot(ctx, false, isTimeout, isSSHConnectionRefusedError)
if failed {
    msg := "reboot handling failed"
    if err != nil {
        msg = err.Error()
    }
    markProvisionPendingWithInfo(s.scope.HetznerBareMetalHost, infrav1.StateEnsureProvisioned, msg)
    return s.recordActionFailure(infrav1.ProvisioningError, msg)
}
```

Change `infrav1.ProvisioningError` to `infrav1.FatalError` in this call.

Like the rescue case, this branch also covers the "connection refused, cannot reach the OS via ssh" case. Both should become fatal errors.

## Expected Behavior

| Scenario | Behavior |
|----------|----------|
| Reboot into rescue times out (slow server) | Fatal error set → machine deleted → host deprovisioned and cleared → host picked up again → provisioning retried |
| Reboot into the OS times out (slow server) | Same: fatal error set → machine deleted → host deprovisioned and cleared → host picked up again → provisioning retried |
| Server can never reach rescue (wrong boot order) | The same loop repeats indefinitely; the server stays visibly stuck instead of silently leaving the pool |
| Other permanent error cases (disk checks, wrong hostname, etc.) | Unchanged: still set a permanent error that stays on the host until cleared manually |

guettli:
## Summary
- When the reboot into the rescue system times out during host registration (`actionRegistering`), set a fatal error instead of a permanent error.
- When the reboot into the installed OS times out (`actionEnsureProvisioned`), set a fatal error instead of a provisioning error.

A fatal error deletes the machine and deprovisions the host, clearing its error so it returns to the pool for another machine to pick up and retry — instead of requiring manual clearing of the `permanent-error` annotation, or retrying forever on the same host with a growing backoff.

**Accepted drawback (from the issue):** a server that truly cannot boot into rescue will retry forever (provision → timeout → fatal error → delete → deprovision → re-pick-up → provision). This is acceptable since the server stays visibly stuck rather than silently leaving the pool. A follow-up initial-check mechanism is out of scope here.

Other permanent error cases (disk checks, wrong hostname, etc.) are unchanged.

Fixes #2147

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -668,7 +668,7 @@ func (s *Service) actionRegistering(ctx context.Context) actionResult {
 
 		failed, err := s.handleIncompleteBoot(ctx, true, isSSHTimeoutError, isSSHConnectionRefusedError)
 		if failed {
-			return s.recordActionFailure(infrav1.PermanentError, err.Error())
+			return s.recordActionFailure(infrav1.FatalError, err.Error())
 		}
 		if err != nil {
 			return actionError{err: fmt.Errorf(errMsgFailedHandlingIncompleteBoot, err)}
@@ -2047,7 +2047,7 @@ func (s *Service) actionEnsureProvisioned(ctx context.Context) (ar actionResult)
 			}
 			markProvisionPendingWithInfo(s.scope.HetznerBareMetalHost,
 				infrav1.StateEnsureProvisioned, msg)
-			return s.recordActionFailure(infrav1.ProvisioningError, msg)
+			return s.recordActionFailure(infrav1.FatalError, msg)
 		}
 		if err != nil {
 			markProvisionPendingWithInfo(s.scope.HetznerBareMetalHost,
--- a/pkg/services/baremetal/host/host_test.go
+++ b/pkg/services/baremetal/host/host_test.go
@@ -1585,6 +1585,29 @@ var _ = Describe("actionRegistering", func() {
 			expectedErrorType: infrav1.ErrorTypeConnectionError,
 		}),
 	)
+
+	It("sets a fatal error when the reboot into rescue times out", func() {
+		host := helpers.BareMetalHost(
+			"test-host",
+			"default",
+			helpers.WithRootDeviceHintWWN(),
+			helpers.WithIPv4(),
+			helpers.WithConsumerRef(),
+			helpers.WithError(infrav1.ErrorTypeHardwareRebootTriggered, "", 1),
+			helpers.WithRebootTriggeredAt(metav1.NewTime(time.Now().Add(-time.Hour))),
+		)
+
+		sshMock := &sshmock.Client{}
+		sshMock.On("GetHostName", mock.Anything).Return(sshclient.Output{Err: timeout})
+
+		service := newTestService(host, &robotmock.Client{}, bmmock.NewSSHFactory(sshMock, sshMock, sshMock), nil, helpers.GetDefaultSSHSecret(rescueSSHKeyName, "default"))
+
+		actResult := service.actionRegistering(ctx)
+
+		Expect(actResult).To(BeAssignableToTypeOf(actionFailed{}))
+		Expect(host.Spec.Status.ErrorType).To(Equal(infrav1.FatalError))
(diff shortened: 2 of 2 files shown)


janiskemper (approved):

guettli:
Test was flaky. I hope we can that to fix flakiness soon: 
[:seedling: fix flaky test via RWMutex by guettli · Pull Request #2036 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/2036#issuecomment-4798535802)

PR #2167, opened 2026-07-13
🐛 Fix server creation getting stuck forever on name-uniqueness error

guettli:
## Summary
- The `createServer` function called the HCloud API's `CreateServer` without handling the case where a server with this exact name already exists. Since the server name is deterministic (`HCloudMachine.Name`), if a previous reconcile created the server successfully but lost the update that would have persisted `ProviderID`/`BootState` (e.g. an API server conflict or controller restart between `CreateServer` and `Close()`), every subsequent reconcile retried `CreateServer` with the same name and failed forever with `server name is already used (uniqueness_error)`.
- This is a regression from #1650 (Refactor Hcloud Provisioning, use Status.BootState), which gated the existing `findServer` lookup (which has a label-based fallback) behind `ProviderID != nil`, removing the safety net for the initial creation path. `release-1.0` forked before that refactor and is unaffected; `main` and branches based on it are affected.
- Fix: when `CreateServer` fails specifically with `hcloud.ErrorCodeUniquenessError`, look up the existing server by name (`hcloud.ServerListOpts{Name: ...}`, an exact-match filter) via a new standalone `findServerByName` helper, and adopt it instead of failing. This is handled entirely inside `createServer()` — no change needed to `handleBootStateUnset` or either creation flow (imageName/imageURL), since both already consume `createServer`'s return value the same way whether the server was just created or adopted.
- `findServer`'s existing label-based fallback is left untouched — the diff is a pure addition (0 lines removed from `server.go`).
- Matching the sibling create-failure/success paths in the same function, adoption both logs and emits a Kubernetes Event (`AdoptedExistingServer`); a failure in the recovery lookup itself is logged rather than silently discarded.

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -1837,7 +1837,7 @@ func (s *Service) createServer(ctx context.Context, userData []byte, image *hclo
 			// concrete fix instead of the raw uniqueness error.
 			msg = fmt.Sprintf(
 				"Server creation failed because a server named %q already exists and it could not be adopted automatically: %s. "+
-					"Delete the conflicting HCloud server, or delete this Machine so its replacement is created with a new name.",
+					"Check for a dangling HCloud server with this name and delete it, or delete this Machine so its replacement is created with a new name.",
 				hm.Name, err.Error())
 		}
 		s.scope.Error(nil, msg)
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -351,25 +351,42 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 			return reconcile.Result{}, nil
 		}
 
+		// A uniqueness_error that createServer's own recovery (adopting the existing server) could
+		// not resolve means a server with this name already exists but is not this machine's. Unlike
+		// invalid_input/resource_unavailable, this can clear from the outside when the conflicting
+		// server is deleted. Nothing watches HCloud, so we requeue on a slow interval to retry the
+		// create, otherwise deleting the conflicting server would never trigger a new attempt.
+		if hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
+			msg := fmt.Sprintf(
+				"a server named %q already exists and could not be adopted, retrying: %s. "+
+					"Delete the conflicting HCloud server, or delete this Machine to create a new one.",
+				s.scope.Name(), err.Error(),
+			)
+			v1beta1conditions.MarkFalse(
+				s.scope.HCloudMachine,
+				infrav1.ServerCreateSucceededCondition,
+				infrav1.ServerCreateFailedReason,
+				clusterv1beta1.ConditionSeverityWarning,
+				"%s",
+				msg,
+			)
+			v1beta2conditions.Set(s.scope.HCloudMachine, metav1.Condition{
+				Type:    infrav1.HCloudMachineServerCreatedV1Beta2Condition,
+				Status:  metav1.ConditionFalse,
+				Reason:  infrav1.HCloudMachineServerCreationFailedV1Beta2Reason,
+				Message: msg,
+			})
+			return reconcile.Result{RequeueAfter: 10 * time.Minute}, nil
+		}
+
 		// Terminal errors like invalid_input (e.g. unsupported location for server type)
 		// or resource_unavailable (e.g. server location disabled) will never succeed on retry.
-		// A uniqueness_error that createServer's own recovery (adopting the existing server)
-		// could not resolve is equally terminal: retrying CreateServer with the same name will
-		// hit the exact same error forever. Mark the machine as irrecoverably failed and stop
-		// reconciling in all these cases.
-		if hcloud.IsError(err, hcloud.ErrorCodeInvalidInput) || hcloud.IsError(err, hcloud.ErrorCodeResourceUnavailable) ||
-			hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
+		// Mark the machine as irrecoverably failed and stop reconciling.
+		if hcloud.IsError(err, hcloud.ErrorCodeInvalidInput) || hcloud.IsError(err, hcloud.ErrorCodeResourceUnavailable) {
 			msg := fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -1800,6 +1800,31 @@ func (s *Service) createServer(ctx context.Context, userData []byte, image *hclo
...
 		}
 
+		// A server with this exact name already exists. This happens if a previous reconcile
+		// created the HCloud server successfully but lost the update that would have persisted
+		// ProviderID/BootState (e.g. an API server conflict or a controller restart between
+		// CreateServer and Close()). Adopt the existing server instead of failing forever with
+		// the same uniqueness error on every retry.
+		if hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
+			existingServer, findErr := s.findServerByName(ctx)
+			if findErr != nil {
```
janiskemper, line 1801: to complete this, wouldn't it be the best idea to mark the machine as failed, if this "taking over" doesn't work?

guettli: Good catch — addressed in ec1c453c8. If the recovery lookup itself fails, or turns up no matching server, retrying `CreateServer` with the same name will just hit the identical `uniqueness_error` forever. That's now folded into the same irrecoverable-error handling in `Reconcile` that already exists for `invalid_input`/`resource_unavailable`: the machine gets `ServerCreateSucceededCondition=False` with reason `ServerCreateFailedIrrecoverableErrorReason`, and reconciliation stops instead of looping forever. This also means the remediation controller correctly skips remediation for it, same as the other irrecoverable cases.

janiskemper (dismissed):

pkg/services/hcloud/server/server_test.go
```diff
@@ -1255,6 +1255,227 @@ var _ = Describe("Reconcile", func() {
...
+		}, nil)
+
+		By("calling reconcile")
+		_, err = service.Reconcile(ctx)
+		Expect(err).To(BeNil())
+
+		By("ensuring the existing server was adopted instead of failing forever")
+		Expect(service.scope.HCloudMachine.Status.BootState).To(Equal(infrav1.HCloudBootStateBootingToRealOS))
+		Expect(*service.scope.HCloudMachine.Spec.ProviderID).To(Equal("hcloud://42"))
+		Expect(v1beta1conditions.IsTrue(service.scope.HCloudMachine, infrav1.ServerCreateSucceededCondition)).To(BeTrue())
```
abdullah599, line 1490: This asserts the v1beta1 condition but not the v1beta2 one the adoption path sets. The rest of the file pairs both with `isPresentWithStatusAndReasonV1Beta2`, since a v1beta2 side regression would pass silently otherwise. Can we add the `HCloudMachineServerCreatedV1Beta2Condition` check here and in the other two new tests?

pkg/services/hcloud/server/server.go
```diff
@@ -353,12 +353,23 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
...
+		if hcloud.IsError(err, hcloud.ErrorCodeInvalidInput) || hcloud.IsError(err, hcloud.ErrorCodeResourceUnavailable) ||
+			hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
 			msg := fmt.Sprintf(
 				"Server creation failed with an irrecoverable error: %s. If the requested resources (server type or location) become available again, delete the Machine to trigger a new creation attempt.",
 				err.Error(),
 			)
+			if hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
+				msg = fmt.Sprintf(
+					"Server creation failed because a server named %q already exists and it could not be adopted automatically: %s. "+
+						"Delete the conflicting HCloud server, or delete this Machine to trigger a new creation attempt.",
```
abdullah599: Deleting the conflicting server can not retrigger a reconcile (no hcloud watch, status only changes are filtered), so this leaves the machine stuck. Can we return a bounded RequeueAfter instead of reconcile.Result{}, nil?

janiskemper: The MachineHealthChecks would also trigger. However, I think it would be okay to do. If it adds complexity I'm totally fine leaving it as is. We should maybe reformulate the comment though, because the statement is a bit misleading. If you delete the Machine manually after the server is deleted, then it would work. Your call - if you wanna add the requeue (maybe 10 min) or just reformulate the sentence

pkg/services/hcloud/server/server_test.go
janiskemper: sure

abdullah599 (dismissed):

pkg/services/hcloud/server/server.go
```diff
@@ -351,6 +351,34 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 			return reconcile.Result{}, nil
 		}
 
+		// A uniqueness_error that createServer's own recovery (adopting the existing server) could
+		// not resolve means a server with this name already exists but is not this machine's. Unlike
+		// invalid_input/resource_unavailable, this can clear from the outside when the conflicting
+		// server is deleted. Nothing watches HCloud, so we requeue on a slow interval to retry the
+		// create, otherwise deleting the conflicting server would never trigger a new attempt.
+		if hcloud.IsError(err, hcloud.ErrorCodeUniquenessError) {
```
janiskemper, line 290: I don't really understand why you overwrite the condition here again, after it was already written in createServer.. If your goal is to requeue, then it's enough to do this IMO.

janiskemper (changes requested):

pkg/services/hcloud/server/server.go
```diff
@@ -1837,7 +1837,7 @@ func (s *Service) createServer(ctx context.Context, userData []byte, image *hclo
 			// concrete fix instead of the raw uniqueness error.
 			msg = fmt.Sprintf(
 				"Server creation failed because a server named %q already exists and it could not be adopted automatically: %s. "+
-					"Delete the conflicting HCloud server, or delete this Machine to trigger a new creation attempt.",
+					"Delete the conflicting HCloud server, or delete this Machine so its replacement is created with a new name.",
```
janiskemper: addition: something like "Check for dangling servers."

janiskemper (dismissed):

abdullah599 (approved):

janiskemper (approved):

PR #2171, opened 2026-07-15
Reduce hcloud GetServer calls during machine provisioning

guettli:
## Summary
- `Reconcile()` fetched the live hcloud server unconditionally at the top of every call, even for `BootState`s (`Initializing`, `EnablingRescue`, `BootingToRescue`, `RunningImageCommand`) that only need `GetAction` polling or SSH to make progress.
- Skips the `GetServer` fetch for those four states, and drops the now-redundant `server.RescueEnabled` re-checks — the `enable_rescue` action result is already the authoritative signal, so `BootingToRescue` goes straight to SSH (an `ECONNREFUSED` retry already covered "not rebooted yet").
- The two remaining hcloud calls that only needed `server.ID` (`EnableRescueSystem`, `PowerOnServer`) now use a `server_id`-only stub built from `Spec.ProviderID`, avoiding a live fetch just to get an ID we already know.
- Tightens `RequeueAfter` to immediate for the states that are now hcloud-API-free (pure SSH), so provisioning can proceed as fast as SSH allows instead of waiting on a fixed interval. States that still poll `GetAction`/`GetServer` keep their existing interval, since tightening those would multiply hcloud API calls instead of reducing them.
- Net effect: `GetServer` calls during a full imageURL provisioning run drop from 7 to 1 (only the `BootingToRealOS` poll remains, since `server.Status` has no SSH substitute).

Relates to #2163

### DIFF
--- a/pkg/services/hcloud/server/server.go
+++ b/pkg/services/hcloud/server/server.go
@@ -144,7 +144,17 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	var server *hcloud.Server
 
-	if s.scope.HCloudMachine.Spec.ProviderID != nil {
+	// Only fetch the live server for the states that actually need it (an "IsCreated" check,
+	// a server.Status transition, or the network/LB attachment reconcile). The remaining states
+	// (Initializing, EnablingRescue, BootingToRescue, RunningImageCommand) drive their progress
+	// via GetAction polling and/or SSH, so skipping the fetch there avoids an hcloud API call on
+	// every reconcile while we wait (see issue #2163).
+	bootState := s.scope.HCloudMachine.Status.BootState
+	needsLiveServer := bootState == infrav1.HCloudBootStateUnset ||
+		bootState == infrav1.HCloudBootStateBootingToRealOS ||
+		bootState == infrav1.HCloudBootStateOperatingSystemRunning
+
+	if s.scope.HCloudMachine.Spec.ProviderID != nil && needsLiveServer {
 		server, err = s.findServer(ctx)
 		if err != nil {
 			// If it is an unauthorized error i.e. wrong HCloudToken do not return an error.
@@ -221,23 +231,23 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 		"bootState", s.scope.HCloudMachine.Status.BootState,
 		"sinceBootStateChanged", sinceBootStateChanged)
 
-	switch s.scope.HCloudMachine.Status.BootState {
+	switch bootState {
 	case infrav1.HCloudBootStateUnset:
 		return s.handleBootStateUnset(ctx)
 	case infrav1.HCloudBootStateInitializing:
-		return s.handleBootStateInitializing(ctx, server)
+		return s.handleBootStateInitializing(ctx)
 	case infrav1.HCloudBootStateEnablingRescue:
-		return s.handleBootStateEnablingRescue(ctx, server)
+		return s.handleBootStateEnablingRescue(ctx)
 	case infrav1.HCloudBootStateBootingToRescue:
-		return s.handleBootStateBootingToRescue(ctx, server)
+		return s.handleBootStateBootingToRescue(ctx)
 	case infrav1.HCloudBootStateRunningImageCommand:
-		return s.handleBootStateRunningImageCommand(ctx, server)
+		return s.handleBootStateRunningImageCommand(ctx)
 	case infrav1.HCloudBootStateBootingToRealOS:
 		return s.handleBootingToRealOS(ctx, server)
 	case infrav1.HCloudBootStateOperatingSystemRunning:
 		return s.handleOperatingSystemRunning(ctx, server)
 	default:
-		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", s.scope.HCloudMachine.Status.BootState)
+		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", bootState)
 	}
 }
 
@@ -441,7 +451,7 @@ func (s *Service) handleBootStateUnset(ctx context.Context) (reconcile.Result, e
 }
 
 // handleBootStateInitializing is for provisioning with imageURL and image-url-command.
-func (s *Service) handleBootStateInitializing(ctx context.Context, server *hcloud.Server) (res reconcile.Result, reterr error) {
+func (s *Service) handleBootStateInitializing(ctx context.Context) (res reconcile.Result, reterr error) {
 	hm := s.scope.HCloudMachine
 
 	durationOfState := time.Since(hm.Status.BootStateSince.Time)
@@ -483,8 +493,6 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 		})
 		return reconcile.Result{}, nil
 	}
-
-	updateHCloudMachineStatusFromServer(hm, server)
 
 	// ActionIDCreateServer gets stored by createServerFromImageURL before the boot state becomes
 	// Initializing. This guard catches it early if that behavior ever changes.
@@ -599,7 +607,11 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 		Type:    hcloud.ServerRescueTypeLinux64,
 		SSHKeys: hcloudSSHKeys,
 	}
-	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, server, rescueOpts)
+	serverStub, err := s.serverStubFromProviderID()
+	if err != nil {
+		return res, fmt.Errorf("serverStubFromProviderID failed: %w", err)
+	}
+	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, serverStub, rescueOpts)
 	if err != nil {
 		if hcloud.IsError(err, hcloud.ErrorCodeLocked) {
 			// a fresh server is locked only for a short time after create, so a short retry
@@ -637,7 +649,7 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 }
 
 // handleBootStateEnablingRescue is for provisioning with imageURL and image-url-command.
-func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcloud.Server) (reconcile.Result, error) {
+func (s *Service) handleBootStateEnablingRescue(ctx context.Context) (reconcile.Result, error) {
 	hm := s.scope.HCloudMachine
 
 	durationOfState := time.Since(hm.Status.BootStateSince.Time)
@@ -678,8 +690,6 @@ func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcl
 		})
 		return reconcile.Result{}, nil
 	}
-
-	updateHCloudMachineStatusFromServer(hm, server)
 
 	// ActionIDEnableRescueSystem gets stored by handleBootStateInitializing before the boot
 	// state becomes EnablingRescue. This guard catches it early if that behavior ever changes.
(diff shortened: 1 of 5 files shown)


pkg/services/hcloud/server/server.go
```diff
@@ -144,7 +144,16 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	var server *hcloud.Server
 
-	if s.scope.HCloudMachine.Spec.ProviderID != nil {
+	// Only fetch the live server for the states that actually need it (a server.Status
+	// transition, or the network/LB attachment reconcile). The remaining states (Unset,
+	// Initializing, EnablingRescue, BootingToRescue, RunningImageCommand) drive their progress
+	// via GetAction polling and/or SSH, so skipping the fetch there avoids an hcloud API call on
+	// every reconcile while we wait (see issue #2163).
```
janiskemper: don't add issues to these comments if not necessary.

pkg/services/hcloud/server/server.go
```diff
@@ -144,7 +144,16 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
 
 	var server *hcloud.Server
 
-	if s.scope.HCloudMachine.Spec.ProviderID != nil {
+	// Only fetch the live server for the states that actually need it (a server.Status
```
janiskemper: we should re-think this approach. I see the following issues:
1. we have a lot of code before we actually look at the state machine, and the code is irrelevant for most of the states.
2. We used to have it here as central entry point and which also made sense, because we e.g. set the "HcloudTokenAvailable" condition here at the start. Since this is not actually called for most states anymore, it also doesn't make sense to use it in this way. We have to replace it anyway, because we obviously need this functionality.

Conclusion: I think we need to do these things per state. This will maybe duplicate the code of handling the situation where a server is not found, but otherwise solve the two problems above.

janiskemper (changes requested):

pkg/services/hcloud/server/server.go
guettli: done in c3692742d

pkg/services/hcloud/server/server.go
guettli: Yes, that makes sense. I had this on my mind, too.

Done in c3692742d

pkg/services/hcloud/server/server.go
```diff
@@ -141,103 +141,120 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 	case infrav1.HCloudBootStateOperatingSystemRunning:
-		return s.handleOperatingSystemRunning(ctx, server)
+		return s.handleOperatingSystemRunning(ctx)
 	default:
-		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", s.scope.HCloudMachine.Status.BootState)
+		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", bootState)
+	}
+}
+
+// resultOrError carries the (result, error) pair getLiveServer wants its caller to return
```
janiskemper: why don't we just keep the two separate? I think I would prefer having three outputs of the function rather than a separate struct.

pkg/services/hcloud/server/server.go
```diff
@@ -598,7 +613,11 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 		Type:    hcloud.ServerRescueTypeLinux64,
 		SSHKeys: hcloudSSHKeys,
 	}
-	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, server, rescueOpts)
+	serverStub, err := s.serverStubFromProviderID()
```
janiskemper: I would prefer if we inline the minimal server here, because the separate function doesn't make the code more readable. 

We can just have a function "serverIDFromProviderID" - then the code is more readable - and you don't have to check what is meant by "stub"

pkg/services/hcloud/server/server.go
```diff
@@ -793,25 +810,13 @@ func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcl
...
-			"RescueNotEnabledYet", clusterv1beta1.ConditionSeverityWarning,
-			"%s", msg)
-		v1beta2conditions.Set(hm, metav1.Condition{
-			Type:    infrav1.HCloudMachineServerProvisionedV1Beta2Condition,
-			Status:  metav1.ConditionFalse,
-			Reason:  infrav1.HCloudMachineRescueNotEnabledYetV1Beta2Reason,
-			Message: msg,
-		})
-		return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	// The server was created with StartAfterCreate=false and has never been started, so
```
janiskemper: can you add that comment to "powerOn" instead of "serverStub"?

pkg/services/hcloud/server/server.go
```diff
@@ -839,13 +844,14 @@ func (s *Service) handleBootStateEnablingRescue(ctx context.Context, server *hcl
 		Reason:  infrav1.HCloudMachineBootingToRescueV1Beta2Reason,
 		Message: "power on to rescue started",
 	})
-	return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+	// The next state (BootingToRescue) polls purely via SSH, with no hcloud API cost, so
```
janiskemper: we know that powerOn will not go instantly. I would prefer if we put a minimum boundary here. Previously we said 10 seconds. We can also put it to 5 or so. But 1 second just doesn't make sense if in 100% of the cases the server takes longer for the operation.

pkg/services/hcloud/server/server.go
```diff
@@ -1035,13 +1028,14 @@ func (s *Service) handleBootStateBootingToRescue(ctx context.Context, server *hc
 		Message: "custom provisioner running",
 	})
 	s.setBootState(infrav1.HCloudBootStateRunningImageCommand)
-	return reconcile.Result{RequeueAfter: 20 * time.Second}, nil
+	// The next state (RunningImageCommand) polls purely via SSH, with no hcloud API cost, so
+	// requeue immediately instead of waiting.
+	return reconcile.Result{RequeueAfter: requeueImmediately}, nil
```
janiskemper: again: if we have a minimum time for the command to run, then let's not go below that. I believe it should be minimum 10-15 seconds- if not the current 20.

janiskemper (changes requested):

pkg/services/hcloud/server/server.go
guettli: I changed that. Now the function call looks like:

```diff 
	server, res, err, done := s.getLiveServer(ctx)
	if done {
		return res, err
	}
```

When looking at that, I think: What happens, when done is false, and err is non nil?

This can't happen (at the moment). That's why I used a struct in the first version.

@janiskemper wdyt?

pkg/services/hcloud/server/server.go
guettli: done in 61f258968

pkg/services/hcloud/server/server.go
guettli: done in 61f258968

pkg/services/hcloud/server/server.go
guettli: done in 61f258968

pkg/services/hcloud/server/server.go
guettli: done in 61f258968

pkg/services/hcloud/server/server.go
janiskemper: why not 

```
if res != <empty> {
}
if err != nil {
}
``` 

We have that pattern also in other places

guettli: Done in d2f131510.

I could not use the `res != empty` pattern here, because three of the stop-paths deliberately return an empty result *and* a nil error (unauthorized token, rate limit while already `Ready`, and server-was-deleted). For those we do not want a requeue, so an empty `res` plus nil `err` does not mean "keep going".

But `done` turned out to be redundant anyway: it was true in exactly the cases where `server` is nil. So I dropped it and the call site is now:

```go
server, res, err := s.getLiveServer(ctx)
if server == nil {
    return res, err
}
```

guettli: Update, this gets used now:

```go 
	server, res, err := s.getLiveServer(ctx)
	if server == nil || err != nil || !res.IsZero() {
		return res, err
	}
```

Otherwise it looked like the err and res from getLiveServer() where never checked. Currently, `|| err != nil || !res.IsZero()` is not really needed. But maybe the function get updated in the future.

pkg/services/hcloud/server/server.go
```diff
@@ -141,105 +141,114 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
 	case infrav1.HCloudBootStateOperatingSystemRunning:
-		return s.handleOperatingSystemRunning(ctx, server)
+		return s.handleOperatingSystemRunning(ctx)
 	default:
-		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", s.scope.HCloudMachine.Status.BootState)
+		return reconcile.Result{}, fmt.Errorf("unknown BootState: %s", bootState)
 	}
 }
 
+// getLiveServer fetches the live hcloud server. It is only called by the two states
```
janiskemper: can you move the utility function below the places where it is used?

pkg/services/hcloud/server/server.go
```diff
@@ -141,105 +141,114 @@ func (s *Service) Reconcile(ctx context.Context) (res reconcile.Result, err erro
...
-					hcloudutil.HandleRateLimitExceededV1Beta1(s.scope.HCloudMachine, err, "findServer")
-					return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
-				}
-				return reconcile.Result{}, nil
-			}
-
-			return reconcile.Result{}, fmt.Errorf("findServer: %w", err)
-		}
-
-		v1beta1conditions.MarkTrue(s.scope.HCloudMachine, infrav1.HCloudTokenAvailableCondition)
```
janiskemper, line 180: you have to add this also to the other states. Since you removed it from here, it has to exist in all states on the first API call

pkg/services/hcloud/server/server.go
```diff
@@ -598,7 +605,11 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
 		Type:    hcloud.ServerRescueTypeLinux64,
 		SSHKeys: hcloudSSHKeys,
 	}
-	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, server, rescueOpts)
+	serverID, err := s.scope.ServerIDFromProviderID()
```
janiskemper, line 518: can you leave empty lines before and after this statement?

pkg/services/hcloud/server/server.go
guettli: done in 639f2905f

pkg/services/hcloud/server/server.go
guettli: done in 639f2905f

pkg/services/hcloud/server/server.go
guettli: done in 639f2905f

Dhairya-Arora01 (dismissed):

pkg/services/hcloud/server/server.go
```diff
@@ -598,8 +514,17 @@ func (s *Service) handleBootStateInitializing(ctx context.Context, server *hclou
...
-	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, server, rescueOpts)
+
+	serverID, err := s.scope.ServerIDFromProviderID()
+	if err != nil {
+		return res, fmt.Errorf("ServerIDFromProviderID failed: %w", err)
+	}
+
+	result, err := s.scope.HCloudClient.EnableRescueSystem(ctx, &hcloud.Server{ID: serverID}, rescueOpts)
 	if err != nil {
+		if handleUnauthorized(hm, err) {
```
abdullah599, line 525: The client wraps unauthorized with ErrUnauthorized in GetServer, GetAction, ListServers and 7 others, but PowerOnServer (client.go:285) and EnableRescueSystem (client.go:337) return the raw error, so errors.Is(err, ErrUnauthorized) is always false at this call and at line 715.

pkg/services/hcloud/server/server.go
```diff
@@ -1402,6 +1308,97 @@ func (s *Service) handleOperatingSystemRunning(ctx context.Context, server *hclo
...
+			if !s.scope.HCloudMachine.Status.Ready {
+				hcloudutil.HandleRateLimitExceededV1Beta1(s.scope.HCloudMachine, err, "findServer")
+				return nil, reconcile.Result{RequeueAfter: 30 * time.Second}, nil
+			}
+			return nil, reconcile.Result{}, nil
+		}
+
+		return nil, reconcile.Result{}, fmt.Errorf("findServer: %w", err)
+	}
+	markHCloudTokenAvailable(s.scope.HCloudMachine)
```
abdullah599, line 1338: findServer already marks HCloudTokenAvailable true on a successful GetServer (server.go:2215), so this sets the same True again on every BootingToRealOS and OperatingSystemRunning reconcile. Since you added markHCloudTokenAvailable and handleUnauthorized to remove this duplication, can we route findServer and Delete through them too and drop the inline blocks?

pkg/services/hcloud/server/server.go
```diff
@@ -1402,6 +1308,97 @@ func (s *Service) handleOperatingSystemRunning(ctx context.Context, server *hclo
 	return reconcile.Result{}, nil
 }
 
+// getLiveServer fetches the live hcloud server. It is only called by the two states
+// (BootingToRealOS, OperatingSystemRunning) that need a server.Status transition or the
+// network/LB attachment reconcile. The remaining states (Unset, Initializing, EnablingRescue,
```
abdullah599: The comment groups Unset with the GetAction and SSH states, but handleBootStateUnset does neither, it creates the server. Can we drop Unset here or give it its own clause, it does not call getLiveServer because the server does not exist yet

abdullah599:
also, RescueNotEnabledYet and WaitForRescueEnabledToBeFalse reason constants (api/v1beta1/conditions_const.go:412,428 and api/v1beta2/conditions_const.go:459,475) were only set by the two RescueEnabled blocks this PR removes. Grep shows no producers left, so can we delete all four here

pkg/services/hcloud/server/server.go
```diff
@@ -2193,35 +2178,13 @@ func (s *Service) findServer(ctx context.Context) (*hcloud.Server, error) {
 	serverID, err := s.scope.ServerIDFromProviderID()
 	if err == nil {
 		server, err = s.scope.HCloudClient.GetServer(ctx, serverID)
+		// findServer stays a plain fetch. Its callers (getLiveServer, Delete) set the
```
janiskemper: this is not for the reader of the code, but of the PR. What should we do with this comment?

abdullah599: removed it
findServer already has a doc comment

pkg/services/hcloud/server/server_test.go
```diff
@@ -1526,8 +1674,12 @@ var _ = Describe("Reconcile", func() {
 		service.scope.HCloudMachine.Status.BootState = infrav1.HCloudBootStateRunningImageCommand
 		service.scope.HCloudMachine.Status.BootStateSince = metav1.Now()
 
-		By("mocking GetServer to return server 42 (which has IP 1.2.3.4)")
-		hcloudClient.On("GetServer", mock.Anything, mock.Anything).Return(newTestServer(), nil).Once()
+		// RunningImageCommand no longer fetches the live server (see issue #2163), so
```
janiskemper: can we remove all of the mentions of the issue? Let's write it for the reader of the code, not the PR reviewer

janiskemper (approved):

PR #2173, opened 2026-07-15
✨ Add status.controlPlaneLoadBalancer.proxyProtocolEnabled

guettli:
## Summary

- Adds `status.controlPlaneLoadBalancer.proxyProtocolEnabled` (bool) to `HetznerCluster`
- Reflects whether the kube-apiserver load balancer service currently has proxy protocol enabled, as observed on the actual HCloud load balancer service
- Follow-up to #2113 / its release-1.1 backport (#2132), which added `spec.controlPlaneLoadBalancer.enableProxyProtocol`. Since enabling proxy protocol on existing clusters is a gated migration (waiting for all control-plane nodes to carry an annotation before the LB service is recreated), the spec field alone doesn't tell you whether it has actually taken effect yet — this status field does.
- Updates the new status field in the same reconcile that actually enables proxy protocol on the LB, instead of only on the next one. Without this, `status.controlPlaneLoadBalancer` is a snapshot taken from the LB state fetched at the *start* of `Reconcile`, before `reconcileServices` (re)creates the kube-API service with proxy protocol on. Since a status-only change is filtered out by `IgnoreInsignificantHetznerClusterStatusUpdates` (it doesn't requeue the object), the field could stay stale until some unrelated event happened to trigger another reconcile, or until the next periodic full resync (`--sync-period`, e.g. up to 30m on some deployments) — even though proxy protocol was already live on HCloud.

### DIFF
--- a/api/v1beta1/types.go
+++ b/api/v1beta1/types.go
@@ -257,6 +257,13 @@ type LoadBalancerStatus struct {
 	InternalIP string               `json:"internalIP,omitempty"`
 	Target     []LoadBalancerTarget `json:"targets,omitempty"`
 	Protected  bool                 `json:"protected,omitempty"`
+
+	// ProxyProtocolEnabled reflects whether the kube-apiserver load balancer service currently
+	// has proxy protocol enabled, as observed on the actual HCloud load balancer. This can lag
+	// behind spec.controlPlaneLoadBalancer.enableProxyProtocol while the migration to proxy
+	// protocol is in progress (see the field's docs for details).
+	// +optional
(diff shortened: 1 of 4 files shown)


janiskemper (commented):
the status is set before the services reconcile. Meaning that whenever you reconcile the services, it will still change. I believe we should see that we update the status based on the changes that we do in the reconcile loop.

guettli:
> the status is set before the services reconcile. Meaning that whenever you reconcile the services, it will still change. I believe we should see that we update the status based on the changes that we do in the reconcile loop.

Does this solve your concern: 73962913f0f7605ece195de6d400b1ce4d48a211 ?

janiskemper (approved):

PR #2176, opened 2026-07-15
🐛 Deactivate LVM before stopping mdraid in wipe-disk.sh

### LINKED ISSUE #2145: WipeDisk script failure
/kind bug

I have the following disk layout for a disk in a `Hetzner Bare Metal` server :

```tree
|
|-- LVG
|        |
|        |-- Boot partition
|        |
|        |-- .... partition
|        |
|        |-- OS partition
|
....
```

The [wipedisk script](https://github.com/syself/cluster-api-provider-hetzner/blob/main/pkg/services/baremetal/client/ssh/wipe-disk.sh) fails, when running in the context of that `Hetzner Bare Metal` server.

The script fails, trying to stop the `RAID` devices.

I expected the wipedisk script to run properly :).

After investigating with @ashish1099, we found out that :

> We need to first disable the `LVM (Logical Volume Manager)` using this command : `vgchange -an`. And then try stopping the `RAID devices`.

I'm attaching a snippet of an explanation by `Claude` :

```
While any LV in the VG is active, the dm devices are holders of the PV — they hold it open, which is exactly what makes mdadm --stop or wipefs on the layer below fail with EBUSY. Deactivating the VG removes those dm mappings, so the PV (and the md device beneath it, and the partition beneath that) is no longer held open. That's the entire reason it appears in a teardown sequence: it's the step that unbusies the lower layers.
```

**Environment:**

- cluster-api-provider-hetzner version: 
- Kubernetes version: (use `kubectl version`)
- OS (e.g. from `/etc/os-release`):

guettli:
## Summary

`wipe-disk.sh` failed with "device or resource busy" on machines where an LVM volume group sits on top of an mdraid array (for example root and boot carved out of a volume group built on RAID). As long as a logical volume is active, it holds the layer below it open, which is what makes `mdadm --stop` fail.

Before stopping the mdraid arrays of a disk, the script now deactivates the volume groups that live on that disk:

- `lsblk` reports every device stacked on the disk, and `lvs` reports which volume group each active logical volume belongs to. A logical volume is matched by its `/dev/mapper` path, which both tools print the same way.
- Because `lsblk` walks the whole stack, a volume group is found no matter what sits between the disk and the logical volume: partitions, an mdraid array, LUKS, or anything else. Nothing is skipped by device type.
- Only active logical volumes hold the layers below them open, and only those show up in `lsblk`, so exactly the volume groups that keep the disk busy get deactivated. Volume groups on other disks are left alone.
- If `lvs` is not installed, there is no LVM on the machine and the step is skipped. Errors from `lvs` stop the script instead of turning into a confusing "device or resource busy" later on.

Fixes #2145


ashish1099:
Hi @guettli  can we push this :)

guettli:
@ashish1099 I  updated the script. Can you test it, if it works for your environment?

PR #2179, opened 2026-07-15
🐛 Determine proxy-protocol readiness from control-plane machines, not the workload cluster

batistein:
## Summary

The kube-apiserver load balancer can run PROXY protocol so the control planes receive the real client address. This PR decides when it is safe to turn PROXY protocol on, and turns it on at load-balancer creation for clusters that request it.

## Why

Turning PROXY protocol on for a load balancer whose backends are not prepared for it sends malformed PROXY headers to those backends and breaks them. The readiness signal has to be reliable, and it must not depend on a live connection to the workload API server, whose own path runs through the load balancer being changed. The provider already has the control-plane machines in the management cluster, so it can decide from state it owns.

## What it does

- `AllControlPlaneMachinesReadyForProxyProtocol` lists the cluster's control-plane `Machine`s (by the `cluster.x-k8s.io/control-plane` label) and returns true only when there is at least one and every one carries the annotation `capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true"`.

  Reading Cluster API `Machine`s covers hcloud, bare-metal, and mixed control planes the same way, and the provider already has list permission on them. While the check is false, the load-balancer reconcile requeues every 10 seconds and re-evaluates.

- The annotation is set on the control-plane `machineTemplate.metadata`, so a machine created from an older template does not carry it. The check stays false until the last such machine is replaced, so PROXY protocol is never turned on while a plain-TCP backend is still in the fleet. A machine only ever becomes an LB target once its apiserver pod is confirmed healthy (a separate, existing gate), so the annotation check alone is enough to guarantee every current target can handle PROXY protocol — no extra machine-health check is needed here.

- `createOptsFromSpec` creates the kube-apiserver service with PROXY protocol taken from `enableProxyProtocol`, but only when the cluster has never had a load balancer before (`status.controlPlaneLoadBalancer` is still unset). A cluster requesting PROXY protocol serves it from creation; if an auto-created load balancer is instead being recreated for a cluster that already had one — e.g. it was deleted out-of-band — the annotation-gated migration path in `reconcileServices` still applies, since that cluster's control-plane machines may not be ready for it yet. A fixed-name/BYO load balancer never goes through `createOptsFromSpec` at all (it's adopted via `ownExistingLoadBalancer`); the first time `reconcileServices` creates a kube-apiserver service on it, that already happens straight from spec too, the same as a genuinely new cluster.

- The control-plane endpoint is set right after the load-balancer reconcile, so control-plane machines are created even while the proxy-protocol check requeues. This runs regardless of whether the load-balancer reconcile itself requeued, errored, or hasn't set a status yet.

## Behavior

- A new cluster with `enableProxyProtocol: true` serves PROXY protocol on the kube-apiserver service from creation.
- Turning PROXY protocol on for a load balancer that already serves without it, or recreating an auto-created load balancer for a cluster that already had one, waits for every control-plane machine to carry the annotation.
- Enabling PROXY protocol is a one-way operation.

## Follow-up

The migration path still deletes and recreates the kube-apiserver LB service to flip proxy protocol on, which briefly removes the service from the LB. Switching that to an in-place update is split out into #2185 (issue #2184), based on this branch.

## Tests

Unit tests cover the annotation-based readiness check (all annotated, missing annotation, no machines), the create-time PROXY protocol value (including the recreate-for-an-existing-cluster case), and that setting the control-plane endpoint doesn't panic when the load-balancer status isn't known yet.


pkg/scope/cluster.go
```diff
@@ -271,48 +270,67 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
 
-	restConfig, err := wlClientConfig.ClientConfig()
-	if err != nil {
-		return false, fmt.Errorf("failed to get workload cluster rest config: %w", err)
-	}
-	wlClient, err := client.New(restConfig, client.Options{})
-	if err != nil {
-		return false, fmt.Errorf("failed to create workload cluster client: %w", err)
+	hcloudMachines := &infrav1.HCloudMachineList{}
+	if err := s.Client.List(ctx, hcloudMachines, listOptions...); err != nil {
```
janiskemper: why don't we list the Machines?

janiskemper (changes requested):

guettli (approved):
lgtm

guettli (approved):
lgtm

abdullah599 (commented):
LGTM

controllers/hetznercluster_controller.go
```diff
@@ -231,6 +231,15 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 
 	// reconcile the load balancers
 	res, err := loadbalancer.NewService(clusterScope).Reconcile(ctx)
+
+	// Set the control-plane endpoint from the load-balancer IP as soon as it is known,
+	// before honoring any requeue from the load-balancer reconcile. The proxy-protocol
+	// migration requeues until every control-plane machine is ready, but those machines
```
janiskemper: we removed this "readiness check". This whole code block / change is not valid anymore AFAIK

pkg/scope/cluster.go
```diff
@@ -271,48 +270,51 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
-	if err != nil {
-		return false, fmt.Errorf("failed to get workload cluster rest config: %w", err)
-	}
-	wlClient, err := client.New(restConfig, client.Options{})
-	if err != nil {
-		return false, fmt.Errorf("failed to create workload cluster client: %w", err)
+// AllControlPlaneMachinesReadyForProxyProtocol returns true when the control plane is
+// fully rolled out to the machine template that expects PROXY protocol at the load
+// balancer. It lists the cluster's control-plane Machines in the management cluster (one
+// list, so it covers both HCloud and bare-metal control planes) and requires every one to
```
janiskemper: we list machines not hcloud and baremetal machines

pkg/scope/cluster.go
```diff
@@ -271,48 +270,51 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
+// carry the annotation capi.syself.com/proxy-protocol-for-controlplane-loadbalancer:
+// "true", which the control-plane machine template stamps and Cluster API propagates onto
```
janiskemper: let's replace "stamp"

pkg/scope/cluster.go
```diff
@@ -271,48 +270,51 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
 			return false, nil
 		}
 	}
+
 	return true, nil
 }
+
+// controlPlaneMachineReadyForProxyProtocol reports whether one control-plane Machine is on
+// the template that expects PROXY protocol, meaning it carries the annotation.
+func (s *ClusterScope) controlPlaneMachineReadyForProxyProtocol(machine *clusterv1.Machine) bool {
```
janiskemper: let's inline this function.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -321,36 +321,37 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 
 	// kubeAPIServiceExists: whether the kube-API service already exists on the LB.
 	// New cluster: service absent → create immediately with EnableProxyProtocol from spec (no annotation check).
-	// Existing cluster migration: service present without proxy protocol → wait for all CP nodes to carry the
-	// annotation before recreating, to avoid sending malformed PROXY-protocol headers to unprepared backends.
+	// Existing cluster migration: service present without proxy protocol → wait for all control-plane
```
janiskemper: let's maybe improve this comment (even before it was not great)

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -449,7 +450,25 @@ func createOptsFromSpec(hc *infrav1.HetznerCluster) (hcloud.LoadBalancerCreateOp
 	// Set name
 	name := utils.GenerateName(nil, fmt.Sprintf("%s-kube-apiserver-", hc.Name))
 
-	proxyprotocol := false
+	// A freshly created load balancer takes proxy protocol straight from the spec, but only when
```
janiskemper: given the other PR based on this branch, I'm not sure if this makes sense. There, we set proxyprotocol in a native way whenever the respective service doesn't exist yet.

janiskemper (changes requested):
I have some points we can change. We should, after we did that change, also get the other PR merged into this one and look at it holistically.

controllers/hetznercluster_controller.go
abdullah599: i may be missing context here, so correct me
after the #2185 merge the readiness check and the 10s requeue still look present in reconcileServices. if we take it out, the requeue returns before the endpoint host is set, and KubeadmControlPlane needs that endpoint to create the control plane machines that carry the annotation, so a load balancer that already serves the API port with proxy requested but has no machines yet would loop forever

pkg/services/hcloud/loadbalancer/loadbalancer.go
abdullah599: I kept it because createOptsFromSpec creates the load balancer together with its service, so the service already exists by the time reconcileServices runs. reconcileServices only sets proxy when the service is missing, so here it does nothing. That leaves createOptsFromSpec as the only place to decide proxy at creation.

it would be cleaner to create the load balancer empty and let reconcileServices add the service 🤔  Then proxy lives in one place. is that what you had in mind?

controllers/hetznercluster_controller.go
janiskemper: We previously checked that the machines are "ready". Now we only check that they have the annotation. Do you still see the "readiness" check? Because we shouldn't have it and only rely on the annotation

controllers/hetznercluster_controller.go
```diff
@@ -231,6 +231,15 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+	// are only created once KubeadmControlPlane sees the endpoint. Setting it here first
+	// breaks that deadlock on a fresh cluster; the deferred scope Close persists it even
+	// on the requeue return below.
+	processControlPlaneEndpoint(hetznerCluster)
```
janiskemper: do we really need to move this function? I am not sure if I pointed it out before, I think I did, but I don't remember the outcome

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -319,46 +319,32 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 	toCreate, toDelete := utils.DifferenceOfIntSlices(wantServiceListenPorts, slices.Collect(maps.Keys(existingServicesByPort)))
 
-	// kubeAPIServiceExists: whether the kube-API service already exists on the LB.
-	// New cluster: service absent → create immediately with EnableProxyProtocol from spec (no annotation check).
-	// Existing cluster migration: service present without proxy protocol → wait for all CP nodes to carry the
-	// annotation before recreating, to avoid sending malformed PROXY-protocol headers to unprepared backends.
+	// Two cases for the kube-API service:
+	//   - present without proxy protocol → an existing cluster enabling it: wait until every
+	//     control-plane machine is annotated, then switch it on in place below.
+	//   - absent → create it below from the spec value (a new cluster does not reach this;
```
janiskemper: improving the comment in the (): 

it's absent only in case an existing load balancer is taken over instead of creating a new one, or if the service got manually deleted

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -449,7 +455,10 @@ func createOptsFromSpec(hc *infrav1.HetznerCluster) (hcloud.LoadBalancerCreateOp
 	// Set name
 	name := utils.GenerateName(nil, fmt.Sprintf("%s-kube-apiserver-", hc.Name))
 
-	proxyprotocol := false
+	// Take proxy protocol from the spec, but only for a new cluster (Status still nil). If this
```
janiskemper: the load balancer cannot be recreated for a cluster. Then the cluster is gone. I don't understand this change.

janiskemper (changes requested):

janiskemper (approved):

PR #2185, opened 2026-07-17
Switch proxy-protocol LB migration from delete+recreate to in-place update

### LINKED ISSUE #2184: Switch proxy-protocol LB migration from delete+recreate to in-place update
## What

When enabling PROXY protocol on an existing load balancer (the migration path, once every control-plane machine carries the readiness annotation), `reconcileServices` currently deletes the kube-apiserver LB service and recreates it with `Proxyprotocol` set:

https://github.com/syself/cluster-api-provider-hetzner/blob/batistein/proxy-protocol-cp-machine-check/pkg/services/hcloud/loadbalancer/loadbalancer.go

Delete+recreate has a window where the kube-apiserver service doesn't exist on the load balancer at all, so every client hitting it (kubelets, kubectl, etc.) is refused for that window — not just backends without PROXY protocol support.

## Proposal

HCloud's `update_service` action (`LoadBalancerClient.UpdateService` in hcloud-go) can flip `Proxyprotocol` on a live service directly. Use it instead of delete+recreate for this migration path, so the kube-apiserver service is never briefly absent from the load balancer.

This requires:
- Adding `UpdateServiceOnLoadBalancer` to the `hcloudclient.Client` interface, the real client, the fake client, and the generated mocks.
- Replacing the delete+recreate pair in `reconcileServices` with a single `UpdateServiceOnLoadBalancer` call for the kube-apiserver port.

guettli:
## Summary

Switches the proxy-protocol migration path (enabling PROXY protocol on an existing kube-apiserver load-balancer service) from delete+recreate to an in-place update.

Closes #2184

## Why

`reconcileServices` used to delete the kube-apiserver LB service and recreate it with `Proxyprotocol` set once every control-plane machine carried the readiness annotation. Delete+recreate has a window where the service doesn't exist on the load balancer at all, so every client hitting it — not just backends without PROXY protocol support — would be refused for that window.

HCloud's `update_service` action can flip `Proxyprotocol` on a live service directly, so the service is never briefly absent.

## What it does

- Adds `UpdateServiceOnLoadBalancer` to the `hcloudclient.Client` interface, the real client (backed by `LoadBalancer.UpdateService`), the fake client, and the generated mocks.
- Replaces the delete+recreate pair for the kube-apiserver port with a single `UpdateServiceOnLoadBalancer(Proxyprotocol: true)` call.
- Fixes the fake client's `AddServiceToLoadBalancer`, which silently dropped the `Proxyprotocol` field, so tests can exercise the flip end-to-end.

## Base branch

This is based on `batistein/proxy-protocol-cp-machine-check` (#2179), which it depends on.

## Tests

Added coverage for the in-place switch: updates without deleting the service once all control-plane machines are annotated, and requeues with the service unchanged otherwise.


pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -396,13 +382,32 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 			// Status.ControlPlaneLoadBalancer was snapshotted from the LB state fetched at the
-			// start of Reconcile, before this service was (re)created, so it still shows the old
+			// start of Reconcile, before this service was created, so it still shows the old
 			// value. Update it now so callers observe the change in this reconcile instead of
 			// waiting for the next one (e.g. the next full resync, up to --sync-period later).
 			s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ProxyProtocolEnabled = proxyProtocol
 		}
 	}
 
+	// Enabling proxy protocol on an existing kube-API service is a one-way switch, applied
```
abdullah599: we can drop the rather than delete+recreate and just say what the code does now

abdullah599 (commented):
LGTM

abdullah599 (approved):

pkg/services/hcloud/client/client.go
```diff
@@ -218,6 +219,11 @@ func (c *realClient) AddServiceToLoadBalancer(ctx context.Context, lb *hcloud.Lo
 	return err
 }
 
+func (c *realClient) UpdateServiceOnLoadBalancer(ctx context.Context, lb *hcloud.LoadBalancer, listenPort int, opts hcloud.LoadBalancerUpdateServiceOpts) error {
+	_, _, err := c.client.LoadBalancer.UpdateService(ctx, lb, listenPort, opts)
```
janiskemper, line 223: we probably should add the unauthorized handling everywhere right? this is maybe not related to this PR but will prevent future mistakes. If you agree, you can write an issue

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -377,8 +362,9 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 	for i, listenPort := range toCreate {
 		proxyProtocol := false
 		if listenPort == kubeAPIServicePort {
-			// Proxy protocol is only relevant for the kube-apiserver port (default 6443).
-			proxyProtocol = kubeAPIProxyProtocol
+			// The kube-API service only reaches this loop when it doesn't exist on the LB yet
```
janiskemper: this comment is not written for people reading the code - but for those reading the PR. IMO We should add a comment about that proxyprotocol is only relevant for the kubeapi service.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -396,13 +382,31 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 			// Status.ControlPlaneLoadBalancer was snapshotted from the LB state fetched at the
-			// start of Reconcile, before this service was (re)created, so it still shows the old
+			// start of Reconcile, before this service was created, so it still shows the old
 			// value. Update it now so callers observe the change in this reconcile instead of
 			// waiting for the next one (e.g. the next full resync, up to --sync-period later).
 			s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ProxyProtocolEnabled = proxyProtocol
 		}
 	}
 
+	// Enabling proxy protocol on an existing kube-API service is applied in place via
```
janiskemper: also here: let's write it in a straightforward way from code perspective: If proxyprotocol is not yet active, but should be, then we'll activate it.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -396,13 +382,31 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+		updateOpts := hcloud.LoadBalancerUpdateServiceOpts{Proxyprotocol: &proxyProtocol}
+		if err := s.scope.HCloudClient.UpdateServiceOnLoadBalancer(ctx, lb, kubeAPIServicePort, updateOpts); err != nil {
+			// return immediately on rate limit
+			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "UpdateServiceOnLoadBalancer")
+			multierr = errors.Join(multierr, fmt.Errorf("failed to update kube-API service on load balancer to enable proxy protocol: %w", err))
+			if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
+				return reconcile.Result{}, multierr
+			}
+		} else {
+			s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ProxyProtocolEnabled = true
```
janiskemper, line 411: let's do this outside the if statement, so that it will be always re-computed in case the status gets lost.

abdullah599: Moving it out would break the new cluster case. There the service does not exist yet, so proxyProtocolAlreadyActive is false, but we create the service with proxy protocol on and set the status to true. An unconditional write at the end would then flip it back to false.

On the status getting lost, that part is already handled. statusFromHCloudLB reads the proxy protocol value from the live load balancer, and Reconcile runs it on every pass before reconcileServices, so a lost status comes back on the next reconcile on its own. This inline write is only there to reflect the flip in the same reconcile instead of waiting.

pkg/services/hcloud/client/client.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2186

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -396,13 +383,31 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 			// Status.ControlPlaneLoadBalancer was snapshotted from the LB state fetched at the
-			// start of Reconcile, before this service was (re)created, so it still shows the old
+			// start of Reconcile, before this service was created, so it still shows the old
 			// value. Update it now so callers observe the change in this reconcile instead of
 			// waiting for the next one (e.g. the next full resync, up to --sync-period later).
 			s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ProxyProtocolEnabled = proxyProtocol
 		}
 	}
 
+	// If proxy protocol is not active yet but should be, activate it in place. HCloud's
```
janiskemper, line 397: this whole code block should be ideally after requeueForProxyProtocol or not? 

So first, if things are not done yet, we requeue, and then, if they are done, we update it. Seems more logical to me

janiskemper (approved):

PR #2190, opened 2026-07-23
🐛 Gate control-plane load balancer PROXY protocol on the InfraMachine annotation

batistein:
## What

The check that decides when to enable PROXY protocol on the control-plane load balancer now reads the annotation `capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true"` from the control-plane infrastructure machines (`HCloudMachine` and `HetznerBareMetalMachine`) instead of from the control-plane `Machine` objects. The annotation is expected on the infrastructure machine template's `spec.template.metadata`, not on the `KubeadmControlPlane` `machineTemplate`.

## Why

PROXY protocol may only be enabled once every control-plane backend has been rolled to a template that accepts a PROXY header. The readiness annotation therefore has to appear only on machines that were actually created from the new template.

An annotation on the `KubeadmControlPlane` `machineTemplate.metadata` cannot express that. Cluster API treats `machineTemplate` labels and annotations as in-place mutable: when the template reference changes (for example through a ClusterClass rollout), Cluster API copies the new metadata onto the existing, not-yet-replaced `Machine`s right away, without a rollout. The annotation then shows up on the old machines immediately and the check passes before any machine has been replaced, which would enable PROXY protocol against backends that do not expect it.

Infrastructure machine templates do not behave this way. Cluster API copies infrastructure-template metadata onto an `InfraMachine` only when it clones a new one, so the annotation lands only on machines created from the new template. Existing infra machines never gain it, and the check stays false until the last old machine is replaced.

Reading the infrastructure machines also keeps the check entirely in the management cluster; no workload-cluster client is needed.

## Changes

- `pkg/scope`: `AllControlPlaneMachinesAnnotatedForProxyProtocol` (which listed `Machine`s) becomes `AllControlPlaneInfraMachinesAnnotatedForProxyProtocol`, listing the control-plane `HCloudMachine`s and `HetznerBareMetalMachine`s and requiring the annotation on all of them.
- `pkg/services/hcloud/loadbalancer`: call the new function.
- `api`: add a `metadata` field to `HetznerBareMetalMachineTemplateResource` so the bare-metal template can carry annotations under `spec.template.metadata`. `HCloudMachineTemplate` already has this; the bare-metal template now matches it and the Cluster API machine-template contract. CRD regenerated.
- Unit tests updated to cover HCloud and bare-metal control-plane machines.

## Action for template authors

Set `capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true"` on the control-plane infrastructure machine template's `spec.template.metadata.annotations`, not on the `KubeadmControlPlane` `machineTemplate.metadata.annotations`.

### DIFF
--- a/pkg/scope/cluster.go
+++ b/pkg/scope/cluster.go
@@ -270,19 +270,13 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
 	return err
 }
 
-// AllControlPlaneMachinesAnnotatedForProxyProtocol returns true when the control plane is
-// fully rolled out to the machine template that expects PROXY protocol at the load
-// balancer. It lists the cluster's control-plane Machines in the management cluster (the
-// Cluster API Machine, so one list covers every control plane whatever its infrastructure)
-// and requires every one to carry the annotation
-// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true", which the
-// control-plane machine template sets and Cluster API propagates onto the Machine.
-//
-// Machines from an earlier template do not carry the annotation, so the check stays false
-// until the last of them is replaced. It returns false (no error) while the cluster has no
-// control-plane machines yet.
-func (s *ClusterScope) AllControlPlaneMachinesAnnotatedForProxyProtocol(ctx context.Context) (bool, error) {
-	machines := &clusterv1.MachineList{}
+// AllControlPlaneInfraMachinesAnnotatedForProxyProtocol returns true when every control-plane
+// infrastructure machine (HCloudMachine and HetznerBareMetalMachine) carries the annotation
+// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true", which is set on the
+// control-plane infrastructure machine template's spec.template.metadata. It lists the machines in
+// the management cluster, so it needs no workload-cluster client, and returns false (no error) while
+// the cluster has no control-plane infrastructure machines yet.
+func (s *ClusterScope) AllControlPlaneInfraMachinesAnnotatedForProxyProtocol(ctx context.Context) (bool, error) {
 	listOptions := []client.ListOption{
 		client.InNamespace(s.Namespace()),
 		client.MatchingLabels{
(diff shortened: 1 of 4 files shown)


abdullah599 (approved):
LGTM

pkg/scope/cluster.go
```diff
@@ -270,41 +270,52 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
-//
-// Machines from an earlier template do not carry the annotation, so the check stays false
-// until the last of them is replaced. It returns false (no error) while the cluster has no
-// control-plane machines yet.
-func (s *ClusterScope) AllControlPlaneMachinesAnnotatedForProxyProtocol(ctx context.Context) (bool, error) {
-	machines := &clusterv1.MachineList{}
+// AllControlPlaneInfraMachinesAnnotatedForProxyProtocol returns true when every control-plane
+// infrastructure machine (HCloudMachine and HetznerBareMetalMachine) carries the annotation
+// capi.syself.com/proxy-protocol-for-controlplane-loadbalancer: "true", which is set on the
+// control-plane infrastructure machine template's spec.template.metadata. It lists the machines in
```
janiskemper: let's remove the last sentence and see if we wanna bring something back from the original comment

pkg/scope/cluster.go
```diff
@@ -270,41 +270,52 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
-	if len(machines.Items) == 0 {
-		s.V(1).Info("proxy protocol: no control-plane machines found yet")
-		return false, nil
+	found := 0
+
+	hcloudMachines := &infrav1.HCloudMachineList{}
+	if err := s.Client.List(ctx, hcloudMachines, listOptions...); err != nil {
+		return false, fmt.Errorf("failed to list control-plane HCloudMachines: %w", err)
+	}
+	for i := range hcloudMachines.Items {
```
janiskemper, line 297: let's leave an empty line here

janiskemper (approved):

PR #2193, opened 2026-07-23
✨ feat: api-lb http-healthcheck

### LINKED ISSUE #1339: support http healthcheck for api lb
/kind feature

I would like to use a http healthcheck for the api lb. This prevents the healthcheck to be ready before the apiserver itself is ready, because the port is opened before the apiserver is ready (and tcp healthcheck will be ready before apiserver is ready).

Maybe this could be added like this:

```yaml
spec:
  controlPlaneLoadBalancer:
    healthcheck:
      type: http # (or tcp / https (http with tls enabled))
      interval: 15s
      timeout: 10s
      retries: 3
      domain: ""
      path: "/readyz"
```

if omitted use the hetzner default values (tcp, 15s interval, 10s timeout, 3 retries)

simonostendorf:

By default, the Hetzner Cloud Load Balancer for the kube-apiserver service uses a plain TCP health check, which only verifies that port 6443 accepts connections — not that the apiserver is actually ready to serve requests. This can route traffic to a control-plane node whose apiserver is reachable via TCP but not healthy (e.g. still starting up, or degraded). This is expecially useful when performing rotating upgrades of controlplane nodes.

This PR adds an optional controlPlaneLoadBalancer.healthCheck field to HetznerCluster (in both v1beta1 and v1beta2) that lets users switch the load balancer to an HTTP(S) health check against a path such as /readyz, with configurable interval, timeout, and retries. The feature is strictly opt-in: if healthCheck is omitted, Hetzner's existing default (TCP, 15s/10s/3) is unchanged and CAPH never touches the load balancer's health check configuration.

Fixes #1339

The healthCheck field intentionally lives only on controlPlaneLoadBalancer, not on extraServices — this matches the scope discussed in the issue.
HTTP(S) health checks are unauthenticated by nature (the load balancer can't present client credentials), so enabling this is a deliberate security tradeoff for the cluster operator to make — documented in docs/caph/03-reference/02-hetzner-cluster.md, not enforced or defaulted by CAPH.
Not an image version change.

**TODOs**:

- [x] squash commits
- [x] include documentation
- [x] add unit tests


janiskemper (commented):
thanks @simonostendorf ! Your idea was apparently very popular after it settled a bit (🙄) and @batistein created also a PR - directly on the release branch, to even include it in the upcoming release. 

I see that both PRs are mostly aligned, which is good, because it makes them comparable. I saw already some things that you included and that are not there in the other PR and commented on it. 

What the other PR has that yours hasn't: A safe upgrade procedure from clusters that don't support this health check yet. This is something we already had for proxy protocol (in the new release) and could re-use. 

My suggestion: You review the other PR as well, we get aligned on that, and then you can update this PR accordingly. Sounds good?

simonostendorf:
Hey @janiskemper thanks for the info about the other PR.

I went through #2204 and compared both MRs.

I've updated this PR with the missing safe migration procedure.

I also renamed the spec fields to match the Hetzner Cloud APIs own health_check object naming (protocol, interval, timeout, retries, path, domain, response, statusCodes), so the two PRs are aligned there too.

Happy to adjust further based on how youd like to reconcile the two PRs, but from my side this should now cover everything from both. Would love to get this merged :)

janiskemper:
Thanks @simonostendorf ! There are still some things missing (e.g. the test setup). @abdullah599 made the respective changes in the other PR and I review it now. In case you want to have a look, you can also review the other PR. 

In any case, I would propose that we get the other one ready first, so that you don't have to go through the review comments yourself. 

After the other PR is ready, I would propose that you take over all the code and adapt it to the main branch. There are some differences (v1beta2, webhook folder, etc.) that you'll have to look at. 

In case you want someone else to do it, it's also fine for me - just say it.

Thanks again for your efforts and for pushing this!

simonostendorf:
@janiskemper Implemented changes from other pr and rebased :)

janiskemper:
great! We'll take a look. However, it's already in the release branch, so it will be in the upcoming release in the next days anyway! :rocket:

simonostendorf:
I will do the next rebase after a first review

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -380,8 +383,49 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+				hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "UpdateServiceOnLoadBalancer")
+				multierr = errors.Join(multierr, fmt.Errorf("failed to update kube-API service on load balancer to apply health check: %w", err))
+				if hcloud.IsError(err, hcloud.ErrorCodeRateLimitExceeded) {
+					return reconcile.Result{}, multierr
+				}
+			}
+		}
+	}
+
+	if requeueForProxyProtocol || requeueForHTTPHealthCheck {
```
abdullah599: this returns before the proxy protocol block below, so a health check that is still waiting also blocks enabling proxy protocol. release-1.1 keeps the proxy return on its own and returns inside the health check branch, can we do the same here?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -380,8 +383,49 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+		return reconcile.Result{RequeueAfter: 10 * time.Second}, multierr
```
abdullah599: this also drops the proxy protocol requeue from 2m to 10s, release-1.1 kept 2m there.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -380,8 +383,49 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+
+	// If the kube-API service already exists and its health check no longer matches the spec,
+	// update it in place.
+	var requeueForHTTPHealthCheck bool
+	kubeAPIDestinationPort := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.Port
+	if wantHealthCheck := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.HealthCheck; wantHealthCheck != nil &&
+		kubeAPIServiceExists && healthCheckNeedsUpdate(wantHealthCheck, existingKubeAPIService.HealthCheck, kubeAPIDestinationPort) {
+		// Switching a live service from a tcp check to an http or https check can mark every
+		// backend unhealthy at once if the backends don't answer the path yet, which would take
+		// the API server offline. Gate that one-way switch on all control-plane infrastructure
```
abdullah599: "gate that one-way switch" was flagged as confusing on the release-1.1 PR. release-1.1 says: "So wait until every control-plane infra machine carries the annotation, the same as the proxy-protocol migration". same phrase in healthCheckMigratesToHTTP below

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -435,6 +479,219 @@ func (s *Service) createLoadBalancer(ctx context.Context) (*hcloud.LoadBalancer,
 	return lb, nil
 }
 
+// healthCheckOptsFields holds the fields shared by the three structurally-identical hcloud
+// health-check options types used for creating, adding and updating a load balancer service.
+// Go doesn't let one type be reused across all three request builders, so this is converted
```
abdullah599: can we move healthCheckOptsFields and healthCheckOptsFieldsFromSpec below the three builders that use them, janis asked for that on the release-1.1 PR. release-1.1 also named these healthCheckOpts and healthCheckDiffers(observed, desired), matching would keep the branches clean

api/v1beta1/types.go
```diff
@@ -279,6 +279,31 @@ type LoadBalancerSpec struct {
...
+	// default behavior (a plain TCP check, with Hetzner's default interval, timeout and retries)
+	// is unchanged.
+	//
+	// Switching Protocol to http or https lets the load balancer check the kube-apiserver's actual
+	// readiness (e.g. path "/readyz") instead of just whether the port accepts connections. Doing
+	// so requires the kube-apiserver to serve that path without authentication, since the load
+	// balancer's health check request is unauthenticated. This is not configured by CAPH; the
+	// cluster operator must allow anonymous access to the configured path on the apiserver.
+	//
+	// On an existing cluster, switching from tcp to an http or https check is a one-way migration
```
abdullah599: it is not one-way, healthCheckMigratesToHTTP compares the live protocol every reconcile so the wait repeats on every tcp to http switch. also the annotation sits on the infra machines, not the CAPI Machine. same text in api/v1beta2/types.go and in the annotation const

api/v1beta1/types.go
```diff
@@ -313,6 +338,67 @@ func (spec LoadBalancerSpec) WantsIPv6() bool {
...
+	// Port is the target port the check runs against. It defaults to the service's destination
+	// port. Set it when the health-check endpoint is served on a different port than the API
+	// server itself.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port *int `json:"port,omitempty"`
+
+	// IntervalSeconds is the time in seconds between two consecutive health checks. If omitted,
+	// Hetzner's default (15s) is used.
```
abdullah599: can we link Hetzner's health_check reference instead of writing 15s here, janis asked for that on the release-1.1 PR since the numbers can change. same for timeoutSeconds, retries, statusCodes, both api versions and the docs table

docs/caph/03-reference/02-hetzner-cluster.md
```diff
@@ -28,6 +28,51 @@ If you are using your own load balancer, you need to point towards it (by settin
...
+This is opt-in and requires the kube-apiserver to serve the configured path without
+authentication, since the load balancer's health check request is unauthenticated. CAPH does not
+configure this for you; you must allow anonymous access to the configured path yourself, e.g. via
+kubeadm's default `system:public-info-viewer` RBAC binding or an
+[AuthenticationConfiguration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests)
+that scopes anonymous access to that path only. If this tradeoff isn't acceptable for your
+environment, leave `healthCheck` unset to keep the plain TCP check.
+
+CAPH only sends the fields you set; a field you never set keeps Hetzner's own default (tcp, 15s
+interval, 10s timeout, 3 retries — see the linked API reference above for the current defaults).
```
abdullah599: there is no linked API reference above, the section only says the field names mirror the health_check object. can we add the link and drop the hardcoded 15s/10s/3

docs/caph/03-reference/02-hetzner-cluster.md
```diff
@@ -28,6 +28,51 @@ If you are using your own load balancer, you need to point towards it (by settin
...
+server itself; left unset, the check runs against the service's own destination port.
+
+Leaving `healthCheck` out entirely means CAPH does not manage the health check at all, so a load
+balancer you point CAPH at (or one that already has a check applied) keeps whatever check it has.
+That also means deleting `healthCheck` after an http check was applied does not undo it — to get a
+tcp check back, keep the field and set `healthCheck.protocol: tcp`.
+
+### Safe migration on an existing cluster
+
+Switching an existing cluster's health check from tcp to http or https is a one-way migration: if
```
abdullah599: "one-way migration" contradicts line 71 which says the gate applies to every switch away from tcp. also "control-plane machine" should be "control-plane infra machine", same two lines below

abdullah599:
Hi @simonostendorf, your latest push looks good, the review comments are addressed.

Context: #2204 built the same feature on `release-1.1` and is already merged. We want both branches to behave the same, so I compared them. The API, CRD schema, defaults, validation and controller behavior all match. Only these small things are left:

| # | What is left | Why it matters | Fix |
|---|--------------|----------------|-----|
| 1 | The CRD files were not regenerated | You fixed the wording in the Go types, but `config/crd/bases/*.yaml` still has the old text from my earlier comments: "one-way migration", "control-plane machine template", and the hardcoded `15s` / `3` / `["2??", "3??"]`. CI is blocked on this too. | `make generate-manifests` and commit. Only the descriptions change, the schema stays the same. |
| 2 | The fake Hetzner client used in tests does not store health checks | It keeps returning the old check, so a test that sets `healthCheck` would never see it applied and the controller would call the update API on every reconcile. Nothing breaks today because your tests use mocks, but it will be a problem for the first envtest that touches this field. | Copy what #2204 added to `pkg/services/hcloud/client/fake/hcloud_client.go`. |
| 3 | `loadbalancer.go:424` returns the error on its own | If adding or deleting a service already failed in the same reconcile, that earlier error is lost. | `errors.Join(multierr, err)`, same as release-1.1. |
| 4 | Two sentences missing from the docs | We do not say that only the kube-apiserver service gets a health check, and we do not warn that CAPH applies the new check without first checking that the servers pass it. Both are true for your PR too. | Copy both from the release-1.1 docs. |
| 5 | No `nil` check in `healthCheckDiffers` | Nothing breaks today because the caller already checks. | Add `if desired == nil { return false }` at `loadbalancer.go:620`, only if you want. |

Two things are missing on the release-1.1 side instead. I will port those back, nothing for you to do:

| What | Detail |
|------|--------|
| How to make the health path answer without credentials | Your PR explains it with `system:public-info-viewer` and AuthenticationConfiguration, release-1.1 does not. |
| Four of your tests | Extra services still being reconciled while CAPH waits for the annotation, proxy protocol still being turned on during that wait, no API call when the live check already matches, and an unset `healthCheck` leaving the current check alone. |

Do you want to finish this yourself, or should I take it over?Just tell me which you prefer. If you keep it, please rebase on main first.

simonostendorf:
Implemented all comments, rebased and squashed, please have a look :)

abdullah599 (approved):
Looks good to me! 🚀

Thanks again @simonostendorf for the feature and for the patience :)

PR #2016, opened 2026-05-13
✨ Add spec.template.metadata to HetznerBareMetalMachineTemplate

ashish1099:
fixes https://github.com/syself/cluster-api-provider-hetzner/issues/1680

sponsored by https://obmondo.com

PR #2201, opened 2026-07-25
🌱 Add spec.template.metadata to HBMM template (port #2191 to main)

abdullah599:
Forward-ports PR #2191 from release-1.1. Adds ObjectMeta to HetznerBareMetalMachineTemplateResource so the bare-metal template can carry labels and annotations, matching HCloudMachineTemplate and the Cluster API machine-template contract.

Unlike release-1.1 (v1beta1 only), main adds the field to both versions: v1beta1 uses clusterv1beta1.ObjectMeta, v1beta2 uses clusterv1.ObjectMeta. Because the metadata type differs between versions, the resource is excluded from conversion-gen and labels/annotations are copied by hand-written converters, following the existing HCloudMachineTemplateResource pattern.

None (forward-port of PR #2191).

Generated files (deepcopy, conversion, CRD) were regenerated with `make generate`. Needed for stacked PR that gates PROXY protocol on the InfraMachine annotation.


janiskemper:
this is a duplicate of #2016. Can you instead review that one?

abdullah599:
The HBMM template metadata part is now on main via PR #2016. The proxy-protocol part moved to a standalone PR: #2205. Closing this one.

PR #2204, opened 2026-07-25
✨ Add a configurable health check for the control-plane load balancer

batistein:
## What

Add an optional health check for the control-plane load balancer, configurable per cluster via `spec.controlPlaneLoadBalancer.healthCheck`.

## Why

The control-plane load balancer uses a plain TCP check on the API server service port. A target counts as healthy as soon as the port accepts a connection, which happens before the API server behind it is ready to serve. During a rollout the load balancer can send API traffic to a control plane whose API server is still starting.

An HTTP or HTTPS check against a readiness path (for example `/readyz`) lets the load balancer route only to a target whose API server actually answers, and drop one that stops answering.

## What changed

- New optional `healthCheck` on the control-plane `LoadBalancerSpec`: `protocol` (tcp, http, https; default tcp), `path`, `port`, `statusCodes`, `intervalSeconds`, `timeoutSeconds`, `retries`.
- Applied when the kube-apiserver service is created, and reconciled in place when the spec changes. Only the kube-apiserver service is affected; extra services keep their behavior.
- **Opt-in and non-breaking**: when `healthCheck` is unset the service keeps the load balancer's default TCP check, so existing clusters are unchanged. `v1beta1` is the storage and hub version, so the new optional field needs no conversion.

## Safe migration from TCP to HTTP

Switching a live cluster from the default TCP check to an HTTP or HTTPS check can mark every backend unhealthy at once if the backends do not answer the path yet, which takes the API server offline. So the switch is gated the same way as `EnableProxyProtocol`:

- CAPH switches in place only after every control-plane infrastructure machine carries `capi.syself.com/http-health-check-for-controlplane-loadbalancer: "true"`, set on the control-plane machine template.
- A machine from an earlier template does not carry it, so the switch happens only once the whole control-plane rollout has replaced every machine with one whose image answers the path. Until then the TCP check stays in place.
- A new cluster gets the configured check from the start (no migration).

## Testing

- Unit tests for the health-check option builders, the drift comparison, the migration decision, and the annotation gate (including that the proxy-protocol annotation does not satisfy the HTTP gate).
- Reconcile tests (with the fake hcloud client extended to model a service's health check) that prove the switch happens only once every control-plane machine is annotated, and requeues otherwise.
- `go build`, `go vet`, `gofmt`, and the loadbalancer + scope suites pass; CRDs and deepcopy regenerated and in sync.


pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -376,6 +376,11 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 			DestinationPort: &destinationPort,
 			Proxyprotocol:   &proxyProtocol,
 		}
+		if listenPort == kubeAPIServicePort {
```
janiskemper: this should be added further up to the existing if statement - without creating a new one

api/v1beta1/types.go
```diff
@@ -251,6 +271,55 @@ type LoadBalancerServiceSpec struct {
 	DestinationPort int `json:"destinationPort,omitempty"`
 }
 
+// LoadBalancerServiceHealthCheck configures the health check the load balancer runs
+// against the kube-apiserver service targets. When it is not set, the service uses
+// the default TCP check on the service port.
+type LoadBalancerServiceHealthCheck struct {
```
janiskemper: the "domain" seems to be missing here (see #2193)

api/v1beta1/types.go
```diff
@@ -251,6 +271,55 @@ type LoadBalancerServiceSpec struct {
...
+	// the port accepts a connection. http and https send a request to Path and
+	// count the target healthy when the response status is in StatusCodes.
+	// +kubebuilder:validation:Enum=tcp;http;https
+	// +kubebuilder:default=tcp
+	Protocol string `json:"protocol"`
+
+	// Path is the request path for an http or https check, for example "/readyz".
+	// It is required for http and https and ignored for tcp.
+	// +optional
+	Path string `json:"path,omitempty"`
```
janiskemper: we should have a webhook to validate this (see #2193)

janiskemper (changes requested):

abdullah599:
## Context

Simon opened [#2193](https://github.com/syself/cluster-api-provider-hetzner/pull/2193) on `main` with the same feature. We want both PRs to land with the same API so the two branches do not drift.

The shape we propose is `LoadBalancerHealthCheckSpec` with `intervalSeconds` and `timeoutSeconds`. Your field names already match, so only the struct name changes.

Your `port` and the `+kubebuilder:default=tcp` on `protocol` stay as they are. Simon's PR will add them instead.

Janis already flagged the missing `domain`, the missing webhook and the duplicate if block. They are repeated below so the whole list is in one place.

## What needs to change here

- Rename `LoadBalancerServiceHealthCheck` to `LoadBalancerHealthCheckSpec`.
- Make `path` a `*string`. Today it is a plain string that is always sent, so `protocol: http` without a path sends `"path": ""` to Hetzner.
- Add `domain` and `response`.
- Add the webhook that rejects the http only fields when the protocol is tcp.
- Fold the new `if listenPort == kubeAPIServicePort` into the existing one in the create loop, and reuse `existingKubeAPIService` instead of looking the service up a second time.
- Add the upper bounds Hetzner enforces: `intervalSeconds` 3 to 60, `timeoutSeconds` 1 to 60, `retries` 0 to 5.
- Fix the `statusCodes` doc comment. The wildcard form is `2??`, `2xx` is not a value the API takes.
- Add the docs section in `docs/caph/03-reference/02-hetzner-cluster.md`, including the note that an http check means the apiserver has to serve that path without auth.

The fake hcloud client health check support is the part we want to keep from this PR.

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+			// every backend unhealthy at once if the backends do not answer the path yet, which
+			// takes the API server offline. Gate that one-way switch on all control-plane
+			// machines carrying the annotation, the same as the proxy-protocol migration: the
+			// annotation is set on the new control-plane machine template, so the switch happens
+			// only after every control plane runs an image that serves the health-check path. A
+			// tcp check, or a change that stays within http, is applied without the gate.
+			if healthCheckMigratesToHTTP(existing.HealthCheck, desired) {
+				allReady, err := s.scope.AllControlPlaneInfraMachinesAnnotatedForHTTPHealthCheck(ctx)
+				if err != nil {
+					return reconcile.Result{}, err
```
abdullah599: this returns err on its own while the block just below joins into multierr. if the delete loop or the proxy-protocol update already failed this pass, that error gets dropped here. should be `errors.Join(multierr, err)`

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+			}
+		}
+	}
 	return reconcile.Result{}, multierr
 }
 
+// healthCheckAddOpts builds the hcloud health-check options for a new kube-apiserver
+// service from the spec. It returns nil when the spec has no health check, so the
+// service keeps the load balancer's default TCP check on servicePort.
+func healthCheckAddOpts(hc *infrav1.LoadBalancerServiceHealthCheck, servicePort int) *hcloud.LoadBalancerAddServiceOptsHealthCheck {
```
abdullah599: these three builders are the same ~30 lines, only the hcloud return type differs, so every field has to be kept in sync by hand across all three and nothing checks that it is. can we read the spec once into a small struct (port default, seconds to duration, the http block) and let each function just plug the values in? the `if hc.Retries != nil` guard can go too, plain assignment is already nil-safe

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
 			s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ProxyProtocolEnabled = true
 		}
 	}
+
+	// Reconcile the kube-apiserver service health check to the spec. Only touch it
+	// when the spec sets one, so a cluster that leaves HealthCheck unset keeps the
```
abdullah599: same claim as the field doc. a cluster that clears HealthCheck does not go back to the tcp check, this block is skipped when desired is nil so the last check applied stays. worth saying that here

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	if desired.Retries != nil && observed.Retries != *desired.Retries {
+		return true
+	}
+	if isHTTPHealthCheck(desired.Protocol) {
+		if observed.HTTP == nil ||
+			observed.HTTP.Path != desired.Path ||
+			observed.HTTP.TLS != (desired.Protocol == "https") {
+			return true
+		}
+		if len(desired.StatusCodes) > 0 && !slices.Equal(observed.HTTP.StatusCodes, desired.StatusCodes) {
```
abdullah599: nit: slices.Equal is order sensitive. if hcloud ever echoes the codes back in another order this diffs forever and we call update_service every reconcile. status codes are a set, can we compare them sorted?

pkg/scope/cluster.go
```diff
@@ -279,6 +279,26 @@ func IsControlPlaneReady(ctx context.Context, c clientcmd.ClientConfig) error {
...
 // control-plane infrastructure machines yet.
 func (s *ClusterScope) AllControlPlaneInfraMachinesAnnotatedForProxyProtocol(ctx context.Context) (bool, error) {
+	return s.allControlPlaneInfraMachinesAnnotated(ctx, infrav1.ProxyProtocolForControlPlaneLoadBalancerAnnotation, "proxy protocol")
+}
+
+// AllControlPlaneInfraMachinesAnnotatedForHTTPHealthCheck returns true when every control-plane
+// infrastructure machine carries capi.syself.com/http-health-check-for-controlplane-loadbalancer:
+// "true", set on the control-plane machine template. It gates the one-way switch of the
+// control-plane load balancer health check from tcp to http in place, so the switch happens only
+// after every control plane runs an image that answers the health-check path. It works the same
```
abdullah599: "an image that answers the health-check path" claims something we never check, we only read the annotation. the proxy-protocol doc right above stays in template terms. can we match that wording here, and in the annotation constant and loadbalancer.go, so all three say the machine comes from a template that expects an http check?

pkg/services/hcloud/client/fake/hcloud_client.go
```diff
@@ -425,10 +425,41 @@ func (c *cacheHCloudClient) AddServiceToLoadBalancer(_ context.Context, lb *hclo
...
+		service.HealthCheck = serviceHealthCheckFromOpts(hc.Protocol, hc.Port, path, tls, codes)
+	}
 	c.loadBalancerCache.idMap[lb.ID].Services = append(c.loadBalancerCache.idMap[lb.ID].Services, service)
 	return nil
 }
 
+// serviceHealthCheckFromOpts builds an observed health check from the fields the add and update
+// options carry, so a reconcile that sets a health check is reflected back on the next read. It
+// takes primitives because the add and update option HTTP structs are separate types with the
+// same fields, and it carries the fields the load-balancer reconcile compares: protocol, port,
```
abdullah599: this says it carries the fields the reconcile compares, but healthCheckDiffers also compares Interval, Timeout and Retries and none of them are set here. a test that sets one of them would see the service drift from the spec on every reconcile. can the fake reflect them too, plus domain and response once those land?

pkg/services/hcloud/loadbalancer/healthcheck_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+func TestHealthCheckAddOpts(t *testing.T) {
+	if got := healthCheckAddOpts(nil, 6443); got != nil {
+		t.Fatalf("healthCheckAddOpts(nil) = %v, want nil so the service keeps the default TCP check", got)
+	}
+
+	hc := &infrav1.LoadBalancerServiceHealthCheck{
+		Protocol:    "https",
+		Path:        "/readyz",
+		StatusCodes: []string{"200"},
```
abdullah599: StatusCodes is set here but never asserted below, and nothing in the file sets TimeoutSeconds, so dropping the statuscodes copy or the timeout conversion would still pass. can the three builder tests share one fixture that sets every field and assert each one back?

pkg/services/hcloud/loadbalancer/loadbalancer_test.go
```diff
@@ -337,10 +337,137 @@ var _ = Describe("createOptsFromSpec", func() {
...
+
 	It("returns ErrControlPlaneEndpointNotSet", func() {
 		hetznerCluster.Spec.ControlPlaneEndpoint = nil
 
 		_, err := createOptsFromSpec(hetznerCluster)
 		Expect(errors.Is(err, ErrControlPlaneEndpointNotSet)).To(BeTrue())
 	})
 })
+
+var _ = Describe("reconcileServices health check migration", func() {
```
abdullah599, line 368: both specs here cover tcp -> http. the branch that applies a change without the gate (already http, only the path changes) has no test, so a regression routing everything through the gate would pass CI. can we add a spec for a path change on an already http check with no annotated machines?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+			// only after every control plane runs an image that serves the health-check path. A
+			// tcp check, or a change that stays within http, is applied without the gate.
+			if healthCheckMigratesToHTTP(existing.HealthCheck, desired) {
+				allReady, err := s.scope.AllControlPlaneInfraMachinesAnnotatedForHTTPHealthCheck(ctx)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+
+				if !allReady {
+					s.scope.V(1).Info("health check: not all control-plane machines are ready for the http check, keeping the tcp check")
```
abdullah599: this logs "health check:" but the gate it just called logs "http health check:". same prefix so one grep catches both?

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+					return reconcile.Result{RequeueAfter: 10 * time.Second}, multierr
+				}
+			}
+
+			updateOpts := hcloud.LoadBalancerUpdateServiceOpts{HealthCheck: healthCheckUpdateOpts(desired, kubeAPIDestinationPort)}
+			if err := s.scope.HCloudClient.UpdateServiceOnLoadBalancer(ctx, lb, kubeAPIServicePort, updateOpts); err != nil {
+				hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "UpdateServiceOnLoadBalancer")
```
abdullah599, line 453: the other three HandleRateLimitExceeded sites have `// return immediately on rate limit` above them, this one is missing it

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,214 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
 		}
 	}
+
+	// Reconcile the kube-apiserver service health check to the spec. Only touch it
+	// when the spec sets one, so a cluster that leaves HealthCheck unset keeps the
+	// load balancer's default TCP check. A service created above already carries the
+	// check from its add options, so this only updates a service that existed before
+	// this reconcile.
+	if desired := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.HealthCheck; desired != nil {
+		kubeAPIDestinationPort := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.Port
```
abdullah599, line 429: a word saying this is the destination port, not kubeAPIServicePort (the listen port) two lines down, would save the reader a hunt

api/v1beta1/hetznercluster_types.go
```diff
@@ -55,6 +55,20 @@ const (
 	// For NEW clusters created with EnableProxyProtocol: true, this annotation is never read:
 	// the LB service is created with proxy protocol from the start, so no migration is needed.
 	ProxyProtocolForControlPlaneLoadBalancerAnnotation = "capi.syself.com/proxy-protocol-for-controlplane-loadbalancer"
+
+	// HTTPHealthCheckForControlPlaneLoadBalancerAnnotation is used only when switching the
+	// control-plane load balancer health check from tcp to http or https on an EXISTING cluster
+	// (migration path). It must be present with value "true" on ALL control-plane machines before
+	// CAPH switches the load balancer health check in place. The annotation is set on the
+	// control-plane machine template, so it is present exactly on the machines whose template
```
janiskemper: it's set on the infra machines, no?

api/v1beta1/types.go
```diff
@@ -232,6 +232,30 @@ type LoadBalancerSpec struct {
...
+	// kube-apiserver service targets. As long as it has never been set, the service uses
+	// a plain TCP check on the service port, so a target counts as healthy as soon as the
+	// port accepts a connection.
+	//
+	// Set an http or https check with a Path to have the load balancer route only
+	// to a target whose API server answers that path (for example a readiness
+	// endpoint), and drop a target that stops answering it. Only the kube-apiserver
+	// service is affected; extra services are never given a custom health check.
+	//
+	// Removing HealthCheck again does not bring the TCP check back. CAPH only writes the
```
janiskemper: This sentence seems to contradict with the one below. Is it now allowed to change it or not?

docs/caph/03-reference/02-hetzner-cluster.md
```diff
@@ -28,6 +28,40 @@ If you are using your own load balancer, you need to point towards it (by settin
...
+rotation instead of receiving traffic. The field names and their limits mirror the Hetzner Cloud
+API's load balancer
+[`health_check` object](https://docs.hetzner.cloud/reference/cloud#tag/load-balancer-actions/add_load_balancer_service).
+
+The endpoint the check requests must answer without authentication, because the load balancer cannot
+present credentials. CAPH does not configure that for you. Use `healthCheck.port` when that endpoint
+is served on a different port than the API server itself.
+
+CAPH sends only the fields you set. A field you never set keeps Hetzner's own default for it (tcp,
+15s interval, 10s timeout, 3 retries). If `healthCheck` is left out entirely, CAPH never touches the
```
janiskemper: where do these defaults come from? If from Hetzner, we should maybe rather link something than hardcoding them here - they might get outdated quickly.

pkg/services/hcloud/client/fake/hcloud_client.go
```diff
@@ -425,10 +426,75 @@ func (c *cacheHCloudClient) AddServiceToLoadBalancer(_ context.Context, lb *hclo
...
+	Domain      *string
+	Path        *string
+	Response    *string
+	StatusCodes []string
+	TLS         *bool
+}
+
+// serviceHealthCheckFromOpts builds an observed health check from the fields the add and update
+// options carry, so a reconcile that sets a health check is reflected back on the next read. It
+// covers every field healthCheckDiffers compares, so a test that sets a field does not see the
```
janiskemper: healthCheckDiffers is a utility from another package. Let's not mention it here

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,258 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	// service created above already carries the check from its add options, so this only
+	// updates a service that existed before this reconcile.
+	if desired := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.HealthCheck; desired != nil {
+		// The port the check runs against by default. kubeAPIServicePort below is the port the
+		// load balancer listens on, which is a different number.
+		kubeAPIDestinationPort := s.scope.HetznerCluster.Spec.ControlPlaneLoadBalancer.Port
+		if kubeAPIServiceExists && healthCheckDiffers(existingKubeAPIService.HealthCheck, desired, kubeAPIDestinationPort) {
+			// Switching a live service from a tcp check to an http or https check can mark
+			// every backend unhealthy at once if the backends do not answer the path yet, which
+			// takes the API server offline. Gate that one-way switch on all control-plane
```
janiskemper: let's formulate "gate that one-way switch" differently - this is very AI-like

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,258 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+			// machines carrying the annotation, the same as the proxy-protocol migration: the
+			// annotation is set on the new control-plane machine template, so the switch happens
```
janiskemper: infra machine

pkg/services/hcloud/loadbalancer/loadbalancer.go
```diff
@@ -413,9 +418,258 @@ func (s *Service) reconcileServices(ctx context.Context, lb *hcloud.LoadBalancer
...
+	StatusCodes []string
+	// TLS is set only for an http or https check, so it also marks whether the HTTP
+	// sub-options have to be built at all.
+	TLS *bool
+}
+
+// healthCheckOptsFromSpec reads the spec into the fields the hcloud option types share. It
+// returns nil when the spec has no health check, and the caller then leaves the load balancer's
+// own check alone. servicePort is used when the spec sets no port.
+func healthCheckOptsFromSpec(hc *infrav1.LoadBalancerHealthCheckSpec, servicePort int) *healthCheckOpts {
```
janiskemper: let's put this utility below the places where it is used

janiskemper (changes requested):

api/v1beta1/types.go
janiskemper: I think that this is still interesting. We have a migration path for the change from tcp to http, AFAIK every time you change it (even if it is the fifth time). When you change it in the other direction, there is no such migration and it would have immediate effect. 

I think we should write that down, wdyt?

Also: Why does empty mean the last applied configuration, and not the Hetzner default?

abdullah599: You are right healthCheckMigratesToHTTP compares the live protocol with the spec on every reconcile, so the wait happens on every tcp to http switch. going back to tcp or changing path is applied right away. Documented now

Also,
Leaving HealthCheck out means CAPH does not manage the check, so a load balancer that already has one keeps it. Removing the field therefore keeps the last check CAPH applied

janiskemper (approved):

PR #2206, opened 2026-07-28
:sparkles: Make the baremetal controlplane loadbalancer address family configurable

batistein:

CAPH attaches a bare metal control plane server to the control plane load balancer by address, and a target holds exactly one address, so reaching a server over both IPv4 and IPv6 needs two separate targets. This PR adds `spec.controlPlaneLoadBalancer.targetAddressFamily` to `HetznerCluster` to choose which addresses are attached: `ipv4`, `ipv6`, or `dualstack`. An unset field means `dualstack`, so the default keeps the current behavior of attaching both.

Set `ipv4` on a cluster whose servers only use IPv4. With the default `dualstack`, CAPH also attaches the server's IPv6 address, and if the server does not answer on IPv6 that target never becomes healthy: it stays unhealthy for as long as the machine exists, makes the load balancer look unhealthy while the control plane is fine, and takes up a target slot. CAPH cannot detect this itself, because Hetzner gives every bare metal server an IPv6 address in the Robot API but does not report whether the installed OS actually uses it, and images that set up only IPv4 are common. Set `ipv6` for a cluster that only uses IPv6.

Only bare metal servers are affected. An HCloud server is attached by its server ID (`server` target type) and the load balancer resolves its addresses on its own. Worker machines are never attached to the control plane load balancer.

You can change the field at any time. CAPH attaches the addresses of the newly selected family that are missing and detaches the targets it no longer selects, so a switch converges without manual cleanup in the HCloud API.

**TODOs**:

- [x] squash commits
- [x] include documentation
- [x] add unit tests


api/v1beta1/types.go
```diff
@@ -232,6 +251,73 @@ type LoadBalancerSpec struct {
...
+	//
+	// This only affects bare metal servers. An HCloud server is a resource of the HCloud
+	// API, so the load balancer references it by its server ID ("server" target type) and
+	// resolves the address on its own. A bare metal server is a Robot resource that the
+	// load balancer cannot reference that way, so it is attached by address instead ("ip"
+	// target type), and such a target holds exactly one address. Reaching one server over
+	// both protocols therefore takes two targets, and this field decides which of them
+	// CAPH creates.
+	//
+	// An empty value means ipv4, so by default only the IPv4 address is attached. Hetzner
```
janiskemper: let's change this default to dualstack

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -712,25 +750,17 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
...
 			IP: net.ParseIP(ip),
 		}
 
 		if err := s.scope.HCloudClient.AddIPTargetToLoadBalancer(ctx, opts, lb); err != nil {
 			hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "AddIPTargetToLoadBalancer")
+			// The address is already a target, which is the state we want, so carry on
+			// with the next address instead of giving up on it.
 			if hcloud.IsError(err, hcloud.ErrorCodeTargetAlreadyDefined) {
-				return nil
+				continue
```
abdullah599, line 771: nothing tests this continue, and it's the change that keeps dualstack attaching the second address after the first comes back already-defined. a silent revert to 'return nil' would still pass CI. add a case where AddIPTarget returns TargetAlreadyDefined for the first IP and assert the second still attaches?

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -745,6 +775,37 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
...
+	return address
+}
+
+// detachIPTargetOfLoadBalancer removes a single address of the host from the load
+// balancer. An address that is not a target anymore is the desired outcome, so the
+// corresponding error of the HCloud API is not treated as a failure.
+func (s *Service) detachIPTargetOfLoadBalancer(ctx context.Context, lb *hcloud.LoadBalancer, host *infrav1.HetznerBareMetalHost, ip string) error {
+	if err := s.scope.HCloudClient.DeleteIPTargetOfLoadBalancer(ctx, lb, net.ParseIP(ip)); err != nil {
+		hcloudutil.HandleRateLimitExceeded(s.scope.HetznerCluster, err, "DeleteIPTargetOfLoadBalancer")
+		if !strings.Contains(err.Error(), "load_balancer_target_not_found") {
```
abdullah599: the detach error path isn't covered, every detach test returns nil so this not_found swallow-vs-propagate branch never runs. worth one detach returning a generic error (expect it to propagate) and one returning load_balancer_target_not_found (expect Succeed)?

api/v1beta1/types.go
```diff
@@ -232,6 +252,67 @@ type LoadBalancerSpec struct {
...
+	switch spec.TargetAddressFamily {
+	case LoadBalancerTargetAddressFamilyIPv4, LoadBalancerTargetAddressFamilyIPv6:
+		return spec.TargetAddressFamily
+	default:
+		return LoadBalancerTargetAddressFamilyDualStack
+	}
+}
+
+// WantsIPv4 reports whether the configured address family selects the IPv4 address of a
+// bare metal server for attachment. Whether that address is attached right now is a
```
janiskemper: can you remove the sentence here and below "whether that ..."? I think that's totally obvious and therefore confuses people

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Bare metal control planes and the load balancer
+metatitle: Choose Which Addresses of a Bare Metal Control Plane Server Serve API Traffic
+sidebar: Load balancer targets
+description: Select the address family CAPH uses when it attaches a bare metal control plane server to the control plane load balancer.
+---
+
+## Why bare metal servers are different
```
janiskemper:
```suggestion
## Difference between bare metal and cloud servers
```

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
```diff
@@ -0,0 +1,77 @@
...
+
+An HCloud server is a resource of the HCloud API, the same API the load balancer belongs to, so CAPH attaches it by its server ID. This is a target of type `server`, and the load balancer resolves the address of that server on its own. There is nothing to choose.
+
+A bare metal server is a Hetzner Robot resource. It has a server ID too, but not one the HCloud load balancer can reference, so CAPH attaches it by address instead. This is a target of type `ip`, and such a target holds exactly one address. Reaching one server over both protocols therefore takes two targets, and CAPH has to decide which of them to create.
+
+`spec.controlPlaneLoadBalancer.targetAddressFamily` is that decision.
+
+The field is limited to that decision, so two things it does not touch are worth naming. Worker machines are never attached to the control plane load balancer, bare metal or not, so the field does not apply to them. And a `Service` of type `LoadBalancer` inside the workload cluster gets its own load balancer from the cloud controller manager, not from CAPH, so the targets of that one are configured there instead.
```
janiskemper: this text is weird. Can we change it?

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
```diff
@@ -0,0 +1,77 @@
...
+    targetAddressFamily: ipv4
+```
+
+The field is optional. Leaving it out is the same as setting `dualstack`.
+
+An address the server does not have is never attached. If a server has no IPv6 address, `dualstack` attaches its IPv4 address and nothing else.
+
+## When to set ipv4
+
+Hetzner routes an IPv6 subnet to a bare metal server, so an IPv6 address for it exists in the Robot API, and CAPH records it as `spec.status.ipv6` on the `HetznerBareMetalHost`. That an address exists is not the same as the server answering on it: the installed OS still has to configure it, and an image that sets up IPv4 only is common. The Robot API does not report what the running OS configured, so CAPH cannot detect this and choose for you.
```
janiskemper: "That an address" -> grammar 

I think you can rephrase the whole paragraph to make it easier to udnerstand. Basically, the paragraph states that the user has to think about whether he wants IPv6 or not and this is directly related to the general architecture he chooses for the nodes and node images.

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
```diff
@@ -0,0 +1,77 @@
...
+
+When the server does not answer on the address, the target is still created and simply never passes its health check. The load balancer then reports an unhealthy target for as long as the machine exists, which buries a genuinely unhealthy control plane in the noise, and one target slot is spent on a target that cannot serve traffic.
+
+The default is `dualstack`, so both addresses are attached. Set `ipv4` on a cluster whose servers only use IPv4, so the load balancer does not carry an IPv6 target that never becomes healthy. Set `ipv6` for a single-stack IPv6 setup.
```
janiskemper: you can remove this paragraph IMO. It's duplicate

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
```diff
@@ -0,0 +1,77 @@
...
+
+Each change is visible in `status.controlPlaneLoadBalancer.targets` on the `HetznerCluster`, which lists the targets as they exist on the load balancer:
+
+```shell
+kubectl get hetznercluster <name> -o jsonpath='{.status.controlPlaneLoadBalancer.targets}'
+```
+
+CAPH also emits events on the `HetznerCluster` when it attaches or detaches an address.
+
+## Upgrading an existing cluster
```
janiskemper: I think we can remove this as well. It seems also duplicate

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -681,16 +681,53 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
...
+	// Nothing to do, which is the case in the vast majority of reconciles.
+	if len(addressesToAttach) == 0 && len(addressesToDetach) == 0 {
+		return nil
 	}
-	if host.Spec.Status.IPv6 == "" {
-		foundIPv6 = true
+
+	lb := &hcloud.LoadBalancer{ID: s.scope.HetznerCluster.Status.ControlPlaneLoadBalancer.ID}
+
+	// Detaching runs before the health gate below. An address that is not selected does
```
janiskemper: I don't think that this is a good reason for running detaching first...

Failing to detach and therefore not attaching could mean that something is not working at all, while detaching is an operation that doesn't matter in 99% of the cases.

I see only one actual reason to do it this way around: In case the target limit is reached, we first need to detach before we can attach. But also this doesn't really make sense, because it is very unlikely to happen (could only happen in one reconcile if we switch from IPv6 to IPv4 only, which is basically impossible.

So should we conclude that there is a reason, or just do one thing after another without reason?

pkg/services/baremetal/baremetal/baremetal.go
```diff
@@ -681,16 +681,53 @@ func (s *Service) reconcileLoadBalancerAttachment(ctx context.Context, host *inf
...
+	// not belong on the load balancer regardless of the state of the API server, and
+	// waiting would keep a target that cannot serve traffic.
+	for _, ip := range addressesToDetach {
```
janiskemper, line 721: in the current way how we do it I think it's better to add this code above without creating a separate slice. 

The difference is that we created another utility for detaching, which we don't have for attaching. 

We can do it either in one way or in another way - but right now it seems strangely mixed.

docs/caph/02-topics/05-baremetal/05-load-balancer-targets.md
janiskemper: @abdullah599 can you also replace the "... is that decision"?

janiskemper (approved):

PR #2207, opened 2026-07-28
✨ PoC: observe whether a server reached the rescue system, and act on it

batistein:
> **Proof of concept, opened as a draft to discuss the approach.** It changes how the rescue boot is verified during `imageURL` provisioning, and touches the provisioning path of every hcloud machine using that flow. Everything here is up for challenge, including whether the approach is right at all.

## The idea

Today CAPH **infers** whether a server reached the rescue system: SSH does not answer, so assume it is still booting, and if six minutes pass, give up. That is a guess with a timer on it, and the diagnosis it produced when we hit this was wrong about what had happened.

This PR stops inferring and starts observing. The API states the fact directly, and once we read it, the six minutes stops being a period to wait out and becomes a budget to fix the problem in.

## Background: how the imageURL flow boots

A Hetzner server must be created from some image. For this flow CAPH creates it from a throwaway one and arranges for that image never to run:

1. Create the server from `ubuntu-24.04`, powered off, no user-data.
2. Enable the rescue system while it is off.
3. Power it on, so the first boot lands in rescue.
4. From inside rescue, write the real image over the disk.

The source says so directly:

```go
// Create the server powered off. Enabling the rescue system while the server is off means
// its first boot goes directly into the rescue system, and the pre-rescue-OS image never boots.
```

So a stock image always sits on that disk, and the design depends on it never executing.

## What we observed

```
08:23:12  create_server   success   finished 08:25:18   (126s)
08:25:18  enable_rescue   success   finished 08:25:29    (11s)
08:25:30  start_server    success   finished 08:25:42    (12s)   <- issued 1s after rescue armed
08:25:45  dial tcp :22 i/o timeout        (repeats)
08:31:31  BootingToRescue -> ProvisioningFailed
          "get hostname failed: ssh i/o timeout ... in this state since 6m1s"
          "hcloudmachine: ProvisioningFailed. Not reconciling this machine."
08:32:41  the server finished booting -- into the throwaway image, 70s after CAPH gave up
```

It booted the wrong image. CAPH could not tell, so it reported an SSH problem for what was an OS-identity problem. The code says why it does not check:

```go
// The server is freshly created and was never started before this boot cycle, so there is no
// prior OS it could mistakenly reach over SSH.
```

There is one, and it is what booted.

Across every server in that project using this flow, the failed one is the only one where `start_server` was issued 1 second after `enable_rescue` finished. The others had 4-5 seconds and all entered rescue. A small sample, not proof, but consistent with the action reporting success before the boot configuration is in effect.

## The fact this rests on

`rescue_enabled` is cleared when the server boots into the rescue system. So a server that is **running** while the flag is **still set** booted something else, and no further SSH polling will change that.

The failed machine supports it: the flag is still set after hours of running the wrong image. Every server that provisioned normally has it cleared.

## The state model

| `rescue_enabled` | server | SSH | meaning | action |
|---|---|---|---|---|
| true | off | — | a recovery step powered it off | power on |
| true | starting/stopping | — | mid-transition | let it settle |
| true | running | silent | booting, or booted wrong with no sshd yet | observe again |
| true | running | answers | **booted the wrong image** | recover |
| false | running | `rescue` | in rescue | proceed |
| false | running | other | rescue consumed, wrong image running | existing hostname check |
| false | any | — | rescue entered | hand back to SSH |

## What changes

**Confirm the rescue system is armed before powering on.** `handleBootStateEnablingRescue` currently powers the server on using a synthesised `&hcloud.Server{ID: serverID}`, verifying only that the *action* finished. It now reads the server back and proceeds only when `RescueEnabled` is true, otherwise requeues with a `WaitingForRescueEnabled` condition.

**Observe, continuously, once a boot has had long enough.** SSH stays the first check, so a healthy boot spends nothing here. After a 45s grace period the API is consulted every 15s, because the answer is what the recovery steps are driven from.

**Recover in steps instead of giving up.**

| step | reading | action |
|---|---|---|
| `PowerCycled` | the arming was fine, the boot was not | power off and on again |
| `Rearmed` | the arming never took effect | enable the rescue system afresh, then power-cycle |
| exhausted | neither worked | report `RescueNotEntered`, stop acting |

The `Rearmed` step benefits from the first change: the armed check gates that boot too, so the retry does not repeat the first attempt's mistake.

The step is recorded in `status.rescueRecovery`, so escalation is visible in status and no step repeats. The existing six-minute timeout stays the single place that decides a machine has failed — this only makes reaching it much less likely.

## Cost

One `GetServer` per machine provisioned, from the armed check. Then one per observation while a server is failing to reach rescue, which is bounded by the state's own timeout and costs nothing on a healthy boot.

An existing assertion caps `GetServer` calls during `imageURL` provisioning. It moves from 1 to 2 with the reason recorded inline rather than being removed, because it exists to stop this path quietly growing API calls against a rate limit shared by every machine being provisioned.

## Testing

`make test-unit` passes, 148 tests; `make lint` clean. Seven specs cover the state model and the ladder: not powering on while unarmed, powering off a running-but-armed server, powering back on afterwards, re-enabling the rescue system on the second failure, stopping once both steps are spent, handing back to SSH when the flag clears, and leaving alone a server something else powered off. Server-package coverage 55.1% -> 60.3%.

The fake client's `EnableRescueSystem` was a no-op; it now sets `RescueEnabled` to match the real API, since callers depend on that flag.

## Deliberately out of scope

- The six-minute timeout and remediation are unchanged. The timeout stays the only place a machine is declared failed, and deleting or recreating a machine remains remediation's job, not the provider's.
- The `imageName` flow, which does not use the rescue system.
- Identifying *which* image booted. `rescue_enabled` answers "is this the rescue system", which is the only question that matters here.

## Open questions

- 45s grace and 15s observation interval are first guesses. The intent is "past any healthy boot" without spending API calls early.
- Two recovery steps may be the wrong number, and there may be a third worth having.
- During the incident CAPH's behaviour implies it read `rescue_enabled` as false, while the API reports true now, with a single `enable_rescue` action on record. That transition is unexplained and may warrant asking Hetzner.

### DIFF
--- a/api/v1beta1/conditions_const.go
+++ b/api/v1beta1/conditions_const.go
@@ -422,6 +422,15 @@ const (
 	HCloudMachineBootingToRescueV1Beta2Reason = "BootingToRescue"
 	// HCloudMachineBootingToRescueTimedOutV1Beta2Reason indicates booting to rescue mode timed out.
 	HCloudMachineBootingToRescueTimedOutV1Beta2Reason = "BootingToRescueTimedOut"
+	// HCloudMachineWaitingForRescueEnabledV1Beta2Reason indicates the API has not yet reported the
+	// rescue system as armed, so the server is not powered on yet.
+	HCloudMachineWaitingForRescueEnabledV1Beta2Reason = "WaitingForRescueEnabled"
+	// HCloudMachineRescueNotEnteredV1Beta2Reason indicates the server booted something other than
+	// the rescue system and the recovery steps did not change that.
+	HCloudMachineRescueNotEnteredV1Beta2Reason = "RescueNotEntered"
+	// HCloudMachineRetryingBootToRescueV1Beta2Reason indicates a recovery step is in progress to
+	// get the server into the rescue system.
+	HCloudMachineRetryingBootToRescueV1Beta2Reason = "RetryingBootToRescue"
 
 	// HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason indicates the image URL command is not accessible.
 	HCloudMachineImageURLCommandNotAccessibleV1Beta2Reason = "ImageURLCommandNotAccessible"
--- a/api/v1beta1/hcloudmachine_types.go
+++ b/api/v1beta1/hcloudmachine_types.go
@@ -183,7 +183,31 @@ type HCloudMachineStatus struct {
 	// Used to prevent reboot loops across successive MHC incidents.
 	// +optional
 	LastRemediatedAt *metav1.Time `json:"lastRemediatedAt,omitempty"`
+
+	// RescueRecovery records the last recovery step taken for a server that was powered on to
+	// the rescue system and booted something else. It makes the escalation visible and stops it
+	// repeating a step that already failed.
+	// +optional
+	RescueRecovery RescueRecoveryStep `json:"rescueRecovery,omitempty"`
 }
+
+// RescueRecoveryStep is a step taken to get a server into the rescue system after it booted
+// something else.
+// +kubebuilder:validation:Enum="";PowerCycled;Rearmed
+type RescueRecoveryStep string
+
+const (
+	// RescueRecoveryNone means no recovery has been attempted.
+	RescueRecoveryNone RescueRecoveryStep = ""
+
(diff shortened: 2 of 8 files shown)


janiskemper:
closing in favor of #2211.

PR #2213, opened 2026-07-31
✨ Add condition for proxy protocol activation wait

### LINKED ISSUE #2212: Add Condition for waiting to activate proxy protocol
/kind feature

We should add a condition that specifies that we are waiting for the control planes to have the annotations to then activate proxy protocol.

We also want to increase the requeue time from 10 sec to min 2 minutes to not hammer the API.

abdullah599:
When proxy protocol is enabled on the control-plane load balancer for an existing cluster, activation waits until every control-plane machine carries the proxy-protocol annotation. Until now this wait was invisible: `LoadBalancerReady` stayed stale-True and the reconcile requeued every 10 seconds.

This surfaces the wait as a reason on the existing `LoadBalancerReady` condition (`Reason=WaitingForProxyProtocol`, with a message explaining it waits for all control-plane machines to be annotated), and raises the requeue interval from 10 seconds to 2 minutes so we stop hammering the API.

Fixes #2212

Followed the existing load balancer condition convention — every LB state here is a reason on `LoadBalancerReady` (v1beta1 + v1beta2 set together), so this adds a reason, not a standalone condition. `LoadBalancerReady` does not gate the CAPI provisioning contract (that is `status.initialization.provisioned`, driven by the LB IPv4), so this is observability only; the cluster `Ready` summary will show False with this reason during the rollout, which is honest since the LB is not yet at its desired spec.

**TODOs**:
- [x] add unit tests
- [ ] squash commits before merge
- documentation: none needed (reason constants only, no CRD/schema change)

### DIFF
--- a/api/v1beta2/conditions_const.go
+++ b/api/v1beta2/conditions_const.go
@@ -31,8 +31,8 @@ const (
 	LoadBalancerServiceSyncFailedV1Beta1Reason = "LoadBalancerServiceSyncFailed"
 	// LoadBalancerFailedToOwnV1Beta1Reason used when no owned label could be set on a load balancer.
 	LoadBalancerFailedToOwnV1Beta1Reason = "LoadBalancerFailedToOwn"
-	// LoadBalancerWaitingForProxyProtocolV1Beta1Reason used while proxy protocol activation waits for all control-plane machines to be annotated.
-	LoadBalancerWaitingForProxyProtocolV1Beta1Reason = "LoadBalancerWaitingForProxyProtocol"
+	// LoadBalancerWaitingToActivateV1Beta1Reason used while proxy protocol activation waits for all control-plane machines to be annotated.
+	LoadBalancerWaitingToActivateV1Beta1Reason = "LoadBalancerWaitingToActivate"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


api/v1beta2/conditions_const.go
```diff
@@ -31,6 +31,8 @@ const (
 	LoadBalancerServiceSyncFailedV1Beta1Reason = "LoadBalancerServiceSyncFailed"
 	// LoadBalancerFailedToOwnV1Beta1Reason used when no owned label could be set on a load balancer.
 	LoadBalancerFailedToOwnV1Beta1Reason = "LoadBalancerFailedToOwn"
+	// LoadBalancerWaitingForProxyProtocolV1Beta1Reason used while proxy protocol activation waits for all control-plane machines to be annotated.
+	LoadBalancerWaitingForProxyProtocolV1Beta1Reason = "LoadBalancerWaitingForProxyProtocol"
```
janiskemper: I think "WaitingToActivate" is a better formulation, no? otherwise people might think that it comes from somewhere else

janiskemper (commented):
I don't think this is what you wanted..
Activate what? :D

janiskemper (approved):

PR #2216, opened 2026-07-31
✨ Add opt-in server recycling for HCloudMachines

Rene-Roscher:

Adds an opt-in `recycle` field to `HCloudMachineSpec`. When it's on, instead of creating a fresh server and deleting it on scale-down, the controller claims a pre-existing server labelled `caph-recycle=true`, rebuilds it with the machine's image and bootstrap data, and puts it back in the pool on delete instead of destroying it.

This lets a fixed set of servers back a MachineDeployment — e.g. reserved servers you want to keep around — so scaling reuses them instead of creating and destroying a server every time. It only works with `imageName` (snapshot); the webhook rejects it together with `imageURL`, since a rebuild can't reproduce the rescue-system install. It needs the `rebuild`-with-`user_data` API, so hcloud-go is bumped to v2.47.

Everything new lives in `pkg/services/hcloud/server/recycle.go` plus two small hooks in the create and delete paths. The boot state machine is untouched — a recycled server goes through the normal reconcile like a freshly created one.

- Claiming isn't atomic (Hetzner has no compare-and-swap on labels), so it relies on machine reconciles being serialized, i.e. the default `--hcloudmachine-concurrency=1`. Documented under Limitations.
- Tested end-to-end against real Hetzner (kind management cluster, workload cluster from a real RKE2 snapshot): claim → rebuild → node joins Ready, return-to-pool on scale-down, repeated up/down reusing the same servers, several servers claimed at once, killing the controller mid-claim, and network detach on return. Unit tests cover the claim/return logic and the webhook.

**TODOs**:

- [x] squash commits
- [x] include documentation
- [x] add unit tests


janiskemper:
Hi @Rene-Roscher ! Thanks a lot for your contribution!

I have two architectural questions / remarks:

1. Cluster API has a feature of in-place upgrades. I'm not aware that anyone actually tried using it with CAPH, but the use case seems very similar to yours. Why not trying that?
2. The PR states that this is not possible for the "image-url" way. I don't believe that this is true though. It actually wouldn't even need a separate endpoint, but basically it would just be about booting into rescue another time. What's your take on this?

Rene-Roscher:
Hi @janiskemper, thanks for taking a look.

**A bit of background on why this exists**: I was trying to get out from under the ~300% price increase of the last few months. I had already moved my RKE2 cluster onto CAPH (rebuilt from scratch), and the old servers on the cheaper pricing were still around. I wanted to keep using those instead of ordering new ones, but there was no clean way to import them so that CAPH picks them up as part of a MachineDeployment. That is the gap this fills.

**1. In-place upgrades**

I think these solve an adjacent but different axis. In-place mutates a Machine that CAPI already manages, without replacing it. What I needed was the opposite: **taking existing servers that CAPI does not know about yet and pulling them into an MD**, then keeping them across scale events instead of destroying them. In-place avoids replacement when you change a machine, not when you scale or adopt one.

For the upgrade/roll side, in-place is probably the nicer long-term path, and a CAPH extension for it would be interesting. Today it is alpha/feature-flagged and, as far as I know, has not been used with CAPH yet, so it was not something I could rely on for this use case. I would see the two as complementary.

**2. imageURL**

You are right, and my webhook comment is too absolute. The accurate, narrower statement is: the hcloud Rebuild API (which the recycle path uses) cannot reproduce an imageURL install, because it only takes a snapshot. But imageURL provisioning does not go through Rebuild at all; it is the rescue state machine (enable rescue, boot into rescue, run the script over SSH, reboot). That can simply be re-run on the claimed server, exactly as you say, with no separate endpoint. I will reword the webhook rationale. Right now it reasons "a rebuild cannot reproduce it, so only imageName works", which does not follow, since the rescue path needs no rebuild.

I left it out for scoping reasons, not because it is impossible. The effort is contained but not zero: the rescue flow currently assumes a freshly created, powered-off server whose first power-on lands in rescue. A recycled server is already running its OS, so it would need EnableRescueSystem plus a hard reset instead of the power-on, plus closing the "SSH reaches the old OS before the reset" window (today only guarded by the hostname == "rescue" check) and skipping the Initializing create-action guard. The disk state is a non-issue: the image-url script rewrites the whole device (sda) anyway.

janiskemper:
Thanks @Rene-Roscher!

I understood the use case and why, at this moment, you wouldn't benefit from the in-place upgrades. 

I'm fine with having this feature for now only for the builds from snapshot and not via the rescue system flow. 

You should check that the docs in your PR are aligned with that. 

We'll be able to review it in a few days probably - right now we are in the release process.

pkg/services/hcloud/server/recycle.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// reserveRecyclableServer takes a server out of the available pool by replacing its labels with the
+// recycle marker and a claimant label naming this machine. The claimant label is intentionally not the
+// label the server lookup matches on, so a reserved server is not yet adopted as this machine's server.
+//
+// Label writes are last-writer-wins and Hetzner offers no conditional (compare-and-swap) label update,
+// so the reservation cannot be made truly atomic. After the update the claimant is re-read as a
+// best-effort check: it catches the common case where another machine has overwritten the claimant, but
+// two machines that each complete their update+re-read without interleaving can both observe their own
```
janiskemper: what does this mean? Which cases do we have, except the one mentioned before of one machine overwriting the label of the other? According to Hetzner we always overwrite the previous set of labels (https://docs.hetzner.cloud/reference/cloud#tag/servers/update_server.body.labels)

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Server recycling
+metatitle: Reusing existing Hetzner Cloud servers instead of creating new ones
+sidebar: Server recycling
+description: Back a MachineDeployment with a fixed set of pre-existing servers that are rebuilt and reused across scale events instead of being created and destroyed.
+---
+
+Server recycling lets an `HCloudMachine` reuse a pre-existing Hetzner Cloud server instead of creating a
+new one. When a machine with recycling enabled is provisioned, the controller looks for a server that is
```
janiskemper, line 9: I think it should be made clear that we are talking about Hetzner labels, not Kubernetes labels. On first reading this, I actually got confused here.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -0,0 +1,112 @@
...
+labelled as recyclable and matches the machine's server type and location, claims it, and rebuilds it
+with the machine's bootstrap data. On deletion the server is returned to the recyclable set instead of
+being deleted.
+
+This is useful when a fixed set of servers should back a `MachineDeployment` — for example servers on a
+reserved or otherwise favourable plan — so that scaling up and down reuses those servers rather than
+provisioning and destroying a fresh server on every scale event. Because a recycled server is a fully
```
janiskemper, line 19: So what exactly happens if a MachineHealthCheck remediates it? I would say right now that it would be returned to the pool of servers. However, this might not be the right strategy. With bare metal, this problem is obvious and cannot be solved without also manual intervention. What should be the strategy here? 

We can of course write that users have to be careful because they might run into a remediate-and-recycle loop with a broken server. Or we have some "elaborate" strategies (e.g. that in such cases the server is always thrown away and not kept, or whatever.

janiskemper (commented):
looked at it from high-level!

pkg/services/hcloud/server/recycle.go
Rene-Roscher: Right, the only race is one machine overwriting another's claim. The re-read catches
the interleaved case (loser sees the other claimant and backs off). The one it can't
catch is two non-interleaved claims, which concurrency=1 prevents. Tightened the comment
to say that.

docs/caph/02-topics/06-advanced/05-server-recycling.md
Rene-Roscher: Good catch, clarified that these are Hetzner Cloud labels, not Kubernetes ones

docs/caph/02-topics/06-advanced/05-server-recycling.md
Rene-Roscher: Documented as a limitation for now: a remediated server goes back to the pool and can
loop on a broken one, mitigation is removing the caph-recycle label. Happy to add
auto-quarantine as a follow-up if you prefer.

Rene-Roscher:
Pushed #3cf72b01 - Addressed all three: labels clarified as Hetzner Cloud not Kubernetes,
imageURL restriction reworded as a snapshot-only scope for now, and remediation documented
as a limitation. No functional changes. Ready for review whenever your release settles

pkg/services/hcloud/server/recycle.go
janiskemper: I'm not sure if this has to be the case. I would suggest something like the following:
1. A writes, then requeues for 1 second to prevent it from reading immediately, while something else can still write
2. B writes
3. A reads after 1 second and realizes B's write
4. B reads and realizes it's the owner
5. C goes nuts and writes even though there is no reason ;) 
6. B realizes while provisioning the server that it is not the owner anymore and marks the whole thing as unhealthy and triggers a remediation, which will delete the ownership labels
7. C realizes also that something is not right after its ownership label is gone and does the same thing.

Basically, there are two important aspects: 
a) a short waiting time that should prevent in 99% of the cases that a second write can collide with a (now delayed) read, and
b) that we still check the labels even after the ownership is "clear" and remediate the machine if something went wrong.

Both together should make it possible IMO to also use concurrency with this feature.

Rene-Roscher: Both parts are in, and concurrency works with the ownership mechanism.

The important correction first: I initially thought (b) wasn't needed because the existing server lookup would notice a takeover. That's wrong. findServer resolves by ProviderID and returns as soon as it finds the server; it never checks the labels. Without the ownership check, a machine could therefore continue provisioning a server that another machine had already claimed. On deletion, it could even shut that server down and return it to the pool while the new owner is using it.

So we now have both safeguards you suggested:

The claim is written and the reconcile requeues for 1s instead of immediately reading it back.
The next reconcile reads the labels again and only proceeds if we are still the claimant.
Every reconcile that touches a live server also verifies the ownership/identity label. If ownership was lost, the machine is marked for remediation via SetErrorAndRemediate.
The delete path also verifies ownership before returning a server to the pool, so a machine that lost ownership cannot accidentally recycle someone else's server.

I also tested your A → B → C scenario against the real API. A writes, B overwrites it, A backs off, B proceeds. If C later overwrites B, B detects the lost ownership and is remediated without returning the server to the pool. C then fails its rebuild because the server is still locked by B's recent action, releases the claim, and the server ends up back in the pool. I ran this three times with the same result.

And this works with controller concurrency as well. With --hcloudmachine-concurrency=3, 6 machines competing for 3 recycled servers resulted in 3 recycled + 3 newly created servers, with 6 distinct servers and no server being provisioned by two machines. So concurrency=1 is no longer required for the ownership mechanism.

Some additional details / nice to know

The 1s delay isn't intended as mutual exclusion. I measured the actual API write/read visibility over 200 cycles:

UpdateServer     p50 56ms   p99 79ms   max 157ms
write->visible   p50 102ms  p99 130ms  max 202ms

There were 0 read-after-write misses and 0 contended writes that failed to settle on one winner over 400 cycles. So 1s gives roughly 5x headroom based on the observed worst case.

I also found and fixed three related edge cases in the claim lifecycle:

A late releaseReservation could previously overwrite a newer claimant's labels. It now verifies ownership before releasing.
An abandoned claim could permanently remove a server from the pool. Claims now have a timestamp, and claims older than 15 minutes are considered stale and released.
AttachToNetwork is async and its action was discarded while the rebuild ran right after, which is the same hcloud action lock as in the scenario above. The action is confirmed before rebuilding now.

For the MachineHealthCheck: with an MHC targeting the machine, the takeover was detected and the MachineSet replaced the machine in about 30 seconds. Without an MHC, the machine was annotated but nothing happened after more than 8 minutes, and CAPH stopped reconciling it. So an MHC is effectively a prerequisite for the remediation path to actually replace the machine.

One unrelated thing I noticed while testing: the PR also handles placement groups when recycling control-plane servers. A recycled server isn't moved between groups automatically, so we now skip candidates in the wrong group and add an ungrouped candidate before rebuilding. I'd be happy to split that into a separate PR if you'd rather keep this PR focused.

The remaining edge case there is that if the OS ignores the ACPI shutdown, the server never reaches off and therefore can't be moved into the placement group. The claim keeps retrying, so this is visible rather than silently wrong, but it would need a hard power-off operation to fully solve.

One question regarding the remediation behaviour: is relying on a MachineHealthCheck the intended CAPI/provider boundary here, or would you prefer the provider to take some more direct action when a machine loses ownership?

janiskemper: Sorry I realized that I missed coming back to you about this. I had in mind that you had to respond to me, not vice versa ;)

> One question regarding the remediation behaviour: is relying on a MachineHealthCheck the intended CAPI/provider boundary here, or would you prefer the provider to take some more direct action when a machine loses ownership?

I'm not sure whether I understand this question 100%. I'll answer in a generic way and then you can see whether this is an appropriate answer or not: CAPH never deletes a machine. It can trigger remediation processes of CAPI, e.g. by annotating a CAPI Machine object, but it never takes direct action to delete something.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -48,23 +54,41 @@ A server joins the recyclable set by carrying two labels:
...
-3. **Reserves** the candidate by dropping the `caph-recycle-available` marker and adding a transient
-   `caph-recycle-claimant` label naming the machine, then re-reads it to confirm the reservation was not
-   lost to another machine that wrote at the same time.
-4. **Rebuilds** the reserved server with the machine's image and bootstrap data.
-5. **Finalizes** the claim by giving the server the machine's name and labels and dropping the claimant
-   marker.
+1. Lists servers carrying `caph-recycle=true`.
+2. Filters them down to those that are available and whose server type and location match the machine,
+   and picks the candidate with the lowest ID first, so concurrent claims from different machines
+   contend deterministically.
```
janiskemper, line 76: wouldn't it be, on the contrary, beneficial to take random machines so that it is not as likely that they would fight?

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -48,23 +54,41 @@ A server joins the recyclable set by carrying two labels:
...
+3. **Claims** the candidate by dropping the `caph-recycle-available` marker and adding a transient
+   `caph-recycle-claimant` label naming the machine, plus `caph-recycle-claimed-at` with the time of the
+   write. It then **stops and requeues** without reading anything back.
```
janiskemper, line 79: >  without reading anything back

what does this mean? Seems like unnecessary AI text

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -48,23 +54,41 @@ A server joins the recyclable set by carrying two labels:
...
+| `caph-recycle-claimant` | machine name | The machine currently claiming the server. |
+| `caph-recycle-claimed-at` | Unix seconds | When the claim was written, so an abandoned claim can be aged out. |
+
+To hand an existing server to CAPH, set both of the first two labels on it (for example with `hcloud server add-label`),
 making sure its server type and location match the machines that should claim it. Any server — including
 one that was originally created by hand — becomes schedulable this way, without rebuilding it up front.
 
 ## How a claim works
 
 When a machine with recycling enabled is created, the controller:
```
janiskemper, line 71: let's be specific here. We talk about HCloudMachines. "machine" doesn't mean anything but indicates that "Machine" (from Cluster API) is meant, which is wrong.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -48,23 +54,41 @@ A server joins the recyclable set by carrying two labels:
...
+5. **Rebuilds** the claimed server with the machine's image and bootstrap data, after re-checking
+   ownership one last time immediately before the rebuild and, if the template names a placement group
+   the server is not yet in, adding it to that group.
+6. **Finalizes** the claim by giving the server the machine's name and labels and dropping the claimant
+   and timestamp markers.
+
+From then on ownership is expressed by the machine identity label. The controller resolves its server by
+ProviderID, which says nothing about who owns it, so that label is re-checked on every reconcile that
+touches a live server: a machine whose server has been taken over is handed to remediation rather than
+provisioning hardware that is not its own, and it will not return that server to the pool on deletion.
```
janiskemper, line 91: Which case do you refer to exactly here? I'm not sure. Is it about a running HCloudMachine that has a server that suddenly got a "claimant" label on Hetzner? Is it a case where this HCloudMachine is still provisioning and then suddenly someone overwrites the HCloudMachine's own claimant label?

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
-- Matching is by server type and location only. A machine never claims a server that does not match its
-  template, but it does not otherwise distinguish between candidates beyond picking the lowest ID.
+- Matching is by server type, location and placement group. Beyond that a machine does not distinguish
+  between candidates: it takes the lowest ID.
+- **Control planes: mind the placement group.** A rebuild does not move a server between placement
+  groups, and hcloud offers no way to take a server out of one. A candidate already sitting in a
+  *different* group is therefore skipped, and one in no group is added to the requested group before the
+  rebuild. This matters most for control planes, which rely on a spread placement group to sit on
+  separate physical hosts: a control plane silently provisioned outside that group turns one host
+  failure into a lost etcd quorum. If a pool is meant for control planes, either leave its servers
```
janiskemper, line 122: why would a server outside of a placement group lead to an etcd failure? The placement group doesn't have any network restrictions etc that I'm aware of.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
+  without a placement group or put them in the one the template names.
+
+  hcloud only moves a **stopped** server into a placement group, and it does so asynchronously. A server
+  that was just returned to the pool may still be shutting down — returning it issues an ACPI request
+  and re-pools it immediately — so the claim asks it to stop and retries on the next reconcile. A server
+  whose OS ignores ACPI never reaches that state and its claim keeps retrying; the machine stays
+  unprovisioned and visible rather than failing silently, but it needs a hand.
```
janiskemper, line 129: this part is not very understandable. Try to use easier language. Ensure that the AI doesn't introduce unnecessarily complex language in general plz.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
 - The size of the recyclable set is not managed by CAPH. If more machines are created than there are
   available recyclable servers, the surplus machines create new servers normally.
-- **Claiming is safe only while machine reconciles are serialized.** A server is reserved with an
-  optimistic label write plus a re-read, but Hetzner has no atomic (compare-and-swap) label update, so
-  two machines reconciling at the same time can, in a narrow window, both believe they reserved the same
-  server. This does not happen with the default `--hcloudmachine-concurrency=1` (reconciles run one at a
-  time); raising that flag while using recycling can cause two machines to claim one server. Keep
-  `--hcloudmachine-concurrency=1` when recycling is in use.
+- **A claim is not a lock.** Hetzner replaces a server's whole label set on update and has no atomic
```
janiskemper, line 132: this whole point doesn't seem like a limitation to me. It's just something the code has to deal with and it does that already. A limitation should be something we cannot handle.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
+  (compare-and-swap) label write, so claiming is optimistic: whoever writes last wins. Splitting the
+  write from the verification across two reconciles makes a collision unlikely — a competing write has
+  time to land and be observed — but two claims further apart than that delay still each read back their
+  own name. Safety therefore does not rest on the claim being exclusive; it rests on ownership being
+  re-checked immediately before the rebuild (the only destructive step), on a machine that has lost its
+  claim backing off instead of touching labels it no longer owns, and on the losing machine simply
+  creating a new server. The worst outcome of a lost race is a wasted reconcile, not a shared server.
+- **Losing a server needs a `MachineHealthCheck` to be recoverable.** When a machine detects that its
```
janiskemper, line 140: this should be cleared: In case this result is totally unexpected, we can just remediate the respective Machine. If this is expected, we can think about dealing with it. 

In either case, I wouldn't say this is a limitation.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
   Node bootstrap does not depend on injected SSH keys (RKE2 joins via cloud-init), but you cannot rely on
   template SSH keys being present on a recycled node.
 - A returned server keeps the deleted machine's name until it is claimed again (only its labels are
   reset). Parked pool servers therefore show a stale machine name in the Hetzner console.
+- **A claim costs six hcloud API calls**, and more when the template asks for extras: two pool listings
+  (one per reconcile), the claim write, one ownership re-check before the rebuild, the rebuild, and the
+  finalize. Attaching a private network adds two, a placement group adds two more. Hetzner rate-limits
+  per project, so a large fleet of recycling machines reconciling at once uses a noticeable share of the
+  budget. The count is pinned by a unit test so it cannot grow unnoticed.
+- **The 15-minute claim TTL is compared against the controller's own clock.** A claim carries the wall
```
janiskemper, line 169: I think this is extremely unlikely to be relevant. Please re-check and remove it if you agree it won't be.

docs/caph/02-topics/06-advanced/05-server-recycling.md
```diff
@@ -88,25 +112,73 @@ never destroyed even if recycling has since been disabled on the machine.
...
+  time of the controller that wrote it, and whichever controller later ages it out uses its own clock to
+  do so. After a leader change that is a different process; if its clock runs far ahead it can reclaim a
+  claim that is still live. Ordinary NTP-synced hosts are nowhere near that margin, and the ownership
+  check before the rebuild catches the consequence, but the assumption is worth knowing.
+- An abandoned claim is reclaimed after 15 minutes, not immediately. A claim is written and verified
+  within seconds, so a controller that dies in between would otherwise pin its server outside the pool
+  forever: a claimed server carries no `caph-recycle-available` marker and matches no lookup. Any machine
+  that finds a claim older than that returns the server to the pool. Claims written before
+  `caph-recycle-claimed-at` existed carry no timestamp, cannot be aged, and are logged instead of
+  reclaimed — relabel those by hand.
```
janiskemper, line 179: I don't understand this part. I guess that it is not a limitation though but rather something to be documented above on how the controller behaves.

janiskemper (commented):
Thanks for the updates.

I gave another review of the docs.

I want to highlight that we align with the Kubernetes guidelines on AI usage, which basically say that the human still has to own 100% of the code and docs. I'm not saying that you don't do it, but you use AI very heavily so that I want to point out that you please always double check everything the AI does.

PR #2217, opened 2026-07-31
🐛 Check image-url-command timeout before reading over SSH

### LINKED ISSUE #2215: hcloud: RunningImageCommand never times out when SSH to the machine breaks
## Summary

CAPH can provision hcloud machines with a custom image command, driven by a boot-state machine (added in PR #1647). One of its states, `HCloudBootStateRunningImageCommand`, deletes a machine that stays in it too long, but that timeout never fires when SSH to the machine breaks, so the machine never leaves the state. Moving the timeout check ahead of the SSH read fixes it.

## The Problem

Each reconcile of an `HCloudMachine` first fetches the live server from the hcloud API, then dispatches to a handler based on the machine's `BootState`. For `HCloudBootStateRunningImageCommand`, `handleBootStateRunningImageCommand` runs three steps in this order:

1. `getSSHClient` reads the SSH private key from the referenced secret and builds a client for the machine's IP. It does not open a connection or touch the machine.
2. `StateOfImageURLCommand` connects to the machine over SSH and reads the current state of the image command.
3. The timeout check computes `time.Since(hm.Status.BootStateSince.Time)`; if the machine has been in this state too long, it calls `SetErrorAndRemediate` to delete the machine.

The timeout check sits after the SSH read it is meant to protect against. When SSH to the machine stops responding, because the machine hung or the network to it is down, `StateOfImageURLCommand` returns an error on every reconcile. The handler returns that error before it reaches the timeout check, and the controller requeues and tries again. Time keeps passing since `BootStateSince`, but nothing ever compares it, so the machine stays in `HCloudBootStateRunningImageCommand` with nothing to end it.

## The Solution

Move the timeout check to the top of `handleBootStateRunningImageCommand`, before `getSSHClient` and `StateOfImageURLCommand`, so it runs the same way as every other boot-state handler. The handler then decides to delete a long-stuck machine from `BootStateSince` alone, which never depends on the machine answering over SSH.

abdullah599:
In `handleBootStateRunningImageCommand` the 20-minute timeout was checked only after `getSSHClient` and `StateOfImageURLCommand`. When a machine stops answering over SSH, those calls fail on every reconcile, so the timeout check was never reached and the machine stayed in `RunningImageCommand` forever. This moves the timeout check above the SSH read, so a stuck machine is remediated from `BootStateSince` alone.

Fixes #2215

The timeout's error log dropped its `logFile` argument, since that value is produced by the SSH read that now runs later.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [x] add unit tests


pkg/services/hcloud/server/server.go
```diff
@@ -948,16 +948,9 @@ func (s *Service) handleBootStateBootingToRescue(ctx context.Context) (reconcile
...
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("getSSHClient failed (wait for image-url-command): %w", err)
-	}
-
-	state, logFile, err := hcloudSSHClient.StateOfImageURLCommand(ctx)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("StateOfImageURLCommand failed: %w", err)
-	}
-
+	// A stuck machine stops answering over SSH, which makes StateOfImageURLCommand fail on
```
janiskemper: let's remove this comment. IMO that's obvious and was wrong before

janiskemper (approved):

guettli:
@abdullah599 Janis approved that. Is the PR ready to be merged?

abdullah599:
Yes it is @guettli

PR #2223, opened 2026-08-03
✨ Add a configurable health check for the control-plane load balancer (port #2204 to main)

guettli:
TODO: is that PR needed? there is

[✨ feat: api-lb http-healthcheck by simonostendorf · Pull Request #2193 · syself/cluster-api-provider-hetzner](https://github.com/syself/cluster-api-provider-hetzner/pull/2193)

## Summary

Port of #2204 (merged into `release-1.1`) to `main`.

Adds an optional health check for the control-plane load balancer, configurable per cluster via `spec.controlPlaneLoadBalancer.healthCheck`.

## Why

The control-plane load balancer uses a plain TCP check on the API server service port. A target counts as healthy as soon as the port accepts a connection, which happens before the API server behind it is ready to serve. During a rollout the load balancer can send API traffic to a control plane whose API server is still starting.

An HTTP or HTTPS check against a readiness path (for example `/readyz`) lets the load balancer route only to a target whose API server actually answers, and drop one that stops answering.

## What changed

- New optional `healthCheck` on the control-plane `LoadBalancerSpec`: `protocol` (tcp, http, https; default tcp), `path`, `domain`, `response`, `port`, `statusCodes`, `intervalSeconds`, `timeoutSeconds`, `retries`.
- Applied when the kube-apiserver service is created, and reconciled in place when the spec changes. Only the kube-apiserver service is affected; extra services keep their behavior.
- **Opt-in and non-breaking**: when `healthCheck` is unset the service keeps the load balancer's default TCP check, so existing clusters are unchanged.

## Safe migration from TCP to HTTP

Switching a live cluster from the default TCP check to an HTTP or HTTPS check can mark every backend unhealthy at once if the backends do not answer the path yet, which takes the API server offline. So the switch is gated the same way as `EnableProxyProtocol`:

- CAPH switches in place only after every control-plane infrastructure machine carries `capi.syself.com/http-health-check-for-controlplane-loadbalancer: "true"`, set on the control-plane machine template.
- A machine from an earlier template does not carry it, so the switch happens only once the whole control-plane rollout has replaced every machine with one whose image answers the path. Until then the TCP check stays in place.
- A new cluster gets the configured check from the start (no migration).

## Differences from #2204

`main` has the `v1beta1`/`v1beta2` API split that `release-1.1` does not, so the port is not a plain cherry-pick:

- The `HealthCheck` field and `LoadBalancerHealthCheckSpec` type were added to **both** `api/v1beta1` and `api/v1beta2`. `v1beta2` is the hub, so the reconcile code uses `infrav2`. `zz_generated.conversion.go` was regenerated — the conversion is lossless in both directions and needed no manual override.
- `HTTPHealthCheckForControlPlaneLoadBalancerAnnotation` lives in `api/v1beta2` only, next to the existing `ProxyProtocolForControlPlaneLoadBalancerAnnotation`.
- The webhook validation (and its test) was added to both `pkg/webhook/v1beta1` and `pkg/webhook/v1beta2`, adapted to `main`'s free-function style (`validateLoadBalancerHealthCheck(r)` rather than a method).
- `createOptsFromSpec` on `main` returns a single value and `ControlPlaneEndpoint` is a value type, so the upstream `ErrControlPlaneEndpointNotSet` test does not apply; `main`'s "zero listen port" test is kept as-is.
- CRDs, deepcopy and conversions regenerated; `make generate` leaves the tree clean.

The health-check logic itself is line-for-line identical to #2204 (modulo the `infrav1` → `infrav2` import rename).

## Testing

- `make test-unit`: 316 tests pass, 1 skipped (unrelated real-server SSH test).
- `make lint-golang`: 0 issues.
- Unit tests for the health-check option builders, the drift comparison, the migration decision, and the annotation gate (including that the proxy-protocol annotation does not satisfy the HTTP gate).
- Reconcile tests (with the fake hcloud client extended to model a service's health check) that prove the switch happens only once every control-plane machine is annotated, and requeues otherwise.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


janiskemper:
@guettli let's take #2193. You can review that one instead (should be no diff)

guettli:
closing, because we focus on #2193

PR #2226, opened 2026-08-04
🐛 Fix stale HetznerCluster.Status causing cp targets to be skipped on LB attach (port #2107 to main)

abdullah599:
Forward-ports PR #2107 from release-1.1. reconcileLoadBalancerAttachment decided "already a load-balancer target" from HetznerCluster.Status, which only the HetznerCluster controller updates. During a rollout a target can be removed and re-added with the same ID while the status is stale, so the attach was skipped and the control-plane node was never added back.

Until a machine's ServerAvailableCondition is True, the HCloud and bare-metal services now fetch the live load balancer from the Hetzner API instead of trusting the status, then fall back to the status cache once confirmed. The HetznerCluster controller also watches control-plane HCloudMachines and HetznerBareMetalMachines and reconciles when a machine's ServerAvailableCondition transitions to True or the machine is deleted, keeping the status target list fresh.

Adapted to main: the controller is on v1beta2 (mapper funcs use the v1beta2 Cluster contract); the two HandleRateLimitExceeded call sites use the V1Beta1 variant; the bare-metal live-fetch branch composes with the address-family logic from PR #2206.

None (forward-port of PR #2107).

The watch predicate and the load-balancer gate read the v1beta1 ServerAvailable condition for now, because the HCloudMachine and HetznerBareMetalMachine machine scopes are still infrav1 (v1beta1) resources, so their v1beta1 conditions are the native ones to read. This will move to the v1beta2 ServerAvailable condition when those scopes are migrated to v1beta2.

### DIFF
--- a/controllers/hetznercluster_controller.go
+++ b/controllers/hetznercluster_controller.go
@@ -43,6 +43,7 @@ import (
 	"sigs.k8s.io/cluster-api/util/annotations"
 	conditions "sigs.k8s.io/cluster-api/util/conditions"
 	deprecatedv1beta1conditions "sigs.k8s.io/cluster-api/util/conditions/deprecated/v1beta1"
+	v1beta1conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"
 	"sigs.k8s.io/cluster-api/util/predicates"
 	"sigs.k8s.io/cluster-api/util/record"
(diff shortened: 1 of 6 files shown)


abdullah599:
- Duplicate of https://github.com/syself/cluster-api-provider-hetzner/pull/2128

PR #2232, opened 2026-08-07
:seedling: Treat an SSH dial timeout as a retry while booting into the rescue system

guettli:
## Summary

While a server boots into the rescue system, `handleBootStateBootingToRescue` polls SSH for the hostname. `ECONNREFUSED` was already handled as an expected result: set the `RetryingSSHConnection` condition and requeue. A dial timeout was not - it fell through to the fallback branch, which calls `s.scope.Error(err, "")` and sets the condition to `GetHostnameFailed`.

A dial timeout is at least as common as `ECONNREFUSED` here, and happens earlier: the network isn't up yet in the first seconds after power-on. Found while soak-testing #2219, where nearly every one of ~90 boots produced such an error log line.

## Change

Handle a dial timeout like `ECONNREFUSED`: set `RetryingSSHConnection` and requeue immediately, so the fallback branch is not reached.

The condition message now says which of the two happened, for example "Waiting for the server to become reachable over SSH while it boots into the rescue system (timeout)".

Detection uses `errors.As(err, &netErr) && netErr.Timeout()`, not `os.IsTimeout(err)`. The dial error is wrapped once with `fmt.Errorf("failed to dial ssh (%s): %w", ...)` before it reaches this check, and `os.IsTimeout` does not unwrap `%w`; it returns false on the wrapped error. `errors.As` walks the wrap chain.

`context.DeadlineExceeded` satisfies `net.Error` with `Timeout() == true` as well, so it is excluded explicitly. It means our own reconcile ran out of time, not that the server is still booting, and should keep being logged.


pkg/services/hcloud/server/server.go
```diff
@@ -816,9 +816,17 @@ func (s *Service) handleBootStateBootingToRescue(ctx context.Context) (reconcile
...
+		isDialTimeout := errors.As(err, &netErr) && netErr.Timeout()
+		isConnRefused := errors.Is(err, syscall.ECONNREFUSED)
+		if isConnRefused || isDialTimeout {
+			// Both are common while the server is still booting: ECONNREFUSED once the network is
+			// up but sshd isn't listening yet, a dial timeout before that. Provide a nice message.
+			reason := "connection refused"
+			if isDialTimeout {
+				reason = "timeout"
+			}
+			msg = fmt.Sprintf("getHostName: ssh not reachable yet (%s). Retrying", reason)
```
janiskemper: I think we discussed something more user-friendly like "Waiting for the server to become reachable over SSH while it boots into the rescue system." and a reason that fits.

janiskemper (commented):
this PR title doesn't seem to fit to the content here. Do you wanna review this already, or do you work on different changes still?

guettli:
> this PR title doesn't seem to fit to the content here. Do you wanna review this already, or do you work on different changes still?

Yes, it's ready for review. The diff should only be `pkg/services/hcloud/server/server.go` and its test. 

Why do you think the title does not match the changes?

pkg/services/hcloud/server/server.go
guettli: @janiskemper Done — updated the message in 0887d6d3ac0ba0024bbffd3f8eff764a048d0df9

janiskemper:
I see no change in logging with as "info" and not "error" in the code. I only see something added related to timeout errors  @guettli.

I actually think I understood it now though. Your AI wrote "log at info, not error" because previously in the context you will have chatted with the AI about logging at error, and now decided for info instead. Please keep such history from the chat with AI out of the actual finalized texts.

guettli:
> I see no change in logging with as "info" and not "error" in the code. I only see something added related to timeout errors @guettli.
> 
> I actually think I understood it now though. Your AI wrote "log at info, not error" because previously in the context you will have chatted with the AI about logging at error, and now decided for info instead. Please keep such history from the chat with AI out of the actual finalized texts.

I am sorry, the PR text was outdated. I update it, and extended the PR a bit (to not swallow timeout of the ctx). Please review again.

PR #2245, opened 2026-08-17
🐛 Skip reboot when the Node is missing in remediation

### LINKED ISSUE #2237: Bare metal and hcloud remediation: skip the reboot when the Node is missing

When a Machine's Node is gone (NodeRef set but the Node was deleted), the MachineHealthCheck marks the Machine unhealthy with reason NodeDeleted ([machinehealthcheck_targets.go#L211-L214](https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.4/internal/controllers/machinehealthcheck/machinehealthcheck_targets.go#L211-L214)) and triggers remediation. Remediation reboots the machine up to RetryLimit and waits out the timeout before it deletes and replaces it. Bare metal reboots the host through the reboot annotation ([remediation.go#L183](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/pkg/services/baremetal/remediation/remediation.go#L183)) and hcloud reboots the server through RebootServer ([remediation.go#L151](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/pkg/services/hcloud/remediation/remediation.go#L151)). A reboot cannot bring back a deleted Node, so the retry window is wasted.

**What should happen:**

When remediation sees the Node is missing, skip the reboots and mark the machine for deletion right away, so it gets replaced instead of rebooted. The Machine already reports this as NodeDeleted, so remediation can check it before rebooting. This applies to both bare metal and hcloud remediation.

Rivaldito:

When a Machine's Node is gone, a reboot cannot bring it back. But remediation still tried to reboot the host or server first, and only replaced the machine after all retries were used up. This wasted time.

This PR adds one check: if the Machine's health check says the Node is gone (reason `NodeDeleted`), remediation skips the reboot and marks the machine for deletion right away. This applies to both bare metal and hcloud remediation.

**Which issue(s) this PR fixes** 
Fixes #2237

The check reads the `HealthCheckSucceeded` condition on the Machine, which the MachineHealthCheck controller sets. Added one test for hcloud remediation and one for bare metal remediation to check that no reboot happens and the machine goes straight to deletion.

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [x] add unit tests


pkg/services/baremetal/remediation/remediation.go
```diff
@@ -71,6 +71,12 @@ func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
 			"exit remediation because hbmm has no host annotation")
 	}
 
+	// Node is gone, so a reboot won't help. Skip reboot, go straight to delete.
```
abdullah599: setOwnerRemediatedConditionToFailed doesn't delete here, it sets MachineOwnerRemediated and hands off to CAPI (the owner controller does the actual delete). can we reword 'go straight to delete' to match, like the line 143 comment? same comment in the hcloud remediation.go

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -71,6 +71,12 @@ func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
...
+	if conditions.GetReason(s.scope.Machine, clusterv1.MachineHealthCheckSucceededCondition) == clusterv1.MachineHealthCheckNodeDeletedReason {
```
abdullah599: hcloud puts this same check before its API call, but here it's after the host GET even though this branch never uses host. can we move it above GetAssociatedHost so a transient host-lookup error can't block the node-gone exit?

pkg/services/baremetal/remediation/remediation.go
```diff
@@ -71,6 +71,12 @@ func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
...
+		return reconcile.Result{}, s.setOwnerRemediatedConditionToFailed(ctx,
```
abdullah599: should the node-gone case respect OnExhaustion: Retire? handlePhaseWaiting calls retireHost for it, but this early exit always uses the reuse path. i think reuse is right here (a deleted Node isn't a hardware failure, and the other early guards do the same) just want to confirm it's intentional

pkg/services/baremetal/remediation/remediation.go
abdullah599, line 76: On bare metal, OnExhaustion lets the operator choose what happens to the host when remediation gives up. Reuse frees the host back into the pool, Retire sets a permanent error so it stays out until someone clears it, and RetireIfUnhealthyCondition (coming in PR #2235) retires only when a node condition the operator listed caused the failure. handlePhaseWaiting already applies this at the end of the normal flow.

This new check always reuses, so it ignores that setting. If an operator set Retire, a missing Node would still put the host back in the pool. Can we honor OnExhaustion here like handlePhaseWaiting does, so Retire calls retireHost and everything else reuses?

pkg/services/hcloud/remediation/remediation.go
```diff
@@ -59,6 +59,15 @@ func (s *Service) Reconcile(ctx context.Context) (reconcile.Result, error) {
 		return reconcile.Result{}, nil
 	}
 
+	// Node is gone, so a reboot won't help. Skip reboot, go straight to delete.
```
abdullah599: small wording thing: 'go straight to delete' reads like this deletes the machine, but setOwnerRemediatedConditionToFailed just sets a condition and hands off to CAPI, which does the actual delete. can we reword it, maybe 'mark the machine for deletion by CAPI'?

abdullah599 (approved):
LGTM

PR #2248, opened 2026-08-20
:bug: Stop requeuing HetznerBareMetalHost when the corresponding Node is not found

### LINKED ISSUE #2236: HBMH: stop requeuing when the corresponding Node is not found

When `actionProvisioned` looks up the Node in the workload cluster (from `machine.Status.NodeRef.Name`) and the Node is not found, we treat it like any other Get error: we mark the `NodeBootIDRetrieved` condition and return the error ([host.go#L2340-L2356](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/pkg/services/baremetal/host/host.go#L2340-L2356)). So the reconcile keeps retrying with backoff forever:

> failed to reconcile HetznerBareMetalHost xxx: action "provisioned" failed: failed to get corresponding Node object from the workload cluster: nodes "bm-xxx" not found

This lookup runs before the reboot logic, so while it retries it also blocks the reboot timeout and the phase that clears the reboot annotation. The host stays stuck reporting a reboot in progress.

A not-found Node is a definitive answer, not a transient error. We should mark the condition accordingly (False, not Unknown) and stop requeuing instead of erroring forever.

opplieam:

### What this PR does
Adds a `NotFound` branch to the workload-cluster Node lookup in `actionProvisioned`.

When the Node is not found, the `NodeBootIDRetrieved` condition is now reported as `False`
with a `NodeNotFound` reason on both the v1beta1 and v1beta2 halves, a warning event is
emitted, and the action returns `actionStop{}` so nothing is self-scheduled.

Every other error path is unchanged. Genuinely transient failures still report `Unknown`
and still return `actionError`.

### Why we need it
A `NotFound` was treated like any other `Get` error, so the action returned `actionError`.
controller-runtime then logged at ERROR and re-enqueued with exponential backoff, forever, 
once per affected host.

But a `NotFound` is a successful query with a negative result. The network, the auth, and
the API server all worked, and the answer was that the Node does not exist. Retrying cannot
change that, so the old behaviour was permanent log noise with no path to recovery.

Fixes #2236

**TODOs**:

### DIFF
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -2340,7 +2340,7 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 	node := &corev1.Node{}
 	err = wlClient.Get(ctx, client.ObjectKey{Name: nodeName}, node)
 	if err != nil {
-		// A NotFound is a definitive answer, not a transient failure, so retrying cannot change it.
+		// The API server answered, the node just isn't there, so asking again won't help.
 		if apierrors.IsNotFound(err) {
 			msg := fmt.Sprintf("node %q not found in the workload cluster", nodeName)
 
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -34,6 +34,7 @@ import (
 	"github.com/syself/hrobot-go/models"
 	"golang.org/x/crypto/ssh"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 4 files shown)


abdullah599 (dismissed):
LGTM :)

cc: @janiskemper

pkg/services/baremetal/host/host.go
```diff
@@ -2339,6 +2340,28 @@ func (s *Service) actionProvisioned(ctx context.Context) actionResult {
 	node := &corev1.Node{}
 	err = wlClient.Get(ctx, client.ObjectKey{Name: nodeName}, node)
 	if err != nil {
+		// A NotFound is a definitive answer, not a transient failure, so retrying cannot change it.
```
janiskemper: I would appreciate if we make this comment less AI-y ;)

janiskemper (dismissed):

janiskemper (approved):

PR #2249, opened 2026-08-21
:bug: Fix Ready condition failing on newly created objects

### LINKED ISSUE #2239: v1beta2 Ready summary fails with empty condition list
**What happens today**

Every object we manage (HetznerCluster, HCloudMachine, HetznerBareMetalHost and so on) reports a top level `Ready` condition, which is how a user sees whether the object is fully up. We do not set `Ready` by hand. In CAPI v1beta2 it is a summary of smaller sub conditions, where each sub condition reports one step, for example `ServerCreated`, `ServerProvisioned` and `ServerAvailable` on an HCloudMachine. `Ready` turns True once the steps it covers are True.

CAPI computes that with [`NewSummaryCondition`](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/pkg/scope/machine.go#L166), which each controller calls in its scope `Close`. We tell it what to summarize through a per resource [`SummaryOpts`](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/api/v1beta1/hcloudmachine_types.go#L274) function (this one is HCloudMachine's). It has two parts: 
- `ForConditionTypes` is the list of sub conditions that make up `Ready`
- `IgnoreTypesIfMissing` is the subset by which we tell CAPI to leave a missing one out instead of counting it as Unknown when computing `Ready`

The problem arises when every condition in `ForConditionTypes` is also in `IgnoreTypesIfMissing` and the object has none of them set yet, the summary has nothing to work with and returns `summary can't be performed when the list of conditions to be summarized is empty`. We hit this on a fresh HCloudMachine, whose [`SummaryOpts`](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/api/v1beta1/hcloudmachine_types.go#L274) lists all five conditions in both lists. The same shape exists on [HCloudRemediation](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/api/v1beta1/hcloudremediation_types.go#L141) and [HetznerBareMetalMachine](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/api/v1beta1/hetznerbaremetalmachine_types.go#L461), so they can hit it too.

**What should happen**

Audit each resource's `SummaryOpts` and keep a condition in `IgnoreTypesIfMissing` only when a missing value does not mean the object is not ready. Every other condition, the real lifecycle steps and the token, stays required.

Two kinds of condition pass that test for us. The negative polarity ones are only set when that specific thing is happening (`HCloudRateLimitExceeded`, `Deleting`, `RemediationSkipped`, `RobotRateLimitExceeded`), so a missing value is the normal healthy case. The optional feature ones only exist when the feature is turned on (`NetworkReady`, `LoadBalancerReady`), so a cluster without that feature never sets them and a missing value is expected. Everything else is a real step of bringing the object up, so a missing value does mean it is not ready yet, and it stays required.

Proposed Solution:
| Resource | Required (keep in the summary) | Ignore if missing |
|---|---|---|
| HetznerCluster | HCloudTokenAvailable, PlacementGroupsSynced, ControlPlaneEndpointSet, TargetClusterReady, TargetClusterSecretReady | HCloudRateLimitExceeded, Deleting, NetworkReady (optional feature), LoadBalancerReady (optional feature) |
| HCloudMachine | HCloudTokenAvailable, ServerCreated, ServerProvisioned, ServerAvailable | HCloudRateLimitExceeded |
| HCloudRemediation | HCloudTokenAvailable | HCloudRateLimitExceeded, RemediationSkipped |
| HetznerBareMetalHost | RobotCredentialsAvailable, SSHKeysAvailable, RootDeviceHintsValidated, ProvisionSucceeded, NodeBootIDRetrieved | ActionCompleted, RebootSucceeded, RobotRateLimitExceeded, Deleting |
| HetznerBareMetalMachine | HCloudTokenAvailable, HostAssociated, HostReady, ServerAvailable | Deleting |
| HetznerBareMetalRemediation | none, no Ready summary | none |

opplieam:
### What this PR does
Trims `IgnoreTypesIfMissing` on HCloudMachine, HCloudRemediation, HetznerBareMetalHost and HetznerBareMetalMachine so only conditions whose absence is normal stay ignored, meaning negative polarity ones and optional features. Every other condition is a real step of bringing the object up, so a missing value now counts as Unknown.

HetznerCluster already followed this rule. Only its comment is updated so all five resources describe the option the same way.

### Why we need it
All five conditions were listed in both `ForConditionTypes` and `IgnoreTypesIfMissing`, so a freshly created object with no conditions set gave `NewSummaryCondition` an empty input. It returned `summary can't be performed when the list of conditions to be summarized is empty`, the scope's `Close` returned that error.

Fixes #2239

- **HCloudMachineTemplate is not addressed.** It also has a `SummaryOpts` that ignores `HCloudTokenAvailable`. Left out because the issue's table does not cover it. Happy to include it if it should be in scope.

**TODOs**:


api/v1beta2/hcloudremediation_types.go
```diff
@@ -149,11 +149,11 @@ func HCloudRemediationSummaryOpts() []conditions.SummaryOption {
...
 			HCloudRemediationSkippedCondition,
 		},
-		// IgnoreTypesIfMissing tells the summary not to treat the absence of a listed condition as
-		// Unknown. Some reconcile paths exit before every condition has been set, and we don't want
-		// those early exits to flip Ready to Unknown.
+		// IgnoreTypesIfMissing lists conditions that may legitimately not be present on the object.
+		// A missing one is left out of the summary input rather than counted as Unknown. At least one
+		// condition in ForConditionTypes has to stay off this list, otherwise an object with none of
+		// them set leaves the summary with an empty input, which CAPI rejects.
 		conditions.IgnoreTypesIfMissing{
```
abdullah599: the skip branch in hcloudremediation_controller.go computes the summary at line 216 and returns, before getAndValidateHCloudToken at 267 and the defer at 294 that sets the token condition. so HCloudTokenAvailable is missing on that path and Ready's message keeps a permanent "Condition not yet reported". status stays False so the envtest passes. can we set the token condition in that branch too?

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -451,18 +451,13 @@ func HetznerBareMetalHostV1Beta2SummaryOpts() []v1beta2conditions.SummaryOption
 			HetznerBareMetalHostNodeBootIDRetrievedV1Beta2Condition,
 			HetznerBareMetalHostDeletingV1Beta2Condition,
 		},
-		// IgnoreTypesIfMissing tells the summary not to treat the absence of a
-		// listed condition as Unknown. Several reconcile paths exit before every
-		// condition has been set (for example, before Robot credentials are
-		// checked or before the host has been provisioned), and we don't want
-		// those early exits to flip Ready to Unknown.
+		// IgnoreTypesIfMissing lists conditions that may legitimately not be present on the object.
```
abdullah599: the old comment said which conditions could be missing and why, this one explains the CAPI rule instead. nothing here says why ActionCompleted and RebootSucceeded stay but ProvisionSucceeded goes. the comment you wrote in machine_test.go:110 is the version I mean, can we put that kind of line next to each list?

api/v1beta2/hcloudmachine_types.go
```diff
@@ -367,7 +362,7 @@ func (r *HCloudMachine) SetBootState(bootState HCloudBootState) {
 	r.Status.BootStateSince = metav1.Now()
 }
 
-//+kubebuilder:object:root=true
+// +kubebuilder:object:root=true
```
abdullah599: unrelated to the fix

api/v1beta1/hetznerbaremetalhost_types.go
abdullah599, line 454: this explains the CAPI rule but not why these four are on the list.

api/v1beta2/hcloudmachine_types.go
abdullah599, line 365: why is this in here? seems unrelated

abdullah599 (commented):
minor nits else LGTM

PR #2250, opened 2026-08-21
🐛 Record missing HCloud server type as ServerTypeNotFound instead of InternalError

### LINKED ISSUE #2238: HCloudMachineTemplate: record a missing server type in a status condition instead of returning an error

When `Status.Capacity` is not set, the HCloudMachineTemplate controller computes it by listing the hcloud server types and looking for the one named in `spec.template.spec.type`. If no server type matches, `getCapacity` returns an error ([machinetemplate.go#L88-L89](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/pkg/services/hcloud/machinetemplate/machinetemplate.go#L88-L89)):

> failed to reconcile machine template for HCloudMachineTemplate xxx: failed to get capacity: failed to find server type for cpx53

The controller treats this like any other reconcile failure. It sets the `Available` condition to False with the reason `InternalError` and returns the error ([hcloudmachinetemplate_controller.go#L241-L248](https://github.com/syself/cluster-api-provider-hetzner/blob/1c4fe745f3bf5b22b94d4dddd6c4b3b98f6770fc/controllers/hcloudmachinetemplate_controller.go#L241-L248)), so the reconcile is logged at error level and retried with exponential backoff. A missing server type is not an internal error. The type can be misspelled in the spec, or it can be a type that hcloud does not temporarily offer, and either way the operator sees only `InternalError` while the real cause repeats in the logs on every retry.

**What should happen:**

Record the missing server type in the `Available` condition with the reason `ServerTypeNotFound`, and stop returning the error so it is no longer logged as a failure or retried with backoff.

Rely on CAPH restarts to recheck the type instead of requeuing on a fixed interval. CAPH reconciles every HCloudMachineTemplate again when it restarts, and it restarts regularly (for example on upgrades and rollouts or manually), so a server type that hcloud starts offering again is picked up on the next restart. A type that was misspelled is fixed by editing the spec, which reconciles on its own. So the reconcile can set the condition and return, without polling for the type.

Rivaldito:

When spec.template.spec.type doesn't match any HCloud server type, the HCloudMachineTemplate controller treated it like an internal error: it set the Available condition to False with reason InternalError and retried with backoff. But a wrong server type isn't an internal error — it's either a typo in the spec or a type HCloud isn't offering right now.

This PR sets the Available condition with reason ServerTypeNotFound and stops retrying. A typo gets fixed by editing the spec, which reconciles on its own. A type that HCloud starts offering again gets picked up on
the next CAPH restart.

Fixes #2238


controllers/hcloudmachinetemplate_controller.go
```diff
@@ -239,12 +239,22 @@ func (r *HCloudMachineTemplateReconciler) Reconcile(ctx context.Context, req rec
...
 		v1beta2conditions.Set(machineTemplate, metav1.Condition{
 			Type:    infrav1.HCloudMachineTemplateAvailableV1Beta2Condition,
 			Status:  metav1.ConditionFalse,
-			Reason:  infrav1.InternalErrorV1Beta2Reason,
+			Reason:  reason,
 			Message: err.Error(),
 		})
+		if notFound {
+			return reconcile.Result{}, nil
+		}
```
abdullah599: Let's make this consistent with what we have. e.g.

https://github.com/syself/cluster-api-provider-hetzner/blob/sa%2F2238/controllers/hcloudmachinetemplate_controller.go#L129-L141


We can have one if block for not found in which we set condition + return reconcile.Result{}, nil instead of 2 conditional block (it is simpler that way)

pkg/services/hcloud/machinetemplate/machinetemplate.go
```diff
@@ -86,7 +89,7 @@ func (s *Service) getCapacity(ctx context.Context) (corev1.ResourceList, error)
 		capacity[corev1.ResourceMemory] = memory
 	}
 	if !foundServerType {
-		return nil, fmt.Errorf("failed to find server type for %s", s.scope.HCloudMachineTemplate.Spec.Template.Spec.Type)
+		return nil, fmt.Errorf("%w: %s", ErrServerTypeNotFound, s.scope.HCloudMachineTemplate.Spec.Template.Spec.Type)
```
abdullah599: @janiskemper small thing: you can look at this change in error. I think previous one was better.

we can add `Server type %q` after `:` instead of `%s` maybe 🤔

abdullah599 (commented):
Minor nits else LGTM 🚀

pkg/services/hcloud/machinetemplate/machinetemplate.go
janiskemper: I'm fine with that

abdullah599 (dismissed):
Nice Work! @Rivaldito :)

janiskemper (commented):
I have a point to align this with the current code base. 

Usually, we have most of the logic in the pkg/ folder, not in the controllers/ folder. In this case, we have the error handling in the controllers folder.

This violates a bit the principle to have most logic in the other folder, but, even worse, makes it hard to read. In case you want to find out how it works exactly, you would have to remember to check both folders to get a complete picture. 

We can easily change this though.

If we align the Reconcile function we have in machinetemplate.go with the other ones then we can return "reconcile.Result" as well. 

Then, in order to make the utility function nicer, we can put the API call to ListServerTypes out of the getCapacity function and then handle directly the error.

WDYT @abdullah599 @Rivaldito

abdullah599:
i agree

One thing to add: we should move the whole Available condition. If we move just that one case, Available gets set both in the service and in the controller, which is harder to follow than what we have now

abdullah599 (commented):
https://github.com/syself/cluster-api-provider-hetzner/pull/2250#discussion_r3835484348

:)

abdullah599 (dismissed):

abdullah599 (dismissed):

controllers/hcloudmachinetemplate_controller_test.go
```diff
@@ -333,6 +333,68 @@ var _ = Describe("HCloudMachineTemplateReconciler", func() {
 				}, timeout, interval).Should(BeTrue())
 			})
 		})
+
+		Context("unknown server type", func() {
```
janiskemper: IMO we should move these tests to the machinetemplate.go, or not? 

cc @abdullah599

pkg/services/hcloud/machinetemplate/machinetemplate.go
```diff
@@ -40,35 +49,65 @@ func NewService(scope *scope.HCloudMachineTemplateScope) *Service {
...
+		Type:   infrav1.HCloudMachineTemplateAvailableV1Beta2Condition,
+		Status: metav1.ConditionTrue,
+		Reason: infrav1.HCloudMachineTemplateAvailableV1Beta2Reason,
+	})
+	return reconcile.Result{}, nil
+}
+
+// getCapacity finds wantType among serverTypes and returns its CPU cores and memory as a
+// ResourceList, or ErrServerTypeNotFound if hcloud does not offer that server type.
+func getCapacity(serverTypes []*hcloud.ServerType, wantType string) (corev1.ResourceList, error) {
```
janiskemper: I think we should still improve this function. 

The specific error type is only necessary because we don't add a third value to return. 

Therefore, I would argue that we should either add this third return value, or we inline the whole function. 

WDYT @Rivaldito @abdullah599 ?

Rivaldito: Got it, refactoring!

abdullah599: I'd say we should opt for third value. This way we can also keep the error message consistent with what it was before :)

`func getCapacity(serverTypes []*hcloud.ServerType, wantType string) (capacity corev1.ResourceList, found bool, err error)`

controllers/hcloudmachinetemplate_controller_test.go
abdullah599: I agree

pkg/services/hcloud/machinetemplate/machinetemplate.go
Rivaldito: Done in the last commit. `getCapacity` now returns `(capacity corev1.ResourceList, found bool, err error)`. `ErrServerTypeNotFound` and the `errors.Is` check are gone. `Reconcile` branches on `!found` and sets the `ServerTypeNotFound` condition with the original message (`failed to find server type for <type>`).

controllers/hcloudmachinetemplate_controller_test.go
Rivaldito: Done in the last commit. Removed the `unknown server type` envtest context from the controller suite. `pkg/services/hcloud/machinetemplate/machinetemplate_test.go` now holds `TestGetCapacity` for the pure function plus `TestReconcileServerType` (service `Reconcile` with the fake HCloud client), which covers the `Available=False` / `ServerTypeNotFound` assertion the envtest did.

janiskemper (commented):
@abdullah599 can give a final review. From my side LGTM

abdullah599 (approved):
LGTM :)

PR #2252, opened 2026-08-22
🐛 Avoid reporting nil public IPv4s for private HCloud servers

miracle2k:
## Summary

For servers created with `publicNetwork.enableIPv4: false`, HCloud returns a nil `net.IP`. Its `String()` representation is "<nil>", which CAPH currently publishes as an ExternalIP. That is not a valid machine address and can be selected by Cluster API/Talos endpoint consumers.

- Skip an absent public IPv4 when populating HCloudMachine.status.addresses
- Use `IsUnspecified()` instead of `IsGlobalUnicast()` for the public IPv6 check too, so both branches ask the same question i.e "is an IP set here at all?"  instead of the IPv6 branch specifically checking for global-unicast
- Guard the IPv6 branch against a malformed (non-16-byte) address, which would otherwise panic when incrementing its last byte to derive the first usable address
- Add regression tests for private-only servers and for a malformed IPv6 address

## Test

```console
go test ./pkg/services/hcloud/server -run '^Test_statusAddresses' -count=1
```


guettli:
@miracle2k the PR allowed edits by maintainers. So I updated it, instead of creating a new PR. Please check. Are my change ok for you?

miracle2k:
definitely! thanks for looking at this.

pkg/services/hcloud/server/server.go
```diff
@@ -2321,20 +2321,24 @@ func statusAddresses(server *hcloud.Server) []clusterv1beta1.MachineAddress {
...
-				Address: ip,
+				Address: server.PublicNet.IPv4.IP.String(),
 			},
 		)
 	}
 
-	if unicastIP := server.PublicNet.IPv6.IP; unicastIP.IsGlobalUnicast() {
+	// The length check is needed for the ip[15]++ below. hcloud-go always returns a 16 byte
+	// address here, the check only makes a broken API response harmless.
+	if !server.PublicNet.IPv6.IsUnspecified() && len(server.PublicNet.IPv6.IP) == net.IPv6len {
```
janiskemper, line 2333: @guettli can you explain why you chose "IsUnspecified" over "IsGlobalUnicast"? I'm not the expert, but to me it sounds weird to apparently lower the checks on the validity of the IPv6 address. Especially if we rightfully add additional checks for the Ipv4 in this PR.

guettli: @janiskemper the move from `IsGlobalUnicast()` to `IsUnspecified()` is not strictly needed, but it
avoids confusion in the future.

It took me a while to understand why `IsGlobalUnicast()` was used here, because checking for unicast
does not really make sense at this spot. The question this code needs answered is "is an IP set here
at all?", and that is what `IsUnspecified()` says. With `IsUnspecified()` on both branches, v4 and v6
now use the same helper for the same question.

On the "lowering the checks" concern: in practice the two only differ for addresses Hetzner never
returns in this field — loopback, link-local, multicast.

If you would rather keep `IsGlobalUnicast()`, I can put it back. It is a small thing either way.

janiskemper: I'm okay with doing it this way. @miracle2k can you update the PR description according to the latest state? I think it is outdated now (the last sentence at least).

Dhairya-Arora01 (approved):
LGTM

PR #2253, opened 2026-08-22
✨ v1beta2 migration: switch the HCloudMachine controller to v1beta2

### LINKED ISSUE #2104: v1beta2 migration: switch the HCloudMachine controller to v1beta2

## Summary

Move the `HCloudMachine` controller, its `MachineScope`, and the HCloud server service from the `api/v1beta1` type to the `api/v1beta2` type that landed in #2074, and move condition handling to the CAPI v1beta2 condition packages.

## What would you like to add

This is part of the ongoing #2017 migration. #2074 changed the `HCloudMachine` v1beta2 API to its final shape, but the controller still imports `api/v1beta1` and writes conditions through the old CAPI condition packages.

This issue is the `HCloudMachine` controller pass. It switches the controller, `MachineScope`, and the HCloud server service to the v1beta2 type, and moves condition handling to the CAPI v1.13 condition packages (`util/conditions` and `util/conditions/deprecated/v1beta1`).

The controller keeps writing both the v1beta2 conditions and the deprecated v1beta1 conditions during the compatibility window, so v1beta1 readers and the `HCloudRemediation` controller keep working. The admission webhooks already validate the v1beta2 type. The storage version stays v1beta1, and the CRD migrator config is unchanged.

This issue does not redesign conditions. The condition set is #1935 and the API shape is #2059 (PR #2074). One later change to fold in: #2156 added the `status.actionIdCreateServer` field. The switch carries it and does not change the flow.

The final changes are:

| Area | Change |
|---|---|
| Imports | Point the `HCloudMachine` controller, `pkg/scope.MachineScope`, and the HCloud server service at `api/v1beta2` and reconcile the v1beta2 type. |
| Condition packages | Use `util/conditions` for the v1beta2 conditions and `util/conditions/deprecated/v1beta1` for the deprecated conditions, with the `util/patch` helper. |
| Condition constants | Use the v1beta2 constants from #2074 (no `V1Beta2` suffix) and the `V1Beta1` deprecated set. A few constants the controller writes today are still missing from `api/v1beta2/conditions_const.go`, such as the `ServerAvailable` condition. Add the missing ones so the code compiles. |
| Dual write | In `machinePatchOpts`, declare both owned sets with `patch.WithOwnedConditions` and `patch.WithOwnedV1Beta1Conditions`, so the controller writes the v1beta2 and the deprecated v1beta1 conditions. |
| Ready summary | Add `HCloudMachineSummaryOpts()` to `api/v1beta2`, returning `[]conditions.SummaryOption`, and compute the Ready summary from it. This replaces the v1beta1 `HCloudMachineV1Beta2SummaryOpts`. |
| Shared helpers | Move the controller off the v1beta1 token and rate limit helpers in `controllers/util_v1beta1.go` to the v1beta2 helpers in `controllers/util.go`. Generalize `reconcileRateLimit` and `hcloudTokenErrorResult`, which are typed to the v1beta2 `HetznerCluster` today, so they also serve the `HCloudMachine`. |
| Initialization | Set `status.initialization.provisioned` as a one time signal and stop writing `status.ready`. |
| Instance state | Read and write `status.instanceState` as the CAPH `InstanceState` type, non pointer. |
| Status fields | Use the non pointer `status.lastRemediatedAt`, and `status.addresses` as `[]clusterv1.MachineAddress`. Read and write the new `status.actionIdCreateServer` field (#2156). |

References:

* Parent issue: #2017.
* HCloudMachine v1beta2 API change: issue #2059, PR #2074.
* HCloudMachine v1beta2 conditions: PR #1935.
* HCloudMachine rescue-boot follow-up (adds `status.actionIdCreateServer`): PR #2156.
* CAPI InfrastructureMachine contract: https://cluster-api.sigs.k8s.io/developer/providers/contracts/infra-machine.

abdullah599:

This switches the HCloudMachine controller from the v1beta1 API types to the v1beta2 types.

Fixes #2104

*Package aliases.* We import v1beta1 as `infrav1` and v1beta2 as `infrav2`. Issue #2092 will flip these repo-wide to match CAPI (`infrav1beta1` for v1beta1, `infrav1` for v1beta2), so this PR keeps the current aliases and does not rename.

*Graceful shutdown before delete.* A running server is shut down before it is deleted. `handleDeleteServerStatusRunning` shuts it down while the `ServerAvailable` condition is True, then sets `ServerAvailable` to False, so the next reconcile deletes the server. The v1beta1 controller used to set the v1beta2 `ServerAvailable` condition to False at the top of `Delete()`, but never the v1beta1 one, because the shutdown gate read the v1beta1 condition. This PR reads the native (v1beta2) `ServerAvailable` in the gate, so `Delete()` can no longer set the v1beta2 condition to False there either. If it did, the gate would see False on the first reconcile and delete the running server without shutting it down.

**TODOs**:
- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta2/conditions_const.go
```diff
@@ -249,6 +249,9 @@ const (
 	// DeprecatedInstanceBootstrapReadyV1Beta1Condition reports on current status of the instance. BootstrapReady indicates the bootstrap is ready.
 	DeprecatedInstanceBootstrapReadyV1Beta1Condition clusterv1.ConditionType = "InstanceBootstrapReady"
 
+	// DeprecatedRateLimitExceededV1Beta1Condition reports whether the rate limit has been reached.
```
janiskemper: we can remove those deprecated conditions. We can do it here and elsewhere. We can also write a separate issue to remove it everywhere if we want.

pkg/scope/machine.go
```diff
@@ -132,59 +120,53 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
...
 //
-// MachineScope keeps its own fields instead of embedding ClusterScope, because the two scopes are on
-// different API versions during the migration. ClusterScope reconciles the v1beta2 HetznerCluster,
-// while the HCloudMachine controller is still on v1beta1 and reads the cluster as v1beta1, so
-// MachineScope holds its own v1beta1 HetznerCluster. Embedding ClusterScope would make that field
-// v1beta2, which cannot hold the v1beta1 cluster this scope works with.
-//
-// TODO: once the HCloudMachine controller is migrated to v1beta2, MachineScope also uses the v1beta2
-// HetznerCluster, so embed ClusterScope here again and drop the duplicated fields.
+// MachineScope keeps its own fields rather than embedding ClusterScope. The two scopes track
```
janiskemper: remove this "rather than ... " because it refers to a decision from the past

pkg/scope/machine.go
```diff
@@ -317,25 +289,22 @@ func (m *MachineScope) ServerIDFromProviderID() (int64, error) {
 	return serverID, nil
 }
 
-// SetReady sets the ready field on the machine.
+// SetReady records that the machine's infrastructure is provisioned. Provisioned is a one-time
+// signal per the CAPI infra-machine contract: once true it stays true, so a false argument is a no-op.
 func (m *MachineScope) SetReady(ready bool) {
```
janiskemper: we should maybe rename this to "SetProvisioned"

pkg/services/hcloud/server/server.go
```diff
@@ -1379,64 +1378,59 @@ func (s *Service) getLiveServer(ctx context.Context) (server *hcloud.Server, res
...
-	hcloudutil.HandleRateLimitExceededV1Beta1(hm, err, functionName)
+	hcloudutil.HandleRateLimitExceeded(hm, err, functionName)
 	return fmt.Errorf("%s: %w", errMsg, err)
 }
 
 // Delete implements delete method of server.
 func (s *Service) Delete(ctx context.Context) (reconcile.Result, error) {
 	// Set phase to deleting.
-	s.scope.HCloudMachine.Status.InstanceState = ptr.To(hcloud.ServerStatusDeleting)
-	v1beta2conditions.Set(s.scope.HCloudMachine, metav1.Condition{
```
janiskemper, line 1430: why do we remove this?

pkg/services/hcloud/server/server.go
```diff
@@ -2371,31 +2371,31 @@ func (s *Service) createLabels() map[string]string {
...
 // getSSHPrivateKey retrieves the SSH private key used for connecting to the rescue systems.
-// It reads the key from the Kubernetes secret referenced by HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.
+// It reads the key from the Kubernetes secret referenced by HetznerCluster.Spec.SSHKeys.RescueSecretRef.
 // On failure it sets SSHPrivateKeyAvailableCondition with a specific reason describing the root cause.
 func (s *Service) getSSHPrivateKey(ctx context.Context) (string, error) {
-	robotSecretName := s.scope.HetznerCluster.Spec.SSHKeys.RobotRescueSecretRef.Name
+	robotSecretName := s.scope.HetznerCluster.Spec.SSHKeys.RescueSecretRef.Name
 	if robotSecretName == "" {
-		v1beta1conditions.MarkFalse(
+		deprecatedv1beta1conditions.MarkFalse(
```
janiskemper, line 2393: it seems as if we don't set the new conditions in this case.. On first glance this seems wrong and I also see other similar cases. Can you maybe check whether we forgot something?

pkg/services/hcloud/server/server.go
abdullah599: We never set the v1beta1 ServerAvailable to False here, and we should not set the v1beta2 one here either. The shutdown step needs ServerAvailable to still be True: handleDeleteServerStatusRunning uses it to tell that a running server still has to be shut down before we delete it, and it sets ServerAvailable to False only after it has sent the shutdown.

The old code did set the v1beta2 ServerAvailable to False here. That worked only because the shutdown check still read the v1beta1 one. After the switch the check reads v1beta2, so setting it False here would make the first delete reconcile skip the shutdown and delete the server while it is still running.

api/v1beta2/conditions_const.go
abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2257

pkg/services/hcloud/server/server.go
janiskemper: wait - we are talking only about the instanceState here then right? What about the condition then? 

We (@Dhairya-Arora01 and me I believe) discussed how we manage to show users that a server is being deleted. Right now that is pretty much invisible.. We'll have to do something about it. 

We planned on using the InstanceState for it as well, because it fits. However, can we do that then to still be able to understand the state of the Hetzner API (e.g. shutdown, etc.)?

abdullah599: Discussed this with Janis, here is what we concluded.

The `ServerAvailable = False` at the top stays removed. The shutdown step needs `ServerAvailable` to still be True: `handleDeleteServerStatusRunning` reads it to know a running server has to be shut down first, and sets it False only after the shutdown is sent to Hetzner ([L2189-L2205](https://github.com/syself/cluster-api-provider-hetzner/blob/ab78e99aadba10a79d97e64862e7c6f20d45b12a/pkg/services/hcloud/server/server.go#L2189-L2205)). Setting it False here would make the first reconcile skip the shutdown and delete a running server.

For showing that a machine is being deleted we use `instanceState` (the Phase column) as planned. It was already set to `deleting` at the start of Delete, but got overwritten with the live server status right after ([L1481](https://github.com/syself/cluster-api-provider-hetzner/blob/ab78e99aadba10a79d97e64862e7c6f20d45b12a/pkg/services/hcloud/server/server.go#L1481)), so it only stayed `deleting` when there was no server left. I dropped that overwrite, so Phase now shows `deleting` on every path until the object is gone. Deletion is visible now, which it was not before.

We still see the Hetzner side through the `ServerAvailable` condition: True while the server runs, then False with reason `Deleting` once the shutdown is sent, and any problem during deletion surfaces through the conditions too (rate limit, invalid token, etc)

controllers/hetznercluster_controller_test.go
```diff
@@ -1711,36 +1711,44 @@ var _ = Describe("reconcileRateLimit", func() {
...
 
 	It("removes HCloudRateLimitExceeded condition and returns wait==false when wait time is over (v1beta2)", func() {
-		hcloudMachine := &infrav1.HCloudMachine{}
-		v1beta1conditions.MarkFalse(hcloudMachine, infrav1.HetznerAPIReachableCondition, infrav1.RateLimitExceededReason, clusterv1beta1.ConditionSeverityWarning, "")
-		conditionList := hcloudMachine.GetConditions()
+		// A still-v1beta1 object is used to exercise the v1beta1 rate-limit helper. HCloudMachineTemplate
+		// is one of the resources still on v1beta1 and carries the same staged v1beta2 conditions.
+		// TODO: remove this test once reconcileRateLimitV1Beta1 is gone (every resource on v1beta2, only
+		// reconcileRateLimit left).
+		machineTemplate := &infrav1.HCloudMachineTemplate{}
```
janiskemper: let's name it hcloudMachineTemplate. machineTemplate sounds like CAPI Machine

janiskemper (commented):
LGTM

controllers/hcloudmachine_controller.go
```diff
@@ -304,8 +299,8 @@ func (r *HCloudMachineReconciler) reconcileNormal(ctx context.Context, machineSc
 	hcloudMachine := machineScope.HCloudMachine
 
 	// If the HCloudMachine doesn't have our finalizer, add it.
-	controllerutil.AddFinalizer(machineScope.HCloudMachine, infrav1.HCloudMachineFinalizer)
-	controllerutil.RemoveFinalizer(machineScope.HCloudMachine, infrav1.DeprecatedHCloudMachineFinalizer)
+	controllerutil.AddFinalizer(machineScope.HCloudMachine, infrav2.HCloudMachineFinalizer)
+	controllerutil.RemoveFinalizer(machineScope.HCloudMachine, infrav2.DeprecatedHCloudMachineFinalizer)
```
Dhairya-Arora01, line 303: shouldn't we stop removing this now ?

pkg/scope/machine.go
```diff
@@ -317,25 +290,22 @@ func (m *MachineScope) ServerIDFromProviderID() (int64, error) {
...
 }
 
-// SetReady sets the ready field on the machine.
-func (m *MachineScope) SetReady(ready bool) {
-	m.HCloudMachine.Status.Ready = ready
+// SetProvisioned records that the machine's infrastructure is provisioned. Provisioned is a one-time
+// signal per the CAPI infra-machine contract: once true it stays true, so a false argument is a no-op.
+func (m *MachineScope) SetProvisioned(provisioned bool) {
+	if provisioned {
+		m.HCloudMachine.Status.Initialization.Provisioned = ptr.To(true)
```
Dhairya-Arora01: Do we ever set this to false ? I see both the callers passing true only ? Shouldn't we change the signature then ?

Dhairya-Arora01 (dismissed):
LGTM

controllers/hcloudmachine_controller.go
abdullah599: Agreed, but this is not only about HCloudMachine. HetznerCluster, HetznerBareMetalMachine, HetznerBareMetalHost, HCloudMachineTemplate and the secrets package all remove their old finalizer in the same way, so I would rather not drop it here alone. I suggest we create an issue for it and remove it everywhere at once

wdyt?

cc: @janiskemper

janiskemper: yes

abdullah599: - https://github.com/syself/cluster-api-provider-hetzner/issues/2261

Dhairya-Arora01 (approved):

PR #2256, opened 2026-08-24
:bug: Do not store "<nil>" as load balancer IP

guettli:
## Summary

`net.IP.String()` returns `"<nil>"` for an unset IP, and that string ended up in `status.controlPlaneLoadBalancer.ipv4` and `.ipv6`. Now an empty string gets stored instead.

This happens for a load balancer without a public interface (`public_interface: false`). caph always creates its load balancers with a public interface, so a cluster can only run into this by adopting an existing load balancer via `spec.controlPlaneLoadBalancer.name`.

The private IP (`.internalIP`) goes through the same helper now. Hetzner always reports a private IP for a load balancer which is attached to a network, so that part is only a safeguard.

`processControlPlaneEndpoint()` compared the IPv4 against the literal `"<nil>"` to decide whether the load balancer can be used as control plane endpoint. It now checks for an empty string. The comparison with `"<nil>"` is kept as a defensive check only: the load balancer status is rewritten from the Hetzner API earlier in the same reconcile, so a value written by an older release never reaches this code.

Such a load balancer can never provide a control plane endpoint, but the condition `ControlPlaneEndpointSet` said "load balancer not ready yet", which sounds like waiting would help. The message now says that the load balancer has no public IPv4.

## Why this is worth changing

This does not repair a broken cluster. caph creates its load balancers with a public interface and never changes that afterwards, so its own load balancers do have a public IPv4. A load balancer adopted by name could have the public interface disabled, but such a cluster can never get a control plane endpoint anyway.

It is still worth cleaning up: `"<nil>"` is how Go formats an unset IP, not an address. It leaks into a status field which users and scripts read, and comparing against it makes the control plane endpoint depend on an implementation detail of the standard library. Code like this also gets copied.

While being here: the fake hcloud client built its IPs as `net.IP("1.2.3.4")`, which is a seven byte slice and not a parsed address. `String()` returned `"?312e322e332e34"` for it, so the tests using the fake client stored that in the load balancer status without anybody noticing. It uses `net.ParseIP()` now.

Same class of bug as syself/cluster-api-provider-hetzner#2252, which addresses it for servers.


controllers/hetznercluster_controller.go
```diff
@@ -303,10 +303,21 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+		"so it can not be used as control plane endpoint"
+)
+
 func processControlPlaneEndpoint(hetznerCluster *infrav2.HetznerCluster) {
 	if hetznerCluster.Spec.ControlPlaneLoadBalancer.Enabled {
-		if hetznerCluster.Status.ControlPlaneLoadBalancer != nil && hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4 != "<nil>" {
-			defaultHost := hetznerCluster.Status.ControlPlaneLoadBalancer.IPv4
+		lbStatus := hetznerCluster.Status.ControlPlaneLoadBalancer
+		// A load balancer without a public interface has no public IPv4, so it can not be used as
+		// control plane endpoint. Older versions of caph stored "<nil>" in that case, current
```
janiskemper: I don't think we need to write about older versions in the comment. IMO we can remove that part.

controllers/hetznercluster_controller.go
```diff
@@ -303,10 +303,21 @@ func (r *HetznerClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	return reconcile.Result{}, nil
 }
 
+const (
+	msgLoadBalancerNotReadyYet = "enabled LoadBalancer but load balancer not ready yet"
+
+	msgLoadBalancerWithoutPublicIPv4 = "enabled LoadBalancer but the load balancer has no public IPv4, " +
```
janiskemper, line 309: I didn't really understand what we try to do here. Currently, from my understanding, in the rare case that the load balancer has no IPv4, we would print "<nil>" in the status. This PR changes this so it would be empty. 

However, we still don't have a proper handling of this case, do we? All the other parts of the code would still fail if they rely on this. 

Can you explain why this PR makes a lot of sense?

controllers/hetznercluster_controller.go
guettli: done.

controllers/hetznercluster_controller.go
guettli: The "issue" was found while working on #2252.

Here (for LoadBalancers) it is unlikely that there is no IP.

Overall, I am happy with closing and not merging the PR. It is a very unlikely edge-case.

janiskemper: Okay. So if you think that it  is really unlikely (aka it would never occur in reasonable situations and edge cases), then we should better close it to not "distract" people in the code.

guettli:
As discussed, this change is not really needed.

PR #1455, opened 2024-08-20
:seedling: Update names of finalizers.

guettli:
Update names of finalizers.

old: foo.infrastructure.cluster.x-k8s.io
new: infrastructure.cluster.x-k8s.io/foo

Related: https://github.com/syself/cluster-api-provider-hetzner/issues/1450

PR #2262, opened 2026-09-01
🌱 Stop removing deprecated finalizers and drop the dead BareMetalHost deleting action

### LINKED ISSUE #2261: Stop removing the deprecated finalizers

CAPH renamed its finalizers in PR #1455, which shipped in v1.0.0 in October 2024. The HCloudMachine finalizer, for example, went from `hcloudmachine.infrastructure.cluster.x-k8s.io` to `infrastructure.cluster.x-k8s.io/hcloudmachine`. Objects created before the rename still carried the old name, so each controller removes it right where it adds the new one, and each API package keeps a `Deprecated<Kind>Finalizer` constant holding the old string.

This still runs on every reconcile, for five kinds and the secrets package:

- HCloudMachine: [L298](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hcloudmachine_controller.go#L298), [L308](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hcloudmachine_controller.go#L308)
- HetznerCluster: [L220](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznercluster_controller.go#L220), [L464](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznercluster_controller.go#L464)
- HetznerBareMetalMachine: [L220](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznerbaremetalmachine_controller.go#L220), [L228](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznerbaremetalmachine_controller.go#L228)
- HetznerBareMetalHost: [L215](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznerbaremetalhost_controller.go#L215), [L366](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hetznerbaremetalhost_controller.go#L366), [host.go#L2710](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/pkg/services/baremetal/host/host.go#L2710)
- HCloudMachineTemplate: [L113](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hcloudmachinetemplate_controller.go#L113)
- Secrets: [L119](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/pkg/secrets/secret.go#L119), [L214](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/pkg/secrets/secret.go#L214)

The HCloudMachineTemplate controller already carries a note about it ([L111-L113](https://github.com/syself/cluster-api-provider-hetzner/blob/e008f14d38f7b1044cdbf900dcff2f60fdfe239b/controllers/hcloudmachinetemplate_controller.go#L111-L113)):

```go
// removing finalizer that was set in previous versions but is not needed
// We can remove that code in 2025.
controllerutil.RemoveFinalizer(machineTemplate, infrav1.DeprecatedHCloudMachineFinalizer)
```

**What should happen:**

Drop the `RemoveFinalizer` calls for the deprecated finalizers in all the places listed above, together with the `Deprecated*Finalizer` constants in `api/v1beta1`, `api/v1beta2` and `pkg/secrets`.

Rivaldito:

This PR does two small cleanups.

First, it stops removing the deprecated finalizers. We renamed them in #1455, which shipped in v1.0.0 back in October 2024. Any object made since then already uses the new names, so stripping the old ones on every reconcile does nothing. This drops those `RemoveFinalizer` calls across the controllers and the secrets package, plus the `Deprecated*Finalizer` constants in `api/v1beta1`, `api/v1beta2` and `pkg/secrets`.

Second, it removes the bare metal host deleting action, which never ran. When a host is in `StateDeleting`, `reconcileSelectedStates` in the controller removes the finalizer and returns a non-empty result, so `Reconcile` returns before the host state machine even starts. And if it did get there, `ReconcileState` calls `checkInitiateDelete()` first, and that returns `actionComplete{}` for any host with a `DeletionTimestamp`, before it looks at the state handlers. So `actionDeleting`, its `handleDeleting` wrapper, the `StateDeleting` handler entry and the `deleteComplete` result were all dead. This also fixes a small comment typo.

Fixes #2261

No behavior change. Finalizer removal on delete still happens in `reconcileSelectedStates`, same as before.

**TODOs**:

- [x] squash commits
- [ ] include documentation
- [ ] add unit tests

### DIFF
--- a/controllers/hetznerbaremetalhost_controller.go
+++ b/controllers/hetznerbaremetalhost_controller.go
@@ -210,9 +210,7 @@ func (r *HetznerBareMetalHostReconciler) Reconcile(ctx context.Context, req ctrl
 	}()
 
 	// Add a finalizer to newly created objects.
-	if bmHost.DeletionTimestamp.IsZero() &&
-		(controllerutil.AddFinalizer(bmHost, infrav1.HetznerBareMetalHostFinalizer) ||
-			controllerutil.RemoveFinalizer(bmHost, infrav1.DeprecatedBareMetalHostFinalizer)) {
+	if bmHost.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(bmHost, infrav1.HetznerBareMetalHostFinalizer) {
 		return ctrl.Result{Requeue: true}, nil
 	}
 
@@ -363,7 +361,6 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(bmHost *infrav1
 		})
 		// remove finalizers.
 		controllerutil.RemoveFinalizer(bmHost, infrav1.HetznerBareMetalHostFinalizer)
-		controllerutil.RemoveFinalizer(bmHost, infrav1.DeprecatedBareMetalHostFinalizer)
 		return reconcile.Result{Requeue: true}
 	}
 	return ctrl.Result{}
--- a/pkg/services/baremetal/host/host.go
+++ b/pkg/services/baremetal/host/host.go
@@ -2707,7 +2707,6 @@ func (s *Service) actionDeprovisioning(ctx context.Context) actionResult {
 
 func (s *Service) actionDeleting(_ context.Context) actionResult {
 	controllerutil.RemoveFinalizer(s.scope.HetznerBareMetalHost, infrav1.HetznerBareMetalHostFinalizer)
-	controllerutil.RemoveFinalizer(s.scope.HetznerBareMetalHost, infrav1.DeprecatedBareMetalHostFinalizer)
 	return deleteComplete{}
 }
 
--- a/api/v1beta1/hcloudmachine_types.go
+++ b/api/v1beta1/hcloudmachine_types.go
@@ -27,9 +27,6 @@ const (
 	// resources associated with HCloudMachine before removing it from the
 	// apiserver.
 	HCloudMachineFinalizer = "infrastructure.cluster.x-k8s.io/hcloudmachine"
-	// DeprecatedHCloudMachineFinalizer contains the old string.
-	// The controller will automatically update to the new string.
-	DeprecatedHCloudMachineFinalizer = "hcloudmachine.infrastructure.cluster.x-k8s.io"
 )
 
 // HCloudMachineSpec defines the desired state of HCloudMachine.
--- a/api/v1beta1/hetznerbaremetalhost_types.go
+++ b/api/v1beta1/hetznerbaremetalhost_types.go
@@ -37,10 +37,6 @@ const (
 	// deprovisioned.
 	HetznerBareMetalHostFinalizer = "infrastructure.cluster.x-k8s.io/hetznerbaremetalhost"
 
-	// DeprecatedBareMetalHostFinalizer contains the old string.
-	// The controller will automatically update to the new string.
-	DeprecatedBareMetalHostFinalizer = "hetznerbaremetalhost.infrastructure.cluster.x-k8s.io"
-
 	// HostAnnotation is the key for an annotation that should go on a HetznerBareMetalMachine to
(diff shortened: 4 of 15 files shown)


pkg/services/baremetal/host/host.go
```diff
@@ -2707,7 +2707,6 @@ func (s *Service) actionDeprovisioning(ctx context.Context) actionResult {
 
 func (s *Service) actionDeleting(_ context.Context) actionResult {
 	controllerutil.RemoveFinalizer(s.scope.HetznerBareMetalHost, infrav1.HetznerBareMetalHostFinalizer)
```
abdullah599, line 2709: just noticed actionDeleting never runs. reconcileSelectedStates handles StateDeleting and returns early, so ReconcileState never runs when the state is deleting.

@Rivaldito can you check what we should do here please?

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -363,7 +361,6 @@ func (r *HetznerBareMetalHostReconciler) reconcileSelectedStates(bmHost *infrav1
 		})
 		// remove finalizers.
```
abdullah599, line 362: `// remove finalizer.`

pkg/services/baremetal/host/host.go
Rivaldito: Checked it. It is dead. The controller handles StateDeleting in reconcileSelectedStates and returns early, so r.reconcile and the state machine never run for that state. And checkInitiateDelete sets StateDeleting and returns actionComplete before the handlers are reached anyway, so handleDeleting/actionDeleting can never be hit.

Removed the function, its handler, the map entry and the now-unused deleteComplete result. The finalizer is still removed in reconcileSelectedStates, which is the path that actually runs.

controllers/hetznerbaremetalhost_controller.go
Rivaldito: Fixed.

abdullah599 (commented):
LGTM

cc: @janiskemper

janiskemper:
I see we remove the delete action. It doesn't say anything about it though in the title and description. Can you add that and the reasoning?

Rivaldito:
> I see we remove the delete action. It doesn't say anything about it though in the title and description. Can you add that and the reasoning?

 Okay, in short, is becasue the `actionDeleting` (plus `handleDeleting`, the `StateDeleting` handler entry, and the `deleteComplete` result) was dead code. Once a host is in `StateDeleting`, `reconcileSelectedStates` removes the finalizer and returns a non-empty result, so `Reconcile` returns before the host state machine ever runs. And even if it got there, `ReconcileState` calls `checkInitiateDelete()` first, which returns `actionComplete{}` for any host with a `DeletionTimestamp` before the state handlers are consulted. So the deleting handler could never fire.

No behavior change; finalizer removal on delete still happens in `reconcileSelectedStates` exactly as before.

janiskemper (approved):

PR #2267, opened 2026-09-10
:seedling: Switch to mgr.GetEventRecorderFor to get the event recorder

### LINKED ISSUE #2264: Migrate to `mgr.GetEventRecorderFor` to get the event recorder
**Context**:
As per the CAPI [migration guide for v1.13 to v1.14](https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.13-to-v1.14.html#deprecation), the `util/record` pkg has been deprecated. The recommended replacement is controller-runtime's `mgr.GetEventRecorderFor`.

**Proposal**:
Use the controller-runtime utility `mgr.GetEventRecorderFor` to get the event recorder. The `EventRecorder` type it returns does not provide `Warn`/`Warnf` as separate methods, instead there is only:

```go
Event(object runtime.Object, eventtype, reason, message string)

// Eventf is just like Event, but with Sprintf for the message field.
Eventf(object runtime.Object, eventtype, reason, messageFmt string, args ...interface{})
```

Migration example:
```go
// Example
// old
record.Eventf(m, "PermanentErrorWasRemoved", "The permanent error was removed, because the annotation %q was removed", infrav1.PermanentErrorAnnotation)

// new
r.EventRecorder.Eventf(
					m,
					corev1.EventTypeNormal, // Event type Normal for normal events and EventTypeWarning for warnings
					"PermanentErrorWasRemoved",
					"The permanent error was removed, because the annotation %q was removed",
                                        infrav1.PermanentErrorAnnotation
				)
```

**Implementation Details**:
For each `Reconciler` we need to add a field `recorder` of type `EventRecorder` from pkg `k8s.io/client-go/tools/record` and initialize it with `mgt.GetEventRecorderFor(<controller-name>)` e.g. `mgt.GetEventRecorderFor(hetznercluster-controller)` in the `SetupWithManger`

And each `record.Event`, `record.Eventf`, `record.Warn`, `record.Warnf` call then needs to be replaced with `Event` or `Eventf` instead.

Dhairya-Arora01:
In CAPI v1.14, the `util/record` pkg has been deprecated. As per the cluster-api migration guidelines the recommended replacement was controller-runtime's `mgr.GetEventRecorderFor`.
This PR updates our code to use the controller-runtime utility `mgr.GetEventRecorderFor` to get the event recorder.

The new EventRecorder doesn't contain the Warn/Warnf instead the Event/Eventf only.
(see more details in the issue #2264)

Fixes #2264

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -742,7 +742,12 @@ func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string)
 		message := fmt.Sprintf("%s. Remove annotation %q, if you want the controller to use the hbmh again.",
 			errMessage, PermanentErrorAnnotation)
 
-		record.Warn(host, "PermanentErrorSet", message)
+		recorder.Event(
+			host,
+			corev1.EventTypeWarning,
+			HetznerBareMetalHostActionCompletedPermanentErrorV1Beta2Reason,
```
janiskemper: should we mix the conditions reasons with the events? I think we didn't do that before.

janiskemper (commented):
I didn't check everything but seems good from high-level. We should double check in the review also that this approach also corresponds with upstream CAPI. @abdullah599 please do that!

controllers/hetznerbaremetalhost_controller.go
```diff
@@ -529,7 +548,12 @@ func hetznerSecretErrorResult(
 			Reason:  infrav1.HetznerBareMetalHostRobotCredentialsInvalidV1Beta2Reason,
 			Message: infrav1.ErrorMessageMissingOrInvalidSecretData,
 		})
-		record.Warnf(bmHost, infrav1.RobotCredentialsInvalidReason, err.Error())
+		r.EventRecorder.Event(
+			bmHost,
+			corev1.EventTypeWarning,
+			infrav1.HetznerBareMetalHostRobotCredentialsInvalidV1Beta2Reason,
```
abdullah599: this renames the event reason from RobotCredentialsInvalid to Invalid, and state_machine.go:213 now emits Invalid too. alone in `kubectl get events` neither one says what was invalid. can we keep the old reasons here and do the rename in its own PR? same everywhere

pkg/services/baremetal/host/host.go
```diff
@@ -2728,7 +3043,12 @@ func (s *Service) handleRobotRateLimitExceeded(err error, functionName string) {
 			Reason:  infrav1.HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason,
 			Message: msg,
 		})
-		record.Warnf(s.scope.HetznerBareMetalHost, "RateLimitExceeded", msg)
+		s.scope.EventRecorder.Event(
+			s.scope.HetznerBareMetalHost,
+			corev1.EventTypeWarning,
+			infrav1.HetznerBareMetalHostRobotRateLimitExceededV1Beta2Reason,
```
abdullah599: utils.go still emits RateLimitExceeded for the same thing. Exceeded on its own doesn't say what was exceeded

pkg/services/baremetal/host/host.go
```diff
@@ -1707,14 +1864,30 @@ func (s *Service) actionImageInstallingStartBackgroundProcess(ctx context.Contex
 			Reason:  infrav1.HetznerBareMetalHostSSHToRescueSystemFailedV1Beta2Reason,
 			Message: msg,
 		})
-		record.Event(s.scope.HetznerBareMetalHost, infrav1.SSHToRescueSystemFailedReason, msg)
+		s.scope.EventRecorder.Event(
+			s.scope.HetznerBareMetalHost,
+			corev1.EventTypeNormal,
```
abdullah599: this is a failure and you changed state_machine.go:183 to Warning, so should we do the same here?

pkg/services/baremetal/host/state_machine.go
```diff
@@ -176,7 +178,12 @@ func (hsm *hostStateMachine) updateOSSSHStatusAndValidateKey(osSSHSecret *corev1
 			hsm.nextState = infrav1.StateImageInstalling
 		case infrav1.StateProvisioned:
 			errMessage := "secret has been modified although a provisioned machine uses it"
-			record.Event(hsm.host, "SSHSecretUnexpectedlyModified", errMessage)
+			hsm.eventRecorder.Event(
+				hsm.host,
+				corev1.EventTypeWarning,
```
abdullah599: this was Normal before. deliberate?

pkg/services/baremetal/host/state_machine.go
```diff
@@ -200,7 +207,12 @@ func (hsm *hostStateMachine) updateOSSSHStatusAndValidateKey(osSSHSecret *corev1
 			Message: msg,
 		})
 
-		record.Warnf(hsm.host, infrav1.SSHKeyAlreadyExistsReason, msg)
+		hsm.eventRecorder.Event(
+			hsm.host,
+			corev1.EventTypeWarning,
+			infrav1.HetznerBareMetalHostSSHKeysInvalidV1Beta2Reason,
```
abdullah599: what about SSHCredentialsInSecretInvalid, like the condition above uses?

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -1105,7 +1106,8 @@ func Test_removePermanentErrorIfAnnotationIsGone_AnnotationPresent(t *testing.T)
 			},
 		},
 	}
-	removed := removePermanentErrorIfAnnotationIsGone(&bmHost)
+	reconciler := &HetznerBareMetalHostReconciler{EventRecorder: record.NewFakeRecorder(10)}
```
abdullah599: line 120 builds the reconciler without an EventRecorder and calls Reconcile. it only passes because the paused guard returns first

pkg/services/baremetal/baremetal/baremetal_suite_test.go
```diff
@@ -44,6 +45,7 @@ func newTestService(
 			Logger:           log,
 			Client:           client,
 			BareMetalMachine: bmMachine,
+			EventRecorder:    record.NewFakeRecorder(10),
```
abdullah599: why 10? host_suite_test.go uses 100

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -722,7 +722,7 @@ func (host *HetznerBareMetalHost) NeedsProvisioning() bool {
 }
 
 // SetError updates the error type and message in the status struct and increases the ErrorCount.
-func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string) {
+func (host *HetznerBareMetalHost) SetError(recorder record.EventRecorder, errType ErrorType, errMessage string) {
```
abdullah599: I think we should notpull client-go/record into api/v1beta1 as that’s the package people import just to read CRs. both callers already hold the recorder, so SetError can return whether it set a permanent error and let them send the event

wdyt?

abdullah599: the doc comment above doesn't mention the event

pkg/scope/cluster.go
```diff
@@ -66,6 +68,9 @@ func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
 	if params.APIReader == nil {
 		return nil, errors.New("failed to generate new scope from nil APIReader")
 	}
+	if params.EventRecorder == nil {
+		return nil, errors.New("cannot create cluster scope without EventRecorder")
```
abdullah599: every other check in here says "failed to generate ...". we should do the same here imo

pkg/scope/machine.go
```diff
@@ -92,6 +93,9 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 	if params.APIReader == nil {
 		return nil, errors.New("failed to generate new scope from nil APIReader")
 	}
+	if params.EventRecorder == nil {
+		return nil, errors.New("cannot create machine scope without EventRecorder")
```
abdullah599: is this branch tested anywhere? nothing in pkg/scope calls the constructors

pkg/services/baremetal/host/state_machine.go
```diff
@@ -35,21 +35,23 @@ import (
...
 // the states of a BareMetalHost.
 type hostStateMachine struct {
-	host       *infrav1.HetznerBareMetalHost
-	reconciler *Service
-	nextState  infrav1.ProvisioningState
-	log        logr.Logger
+	host          *infrav1.HetznerBareMetalHost
+	reconciler    *Service
+	nextState     infrav1.ProvisioningState
+	log           logr.Logger
```
abdullah599: hsm.reconciler.scope.EventRecorder is the same value, host.go:126 passes both side by side. can we drop the field?

controllers/csr_controller.go
```diff
@@ -313,6 +334,13 @@ func (r *GuestCSRReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Mana
 			},
 		}).
 		Complete(r)
+	if err != nil {
+		return fmt.Errorf("error creating controller: %w", err)
+	}
+
+	r.EventRecorder = mgr.GetEventRecorderFor("csr-controller")
```
abdullah599: this also moves CSR events to the workload cluster, where the CSR actually lives. before they went to the management cluster. can you mention it in the description?

controllers/hcloudmachinetemplate_controller.go
```diff
@@ -254,11 +257,17 @@ func (r *HCloudMachineTemplateReconciler) reconcile(ctx context.Context, machine
...
+	err := ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options).
 		For(&infrav1.HCloudMachineTemplate{}).
 		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), ctrl.LoggerFrom(ctx), r.WatchFilterValue)).
 		Complete(r)
+	if err != nil {
+		return fmt.Errorf("error creating controller: %w", err)
+	}
+
+	r.EventRecorder = mgr.GetEventRecorderFor("hcloudmachinetemplate-controller")
```
abdullah599: this one was a single return before. the assignment doesn't need Complete to have succeeded, so it could go at the top and keep the old shape. same in csr, hcloudremediation and hetznerbaremetalremediation

controllers/hetznerbaremetalhost_controller_test.go
abdullah599, line 120: can we set a recorder here as well?

pkg/services/baremetal/host/state_machine.go
```diff
@@ -35,21 +35,23 @@ import (
...
+	eventRecorder record.EventRecorder
```
abdullah599: hsm.reconciler.scope.EventRecorder is the same value no?

abdullah599 (commented):
one more: .golangci.yaml:212 says the exclusion is there because record.InitFromRecorder takes the old client-go recorder. that function is gone from the repo after this PR

pkg/services/baremetal/host/host.go
```diff
@@ -137,11 +136,25 @@ func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err e
...
 
+// setHostError updates the host's error status and, if the error is permanent, emits the
+// corresponding warning event since HetznerBareMetalHost.SetError itself must stay free of any
+// EventRecorder dependency.
+func (s *Service) setHostError(errType infrav1.ErrorType, errMessage string) {
+	if permanentErrorSet, message := s.scope.HetznerBareMetalHost.SetError(errType, errMessage); permanentErrorSet {
+		s.scope.EventRecorder.Event(
+			s.scope.HetznerBareMetalHost,
+			corev1.EventTypeWarning,
+			"PermanentErrorSet",
```
abdullah599: this was a single literal inside SetError before, now it's here, in remediation.go:299 and in the doc comment. every other reason around here comes from a constant, can we add one?

controllers/hetznerbaremetalhost_controller_test.go
```diff
@@ -120,6 +121,7 @@ func TestHetznerBareMetalHostReconciler_ReconcileSkipsPausedCluster(t *testing.T
 		Client:             c,
 		APIReader:          c,
 		RobotClientFactory: robotFactory,
+		EventRecorder:      record.NewFakeRecorder(10),
```
abdullah599, line 124: [this one is new and still 10](https://github.com/syself/cluster-api-provider-hetzner/pull/2267#discussion_r3996015010 also baremetal_test.go:1025 and :1542

pkg/scope/machine.go
abdullah599, line 97: this one still differs from its neighbours. same in hcloudremediation.go, baremetalremediation.go and hcloudmachinetemplate.go

pkg/services/baremetal/host/host.go
```diff
@@ -137,11 +136,25 @@ func (s *Service) Reconcile(ctx context.Context) (result reconcile.Result, err e
...
 
 func (s *Service) recordActionFailure(errorType infrav1.ErrorType, errorMessage string) actionFailed {
-	s.scope.HetznerBareMetalHost.SetError(errorType, errorMessage)
+	s.setHostError(errorType, errorMessage)
 	s.scope.Error(errActionFailure, errorMessage, "errorType", errorType)
 	return actionFailed{ErrorType: errorType, errorCount: s.scope.HetznerBareMetalHost.Spec.Status.ErrorCount}
 }
 
+// setHostError updates the host's error status and, if the error is permanent, emits the
+// corresponding warning event since HetznerBareMetalHost.SetError itself must stay free of any
```
abdullah599, line 145: the second half explains why we shaped SetError that way, which only makes sense if you read the review. can we just say what setHostError does?

abdullah599 (commented):
minor nits else LGTM :)

cc: @janiskemper

api/v1beta1/hetznerbaremetalhost_types.go
```diff
@@ -718,7 +717,9 @@ func (host *HetznerBareMetalHost) NeedsProvisioning() bool {
 }
 
 // SetError updates the error type and message in the status struct and increases the ErrorCount.
-func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string) {
+// When errType is PermanentError, it reports permanentErrorSet as true along with the message that
+// callers holding an EventRecorder should emit as a "PermanentErrorSet" warning event.
+func (host *HetznerBareMetalHost) SetError(errType ErrorType, errMessage string) (permanentErrorSet bool, message string) {
```
janiskemper: I don't yet understand why we change this function.

pkg/services/baremetal/host/host.go
janiskemper: @Dhairya-Arora01 I believe "SetError must stay free of eventRecorder" is something we shouldn't necessarily have here.

pkg/services/baremetal/host/host.go
```diff
@@ -2217,9 +2497,14 @@ func (s *Service) handleCloudInitNotStarted(ctx context.Context) actionResult {
 
 		s.scope.HetznerBareMetalHost.Spec.Status.RebootTriggeredAt = ptr.To(metav1.Now())
 
-		createSSHRebootEvent(ctx, s.scope.HetznerBareMetalHost, "machine image and cloud-init data got installed")
-		record.Eventf(s.scope.HetznerBareMetalHost,
-			"SSHRebootAfterCloudInitSigTermFound", "rebooted via ssh after cloud init logs contained sigterm: %s", trimLineBreak(out.StdOut))
+		s.createSSHRebootEvent(ctx, "machine image and cloud-init data got installed")
```
janiskemper: I believe we generate two events for the same thing here. Can you check? (unrelated to switching the event recorder)

Dhairya-Arora01: yes you are right

janiskemper (dismissed):

janiskemper (approved):

PR #2270, opened 2026-09-11
🐛 Stop storing the ssh connection refused error in errorType

### LINKED ISSUE #2269: HetznerBareMetalHost: the connection refused error and the fatal error overwrite each other in `errorType`

To provision a bare metal server, the HetznerBareMetalHost controller first reboots it into the rescue system and waits until ssh answers, to collect the hardware details. When ssh refuses the connection, the controller writes `connection refused error of SSH command` into `spec.status.errorType`. On the next reconcile it checks whether the reboot was more than ten minutes ago. If it was, the controller gives up on the host and writes `fatal error` into the same field.

Using `spec.status.errorType` for both is the problem. The controller only checks the ten minute timeout while the field still holds `connection refused error of SSH command` ([host.go#L406-L408](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L406-L408)), and the fatal error replaces that value.

So once ssh has been refused for more than ten minutes, the reconciles alternate:

- [reconcile 1](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L408-L428): the field holds `connection refused error of SSH command` and the timeout has passed, so the controller [writes `fatal error`](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L671-L674). The ProvisionSucceeded reason becomes SSHConnectionRefused and it requeues after 1 to 2 minutes
- [reconcile 2](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L430-L433): the field holds `fatal error`, so the timeout is not checked. The controller writes `connection refused error of SSH command` back and returns. The reason becomes StillProvisioning and it requeues after 10 seconds
- reconcile 3: same as reconcile 1

This was observed on a host that stayed in registering for over an hour after the reboot

> [!IMPORTANT]
> `handleIncompleteBoot` is called in two places, `actionRegistering` and `actionEnsureProvisioned`. Everything above is about `actionRegistering`.
>
> `actionEnsureProvisioned` overwrites the condition as well. `handleIncompleteBoot` sets the ProvisionSucceeded reason to SSHConnectionRefused when it gives up, and [right after that call the controller writes StillProvisioning over it](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L2056-L2057). A host the controller gave up on then reports that it is still provisioning.

**What should happen:**

The controller should stop writing `connection refused error of SSH command` into `errorType`.

> [!note]
> We used to put `connection refused error of SSH command` into `errorType` but after this it will only be logged. This is fine because the other ssh errors are already only logged, for example a wrong ssh key or an empty hostname.

The ten minute timeout is already measured from [`spec.status.rebootTriggeredAt`](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L408), so the check can read that timestamp and does not need anything in `errorType`.

Then the `ProvisionSucceeded` reason stays `SSHConnectionRefused`, `errorType` stays `fatal error`, and `errorCount` increases with every retry.

`actionEnsureProvisioned` should not write the condition again after `handleIncompleteBoot` gave up, the same way [`actionRegistering`](https://github.com/syself/cluster-api-provider-hetzner/blob/b5e7742262bb0666424ba101ace3c9fffdfe8ea2/pkg/services/baremetal/host/host.go#L671-L674) does not. An operator then reads `SSHConnectionRefused` or `RebootTimedOut` on the host instead of `StillProvisioning`.

abdullah599:
Stop storing the ssh connection refused error in errorType

Fixes #2269 

none

**TODOs**:

- [ ] squash commits
- [ ] include documentation
- [ ] add unit tests


janiskemper (commented):
LGTM. Didn't check the tests though

Dhairya-Arora01 (approved):
LGTM

